PR #148, opened 2023-07-04
GO-1490: add cache for state

AnastasiaShemyakinskaya:


fat-fellow (changes requested):
Need to discuss if we need to merge

PR #194, opened 2023-07-17
GO-1677 Add debug open object list

fat-fellow:


uburoiubu (approved):

core/block/service.go
```diff
@@ -255,6 +257,7 @@ func (s *Service) OpenBlock(
 		FileWatcherMs:  afterHashesTime.Sub(afterShowTime).Milliseconds(),
 		SmartblockType: int(sbType),
 	})
+	s.openedObjects[id] = true
```
deff7, line 260: it's not concurrency safe, you should use mutex

PR #195, opened 2023-07-17
improve makefile and readme

requilence:
- install grpc-web plugin to the deps instead of global
- install mockery to the deps
- add install instruction for linux


fat-fellow (changes requested):
Please return -race flag

fat-fellow (approved):

PR #199, opened 2023-07-18
GO-1609: add tests for state store and simple block diffs

AnastasiaShemyakinskaya:


fat-fellow (approved):

core/block/simple/dataview/diff_test.go
```diff
@@ -0,0 +1,1126 @@
...
+		b2 := testBlock()
+
+		// when
+		b2.content.IsCollection = true
+		diff, err := b1.Diff(b2)
+
+		// then
+		require.NoError(t, err)
+		require.Len(t, diff, 1)
+		change := diff[0].Msg.Value.(*pb.EventMessageValueOfBlockDataviewIsCollectionSet).BlockDataviewIsCollectionSet
```
deff7, line 31: I think we should use more declarative-style assertions, something like this

```
		assert.Equal(t, makeEvents(
			&pb.EventMessageValueOfBlockDataviewIsCollectionSet{
				BlockDataviewIsCollectionSet: &pb.EventBlockDataviewIsCollectionSet{
					Value: true,
				},
			}), diff)
```

where makeEvents is helper for building []simple.EventMessage


Such approach will lead to a couple of positive effects:
1. Tests will become more comprehensible and less error-prone
2. If Diff will produce changes in other fields then test shows it immediately

AnastasiaShemyakinskaya: Thank you, fixed

deff7 (approved):

PR #213, opened 2023-07-21
GO-1677 Add debug open object list

fat-fellow:


core/block/service.go
```diff
@@ -257,7 +266,7 @@ func (s *Service) OpenBlock(
 		FileWatcherMs:  afterHashesTime.Sub(afterShowTime).Milliseconds(),
 		SmartblockType: int(sbType),
 	})
-	s.openedObjects[id] = true
+	mutex.WithLock(s.openedObjs.lock, func() any { s.openedObjs.objects[id] = true; return nil })
```
deff7, line 269: it doesn't look idiomatic in go

fat-fellow: We have decided to leave it as it is

deff7 (approved):

PR #220, opened 2023-07-26
GO-1777: add simple nested query filter

deff7:


pkg/lib/database/filter.go
```diff
@@ -16,8 +18,10 @@ var (
 	ErrValueMustBeListSupporting = errors.New("value must be list supporting")
 )
 
-func MakeAndFilter(protoFilters []*model.BlockContentDataviewFilter, store OptionsGetter) (AndFilters, error) {
-
+func MakeAndFilter(protoFilters []*model.BlockContentDataviewFilter, store ObjectStore) (AndFilters, error) {
```
AnastasiaShemyakinskaya, line 21: This name of the function is strange to me :(

pkg/lib/database/filter.go
```diff
@@ -33,135 +37,168 @@ func MakeAndFilter(protoFilters []*model.BlockContentDataviewFilter, store Optio
...
 }
 
-func MakeFilter(proto *model.BlockContentDataviewFilter, store OptionsGetter) (Filter, error) {
+func MakeFilter(rawFilter *model.BlockContentDataviewFilter, store ObjectStore) (Filter, error) {
+	if store == nil {
+		return AndFilters{}, fmt.Errorf("objectStore dependency is nil")
+	}
+	parts := strings.SplitN(rawFilter.RelationKey, ".", 2)
+	if len(parts) == 2 {
+		rawNestedFilter := proto.Clone(rawFilter).(*model.BlockContentDataviewFilter)
```
AnastasiaShemyakinskaya, line 46: May be extract to separate function?

pkg/lib/database/filter.go
```diff
@@ -201,6 +252,18 @@ func (a OrFilters) String() string {
 	return fmt.Sprintf("(%s)", strings.Join(orS, " OR "))
 }
 
+func (a OrFilters) IterateNestedFilters(fn func(nestedFilter Filter) error) error {
+	for _, f := range a {
```
AnastasiaShemyakinskaya, line 256: It's seems like duplicated logic to 
```
func (a AndFilters) IterateNestedFilters(fn func(nestedFilter Filter) error) error {
	for _, f := range a {
		if withNested, ok := f.(WithNestedFilter); ok {
			err := withNested.IterateNestedFilters(fn)
			if err != nil {
				return err
			}
		}
	}
	return nil
}
```

pkg/lib/database/filter.go
fat-fellow: the same issue, but do you have any idea how to make it batter?
possibly MakeFilterAnd can sound slightly better

AnastasiaShemyakinskaya: Yes, sounds better. Probably BuildFilterAnd

deff7: Thank you, I renamed AndFilter and OrFilter to `FiltersAnd` and `FiltersOr`, think it's less confusing. Also added prefix `Filter` for all filters

deff7: Now it's `MakeFiltersAnd`

pkg/lib/database/filter.go
deff7: done!

pkg/lib/database/filter.go
deff7: extracted, thank you

fat-fellow (approved):

PR #223, opened 2023-07-26
GO-1661: protobuf does not export files

AnastasiaShemyakinskaya:


core/block/export/export.go
```diff
@@ -274,12 +274,25 @@ func (e *export) getExistedObjects(includeArchived bool, isProtobuf bool) (map[s
...
-				return mw.Add(b.NewState().Copy())
+				if err = mw.Add(b.NewState().Copy()); err != nil {
+					return err
+				}
+				if !includeFiles {
+					return nil
+				}
+				for _, fh := range b.GetAndUnsetFileKeys() {
+					if werr := e.saveFile(wr, fh.Hash); werr != nil {
+						if werr = e.saveImage(wr, fh.Hash); werr != nil {
```
deff7, line 290: If we got error while saving file it's not implying that we have an image. Maybe just call safe file? Or collect image hashes separately, but I dont' think it's necessary

deff7 (approved):

PR #229, opened 2023-07-27
GO-1569: fix for imported bookmarks without title, image and description

AnastasiaShemyakinskaya:


core/block/editor/bookmark/bookmark.go
```diff
@@ -201,15 +202,21 @@ func (b *sbookmark) MigrateBlock(bm bookmark.Block) error {
...
 	}
 
-	pageId, _, err := b.bookmarkSvc.CreateBookmarkObject(bm.ToDetails(), func() *model.BlockContentBookmark {
+	bookmarkContent := func() *model.BlockContentBookmark {
 		return content
-	})
+	}
+
+	if content.Description == "" && content.FaviconHash == "" && content.Title == "" {
+		bookmarkContent = b.bookmarkSvc.FetchBookmarkContent(content.Url)
```
deff7, line 210: Maybe do it in bookmark syncer? This code is called only in import

deff7: I think we have to delete MigrateBlock completely because it's called only in state migration. Seems like import is the only case when something meaningful is happening here

AnastasiaShemyakinskaya: Removed migration completely

deff7 (approved):

PR #257, opened 2023-08-03
GO-1783 add disableLocalNetworkSync to AccountSelect/Create

requilence:
alongside with AccountStartLocalNetworkSync RPC

Client can disable or postpone the local network discovery/sync. It is useful on iOS to avoid "Allow access to local network" on start


core/account.go
```diff
@@ -131,7 +132,11 @@ func (mw *Middleware) AccountCreate(cctx context.Context, req *pb.RpcAccountCrea
 		response(nil, pb.RpcAccountCreateResponseError_FAILED_TO_STOP_RUNNING_NODE, err)
 	}
 
-	cfg := anytype.BootstrapConfig(true, os.Getenv("ANYTYPE_STAGING") == "1", true)
```
fat-fellow, line 134: DRY. Please extract to the function

fat-fellow (approved):

PR #269, opened 2023-08-04
GO-1815: fix absent tags in relations

AnastasiaShemyakinskaya:
---

- [ ] I understand that contributing to this repository will require me to agree with the [CLA](https://github.com/anyproto/.github/blob/main/docs/CLA.md)

---

### Description

After repeated imports new imported tags aren't added to corresponding relation, because they have different relation key in details during creation. So we replace it with the right one

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

https://linear.app/anytype/issue/GO-1815/tags-do-not-migrate

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/core/block/import/importer.go
+++ b/core/block/import/importer.go
@@ -3,6 +3,7 @@ package importer
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/anyproto/any-sync/app"
@@ -31,6 +32,7 @@ import (
 	"github.com/anyproto/anytype-heart/pkg/lib/bundle"
 	"github.com/anyproto/anytype-heart/pkg/lib/core"
 	sb "github.com/anyproto/anytype-heart/pkg/lib/core/smartblock"
+	"github.com/anyproto/anytype-heart/pkg/lib/localstore/addr"
 	"github.com/anyproto/anytype-heart/pkg/lib/localstore/filestore"
 	"github.com/anyproto/anytype-heart/pkg/lib/localstore/objectstore"
 	"github.com/anyproto/anytype-heart/pkg/lib/logging"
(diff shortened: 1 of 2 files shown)


core/block/import/importer.go
```diff
@@ -294,6 +297,18 @@ func (i *Import) getIDForAllObjects(ctx *session.Context, res *converter.Respons
 	return oldIDToNew, createPayloads, nil
 }
 
+func (i *Import) replaceRelationKeyWithNew(option *converter.Snapshot, oldIDToNew map[string]string) {
```
fat-fellow, line 300: do we need some tests here, what do you think?

AnastasiaShemyakinskaya: Do you mean only `replaceRelationKeyWithNew`? Because testing `getIDForAllObjects` makes sense with real object store

AnastasiaShemyakinskaya: Added test

fat-fellow (approved):

core/block/import/importer_test.go
```diff
@@ -606,3 +610,72 @@ func Test_ImportErrLimitExceededIgnoreErrorMode(t *testing.T) {
...
+							bundle.RelationKeyRelationKey.String(): pbtypes.String("key"),
+						},
+					},
+				},
+			},
+			SbType: smartblock.SmartBlockTypeSubObject,
+		}
+		oldIDToNew := make(map[string]string, 0)
+
+		// then
```
deff7, line 632: `then` and `when` are in wrong order

AnastasiaShemyakinskaya: Fixed

PR #280, opened 2023-08-07
GO-1724: fix extra untitled objects and adjust Notion progress bar process count

AnastasiaShemyakinskaya:
---

---

### Description

1. If you import a table with for example 1 object, that do not have a name, and then import the same table, but with renamed object, then the collection will have extra objects. That's because we identify same objects by it's sourceFileName relation. But in Notion import we set it as page/db URL. The problem is different URLs for objects with and without name
Object with name has URL like "Test-id"
Object without name has URL like "id"
So both objects considered as different objects by our import. To avoid it I use page/db ID as a sourceFileName to uniquely identify them
!BUT it could break existing objects import in account, because we change sourceFilePath for all objects. So if user want to repeat his/her import after this change, objects will be duplicated
2. Notion progress bar doesn't take into account creation of subobjects. So progress bar window can disappear earlier than import finishes.

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

https://linear.app/anytype/issue/GO-1724#comment-172af1e8

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/core/block/import/notion/api/database/database.go
+++ b/core/block/import/notion/api/database/database.go
@@ -218,7 +218,7 @@ func (ds *Service) getRelationDetails(databaseProperty property.DatabaseProperty
 
 func (ds *Service) getCollectionDetails(d Database) map[string]*types.Value {
 	details := make(map[string]*types.Value, 0)
-	details[bundle.RelationKeySourceFilePath.String()] = pbtypes.String(d.URL)
+	details[bundle.RelationKeySourceFilePath.String()] = pbtypes.String(d.ID)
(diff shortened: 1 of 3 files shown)


fat-fellow (approved):

AnastasiaShemyakinskaya:
Need to discuss the solution

PR #283, opened 2023-08-08
makefile: add ANYTYPE_NO_LOCAL_AUTH env var

requilence:
### Description

This PR add an easy ability to pass the noauth build flag. It is used to test the local grpc API without the need to auth. 
⚠️ should be used only for testing purposes. In case auth is disabled anyone can access the local API and read the account's data

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [v] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [ ] 👍 yes
- [v] 🙅 no, because they aren't needed

### Added to documentation?

- [ ] 📜 README.md
- [v] 🙅 no documentation needed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -237,6 +237,10 @@ build-server: setup-network-config
 ifdef ANY_SYNC_NETWORK
 	@$(eval TAGS := $(TAGS) envnetworkcustom)
 endif
+
+ifdef ANYTYPE_NO_LOCAL_AUTH
+	@$(eval TAGS := $(TAGS) noauth)
+endif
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -237,6 +237,10 @@ build-server: setup-network-config
 ifdef ANY_SYNC_NETWORK
 	@$(eval TAGS := $(TAGS) envnetworkcustom)
 endif
+
```
fat-fellow, line 240: add a description in readme

requilence: it's on purpose. we don't want to highlight it because it's not secure

PR #295, opened 2023-08-11
GO-1817: Some optimizations

deff7:
- Use strings builder in state.SearchText method
- Remove unnecessary code from fetchMeta function. GetObjectTypes is no longer needed here, but it consumes a lot of resources
- Restrictions: use simplified HasObjectType function instead of resource-intensive GetObjectType
- Don't copy State for indexation
- Don't rebuild full-text index when account is just recovered


deff7:
For example, SearchText allocated more than 1GB memory when recovering a big account:
```
    1.32GB 13.34% 48.53%     1.32GB 13.34%  github.com/anyproto/anytype-heart/core/block/editor/state.(*State).SearchText.func1
```

fat-fellow (approved):

PR #348, opened 2023-08-30
GO-1588: import html doesn't import image

AnastasiaShemyakinskaya:
---

---

### Description

Currently HTML imports have a problem displaying images, that are located on the machine. So the import works fine with web links to images, but it doesn't work with local files.

Therefore, when importing html with links to a local file, files loading does not work. Because the path is not always absolute, or we are importing an archive. 

Therefore, when importing an html file that has a link to a local file, we look at the path to this file. If path is relative, then we replace the path with an absolute one (in particular, when importing html from Notion, the links to files are relative). If the file is in a zip archive, then we upload it to a temporary directory

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

https://linear.app/anytype/issue/GO-1588/import-html-doesnt-import-image

### Added tests?

- [x] 👍 yes

### Added to documentation?

- [ ] 📜 README.md


core/block/import/syncer/file.go
```diff
@@ -50,6 +49,5 @@ func (fs *FileSyncer) Sync(ctx *session.Context, id string, b simple.Block) erro
 	if err != nil {
 		return fmt.Errorf("failed syncing file: %s", err)
 	}
-	os.Remove(b.Model().GetFile().Name)
```
AnastasiaShemyakinskaya, line 53: Remove it, because it's not safe to remove files on user local machine. Originally we delete file after its loading, because in Markdown import we copy media files to temporary directory and then load them. So we try to remove file from temporary directory here after loading. But its common code for all import now, and there is a probability, that file Name contain path to not temporary file on local machine. And this code can delete some important files in theory

core/block/import/source/source.go
```diff
@@ -37,3 +38,13 @@ func isSupportedExtension(ext string, expectedExt []string) bool {
 func isFileAllowedToImport(fileName, ext string, expectedExt, includeFiles []string) bool {
 	return isSupportedExtension(ext, expectedExt) || lo.Contains(includeFiles, filepath.Base(fileName))
 }
+
+func CountFilesWithGivenExtension(fileReaders map[string]io.ReadCloser, extension string) int {
```
AnastasiaShemyakinskaya, line 42: We use this function to calculate count of files with extensions like html or md. Because source directory or zip archive can contain not only html or md files, but also image files or video.

core/block/import/markdown/blockconverter.go
```diff
@@ -61,14 +57,14 @@ func (m *mdConverter) processFiles(importPath string, mode string, allErrors *ce
...
-	readers, err := s.GetFileReaders(importPath, supportedExtensions, nil)
+	readers, err := importSource.GetFileReaders(importPath, supportedExtensions, nil)
 	if err != nil {
 		allErrors.Add(err)
 		if mode == pb.RpcObjectImportRequest_ALL_OR_NOTHING.String() {
 			return nil
 		}
 	}
-	if len(readers) == 0 {
+	if source.CountFilesWithGivenExtension(readers, ".md") == 0 {
```
AnastasiaShemyakinskaya, line 67: We consider, that there is no files to import in case, when directory or zip archive doesn't have any `md` files. If directory or zip archive contain other files, like images or video, it still means, that there is no file to import.

core/block/import/source/source.go
```diff
@@ -16,6 +16,7 @@ var extensions = []string{".md", ".csv", ".txt", ".pb", ".json", ".html"}
 
 type Source interface {
 	GetFileReaders(importPath string, ext []string, includeFiles []string) (map[string]io.ReadCloser, error)
+	Close()
```
AnastasiaShemyakinskaya, line 19: We use method Close() to close all imported files and all files from source archive or directory

fat-fellow (approved):

PR #366, opened 2023-09-04
GO-1915: Fix critical bug in query

deff7:


pkg/lib/localstore/objectstore/queries_test.go
```diff
@@ -404,11 +407,42 @@ func TestQuery(t *testing.T) {
 		}, recs)
 	})
 
+	t.Run("with offset", func(t *testing.T) {
```
fat-fellow, line 410: consider adding //given //when //then

fat-fellow (approved):

PR #383, opened 2023-09-06
Go 1942: fix many problems from import logs

AnastasiaShemyakinskaya:
---

---

### Description

1. Fix problem with importing icon from blocks (callout block). In case it is PB import, we already have hash in icon, so we skip upload of custom icon
2. Notion properties can be nameless. So in such case we try to create object without name, and it leads to errors. So I added name to such properties
3. To avoid errors with imported file relations, I add empty hash in case we failed to upload file
4. To avoid problem with many open files during import, I added iterator. So we open file, when it's needed and close it immediately after usage

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [x] 🎨 Style
- [x] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

https://linear.app/anytype/issue/GO-1942/research-import-logs

### Added tests?

- [x] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/core/block/import/importer.go
+++ b/core/block/import/importer.go
@@ -131,8 +131,9 @@ func (i *Import) sendFileEvents(returnedErr error) {
 func (i *Import) importFromBuiltinConverter(ctx *session.Context,
 	req *pb.RpcObjectImportRequest,
 	c converter.Converter,
-	progress process.Progress) (string, error) {
-	allErrors := converter.NewError(0)
+	progress process.Progress,
+) (string, error) {
+	allErrors := converter.NewError(req.Mode)
 	res, err := c.GetSnapshots(req, progress)
 	if !err.IsEmpty() {
 		resultErr := err.GetResultError(req.Type)
@@ -157,8 +158,11 @@ func (i *Import) importFromBuiltinConverter(ctx *session.Context,
 	return rootCollectionID, resultErr
 }
 
-func (i *Import) importFromExternalSource(ctx *session.Context, req *pb.RpcObjectImportRequest, progress process.Progress) error {
-	allErrors := converter.NewError(0)
+func (i *Import) importFromExternalSource(ctx *session.Context,
+	req *pb.RpcObjectImportRequest,
+	progress process.Progress,
+) error {
+	allErrors := converter.NewError(req.Mode)
 	if req.Snapshots != nil {
 		sn := make([]*converter.Snapshot, len(req.Snapshots))
 		for i, s := range req.Snapshots {
@@ -252,7 +256,12 @@ func (i *Import) ImportWeb(ctx *session.Context, req *pb.RpcObjectImportRequest)
 	return res.Snapshots[0].Id, details[res.Snapshots[0].Id], nil
 }
 
-func (i *Import) createObjects(ctx *session.Context, res *converter.Response, progress process.Progress, req *pb.RpcObjectImportRequest, allErrors *converter.ConvertError) (map[string]*types.Struct, string) {
+func (i *Import) createObjects(ctx *session.Context,
+	res *converter.Response,
+	progress process.Progress,
+	req *pb.RpcObjectImportRequest,
+	allErrors *converter.ConvertError,
+) (map[string]*types.Struct, string) {
 	oldIDToNew, createPayloads, err := i.getIDForAllObjects(ctx, res, allErrors, req)
 	if err != nil {
 		return nil, ""
@@ -281,7 +290,11 @@ func (i *Import) getFilesIDs(res *converter.Response) []string {
 	return fileIDs
 }
 
-func (i *Import) getIDForAllObjects(ctx *session.Context, res *converter.Response, allErrors *converter.ConvertError, req *pb.RpcObjectImportRequest) (map[string]string, map[string]treestorage.TreeStorageCreatePayload, error) {
+func (i *Import) getIDForAllObjects(ctx *session.Context,
+	res *converter.Response,
+	allErrors *converter.ConvertError,
+	req *pb.RpcObjectImportRequest,
+) (map[string]string, map[string]treestorage.TreeStorageCreatePayload, error) {
 	relationOptions := make([]*converter.Snapshot, 0)
 	oldIDToNew := make(map[string]string, len(res.Snapshots))
 	createPayloads := make(map[string]treestorage.TreeStorageCreatePayload, len(res.Snapshots))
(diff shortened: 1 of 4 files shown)


core/block/import/csv/converter.go
```diff
@@ -60,118 +60,118 @@ func (c *CSV) GetSnapshots(req *pb.RpcObjectImportRequest, progress process.Prog
 	if params == nil {
 		return nil, nil
 	}
-	cErr := converter.NewError()
-	result, cancelError := c.createObjectsFromCSVFiles(req, progress, params, cErr)
-	if !cancelError.IsEmpty() {
```
AnastasiaShemyakinskaya, line 65: Move error check in function `needToReturnError`

core/block/import/syncer/relationsyncer.go
```diff
@@ -93,11 +92,14 @@ func (fs *FileRelationSyncer) updateFileRelationsDetails(st *state.State, name s
...
 		return
 	}
-	if st.Details().Fields[name].GetListValue() != nil && len(allFilesHashes) != 0 {
+	if st.Details().Fields[name].GetListValue() != nil {
 		st.SetDetail(name, pbtypes.StringList(allFilesHashes))
 	}
-
-	if st.Details().Fields[name].GetStringValue() != "" && len(allFilesHashes) != 0 {
-		st.SetDetail(name, pbtypes.String(allFilesHashes[0]))
+	hash := ""
```
AnastasiaShemyakinskaya, line 98: If we failed to load file, we make file relation empty

core/block/import/source/source.go
```diff
@@ -15,7 +15,10 @@ var log = logging.Logger("import-source")
 var extensions = []string{".md", ".csv", ".txt", ".pb", ".json", ".html"}
 
 type Source interface {
-	GetFileReaders(importPath string, ext []string, includeFiles []string) (map[string]io.ReadCloser, error)
+	Initialize(importPath string) error
```
AnastasiaShemyakinskaya, line 18: In `Initialized` we collect all files from given resource (zip archive, directory or simple file) and save them in `map`

core/block/import/notion/api/database/database.go
```diff
@@ -231,6 +231,9 @@ func (ds *Service) getRelationSnapshot(relationKey string, databaseProperty prop
 }
 
 func (ds *Service) getRelationDetails(databaseProperty property.DatabasePropertyHandler, name, key string) *types.Struct {
+	if name == "" {
```
AnastasiaShemyakinskaya, line 234: Notion properties can be nameless, so in such case we name them `Untitled`

core/block/import/markdown/blockconverter.go
```diff
@@ -241,40 +236,33 @@ func (m *mdConverter) createBlocksFromFile(shortPath string, f io.ReadCloser, fi
...
+		if err != nil {
+			return err
+		}
+		defer tmpFile.Close()
+		w := bufio.NewWriter(tmpFile)
+		_, err = w.ReadFrom(fileReader)
+		if err != nil {
+			return err
+		}
+		if err := w.Flush(); err != nil {
```
KirillSto, line 259: It seems that you reinitialize err one more time - err is already created in line 249

core/block/import/txt/converter.go
```diff
@@ -162,3 +164,9 @@ func (t *TXT) getSnapshot(blocks []*model.Block, p string) (*converter.Snapshot,
 	}
 	return snapshot, snapshot.Id
 }
+
+func (t *TXT) shouldReturnError(req *pb.RpcObjectImportRequest, cErr *converter.ConvertError, path []string) bool {
+	return (!cErr.IsEmpty() && req.Mode == pb.RpcObjectImportRequest_ALL_OR_NOTHING) ||
+		(cErr.IsNoObjectToImportError(len(path))) ||
```
KirillSto, line 170: I think outer braces are redundant here. It is a style thing, but this complex logical expression will look a little bit more readable without them.
Moreover `&&` is computed before `||` , so return value could be simplified to:
```
return !cErr.IsEmpty() && req.Mode == pb.RpcObjectImportRequest_ALL_OR_NOTHING ||
		cErr.IsNoObjectToImportError(len(path)) || 
		errors.Is(cErr.GetResultError(pb.RpcObjectImportRequest_Txt), converter.ErrCancel)
```

core/block/import/syncer/icon.go
```diff
@@ -2,17 +2,22 @@ package syncer
...
 	"github.com/anyproto/anytype-heart/core/block/editor/basic"
 	"github.com/anyproto/anytype-heart/core/block/editor/smartblock"
 	"github.com/anyproto/anytype-heart/core/block/simple"
 	"github.com/anyproto/anytype-heart/core/session"
 	"github.com/anyproto/anytype-heart/pb"
+	"github.com/anyproto/anytype-heart/pkg/lib/logging"
+	oserror "github.com/anyproto/anytype-heart/util/os"
 )
 
+var log = logging.Logger("import")
```
KirillSto, line 19: Maybe it is better to make more specific name to this logger, we have two logger with name **import** already

core/block/import/html/converter_test.go
```diff
@@ -3,7 +3,7 @@ package html
 import (
 	"archive/zip"
 	"errors"
-	"io"
+	"github.com/anyproto/anytype-heart/core/block/import/source"
```
KirillSto, line 6: Import order

core/block/import/source/source.go
```diff
@@ -15,7 +15,10 @@ var log = logging.Logger("import-source")
...
+	Iterate(callback func(fileName string, fileReader io.ReadCloser) bool) error
```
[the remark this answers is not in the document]
AnastasiaShemyakinskaya: As far as I understand you, `Iterate(callback func(fileName string, fileReader io.ReadCloser) bool) error` has callback with `bool` returned parameter. So client send flag, is it want to continue iteration.

core/block/import/txt/converter.go
AnastasiaShemyakinskaya: Removed this code and move it to `ShouldAbortImport` function in `ConveterError`

core/block/import/syncer/icon.go
AnastasiaShemyakinskaya: I think it's better to use one name, because it makes search in graylog easier (imho)

core/block/import/source/source.go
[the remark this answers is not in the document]
deff7: I mean `state.Iterate` continues iteration if `true` is returned, your Iterate method stops

core/block/import/importer.go
```diff
@@ -131,7 +131,7 @@ func (i *Import) importFromBuiltinConverter(ctx *session.Context,
 	req *pb.RpcObjectImportRequest,
 	c converter.Converter,
 	progress process.Progress) error {
```
KirillSto, line 133: Maybe we should use the same form of function definition with big amount of arguments/return values, as in https://github.com/anyproto/anytype-heart/pull/383/files#diff-39551f40c77377af86817ac0f6e0fb01b38738e020f21182f71921214b126038R127 ?
This way reader can easily identify the border between func definition and the very function body

core/block/import/importer.go
```diff
@@ -131,7 +131,7 @@ func (i *Import) importFromBuiltinConverter(ctx *session.Context,
...
-	allErrors := converter.NewError()
+	allErrors := converter.NewError(0)
```
KirillSto, line 134: It seems it is better to use explicit value of pb.RpcObjectImportRequest_ALL_OR_NOTHING constant instead of implicit to increase readability. There are 8 places, where NewError is invokated with constant - 3 of them are from tests

deff7 (approved):

KirillSto (approved):

PR #432, opened 2023-09-25
GO-1978 Uninstall relations and types using isUninstall flag

deff7:


core/block/object/objectcreator/creator.go
```diff
@@ -311,71 +311,47 @@ func (w *Creator) createRelationOption(ctx context.Context, spaceID string, deta
...
+	key := uniqueKey.InternalKey()
+
 	object := pbtypes.CopyStruct(details)
 	rawRecommendedLayout := pbtypes.GetInt64(details, bundle.RelationKeyRecommendedLayout.String())
 	recommendedLayout, err := bundle.GetLayout(model.ObjectTypeLayout(int32(rawRecommendedLayout)))
 	if err != nil {
 		return "", nil, fmt.Errorf("invalid recommended layout %d: %w", rawRecommendedLayout, err)
 	}
 
+	var recommendedRelationKeys []string
```
KirillSto, line 347: Do we add recommended relations only from layout, but not from recommendedRelations detail as well?

core/block/import/objectcreator.go
```diff
@@ -197,7 +204,9 @@ func (oc *ObjectCreator) createNewObject(
 			SpaceID:     spaceID,
 		}
 	})
-
+	if errors.Is(err, treestorage.ErrTreeExists) {
```
KirillSto, line 207: Maybe we should use unified format of func header here:
```
func foo(
    arg1,
    arg2,
) return {
```

core/block/object/objectcreator/creator.go
deff7: I discussed it with Razor, clients don't pass recommendedReleations detail, so I decided to remove this code

KirillSto (approved):

PR #435, opened 2023-09-25
GO-2064: File uploader: use simpler heuristic to estimate limit

deff7:


core/filestorage/filesync/upload.go
```diff
@@ -227,17 +243,54 @@ func (f *fileSync) prepareToUpload(ctx context.Context, spaceId string, fileId s
...
+/*
+- dir (root)
+  	- dir (content and meta pair)
+  		- meta
+  		- content (just content if file < 1Mb, and chunks list otherwise)
+			- chunk1
+			- chunk2
+			...
+*/
+func (f *fileSync) estimateFileSize(fileID string) (int, error) {
```
AnastasiaShemyakinskaya, line 272: Maybe to add tests for this functionality, if it's possible?

deff7: It's too hard to write tests for this behavior properly. It's better to invest time on writing tests for `core/files` package. And then comes Files as objects project where we will fix the reason for this size estimation

AnastasiaShemyakinskaya (approved):

PR #445, opened 2023-09-28
GO-2054: sets queries are lost after importing any block

AnastasiaShemyakinskaya:
---

---

### Description

If we make an import between two spaces in account, we get unique from `UniqueKey` relation, no `Id`

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

https://linear.app/anytype/issue/GO-2054/sets-queries-are-lost-after-importing-any-block

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md


core/block/import/objectidgetter.go
```diff
@@ -88,7 +88,11 @@ func (ou *ObjectIDGetter) Get(
 		id := pbtypes.GetString(sn.Snapshot.Data.Details, bundle.RelationKeyId.String())
 		uk, err := domain.UnmarshalUniqueKey(id)
 		if err != nil {
-			return "", treestorage.TreeStorageCreatePayload{}, err
+			id = pbtypes.GetString(sn.Snapshot.Data.Details, bundle.RelationKeyUniqueKey.String())
```
deff7, line 91: Maybe reverse the order of actions? Firstly, check for UniqueKey detail, then try to parse ID. It seems like importing the new version of objects will occur much more often than legacy ones

core/block/export/export.go
```diff
@@ -505,7 +505,9 @@ func validType(sbType smartblock.SmartBlockType) bool {
 		sbType == smartblock.SmartBlockTypeTemplate ||
 		sbType == smartblock.SmartBlockTypeDate ||
 		sbType == smartblock.SmartBlockTypeWorkspace ||
-		sbType == smartblock.SmartBlockTypeWidget
+		sbType == smartblock.SmartBlockTypeWidget ||
+		sbType == smartblock.SmartBlockTypeObjectType ||
+		sbType == smartblock.SmartBlockTypeRelation
```
deff7, line 510: We now have RelationOption as well

deff7 (approved):

PR #533, opened 2023-10-19
GO-2227 change lastModifiedDate on Apply

requilence:


fat-fellow (approved):

core/block/editor/smartblock/smartblock.go
```diff
@@ -686,6 +686,7 @@ func (sb *smartBlock) Apply(s *state.State, flags ...ApplyFlag) (err error) {
 	}
 
 	s.SetLocalDetail(bundle.RelationKeyLastModifiedBy.String(), pbtypes.String(sb.currentProfileId))
+	s.SetLocalDetail(bundle.RelationKeyLastModifiedDate.String(), pbtypes.Int64(lastModified.Unix()))
```
deff7, line 689: Maybe we should change this relation only if user performed some action?

deff7 (changes requested):

core/block/editor/smartblock/smartblock.go
requilence: it's canceled later here https://github.com/anyproto/anytype-heart/blob/go-2227-lastmodifieddate-not-changing/core/block/editor/state/state.go#L636-L652

The logic is pretty complicated but I don't see the way to simplify it. We need to have events BEFORE we push. And we may have local-only changes (e.g. isFavorited or lastOpenedDate details).

deff7 (approved):

PR #543, opened 2023-10-23
GO-1998: add import date to root collection name

AnastasiaShemyakinskaya:
---

---

### Description

Currently common collection with imported files looks like this "Notion Import", "Markdown Import" and etc.
I add date for every new created import collection, so now it looks like this "Notion Import 23.10.2023 16:34:00"

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

https://linear.app/anytype/issue/GO-1998/markdown-import-collection-drops-relation-in-all-view-after-importing

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md


core/block/import/converter/collection.go
```diff
@@ -16,6 +19,8 @@ import (
 	"github.com/anyproto/anytype-heart/util/pbtypes"
 )
 
+const importDateLayout = "02.01.2006 15:04:05"
```
requilence, line 22: better to be 2 Jan 2006, because US folks swap month/dat like and it will be ambiguous 01.02.2006

deff7: so maybe use format from locale?

AnastasiaShemyakinskaya: Well we have different languages in Anytype. So maybe `2 Jan 2006` is not suitable for non-english users

requilence: we can use RFC3339, in this case even the sort by name will be consistent :-)

fat-fellow: I also support RFC3339 for the same reasons

AnastasiaShemyakinskaya: IMHO, RFC3339 looks not really user-friendly) But ok

AnastasiaShemyakinskaya: @requilence please check out

KirillSto (approved):

PR #546, opened 2023-10-23
GO-2246 migration issues

requilence:
- fix global indexer checksum migration (use it in case exist)
- delete old subobjects details from objectstore


deff7 (approved):

pkg/lib/localstore/objectstore/delete.go
```diff
@@ -8,32 +8,37 @@ import (
...
-		} {
-			if err := txn.Delete(k.Bytes()); err != nil {
-				return fmt.Errorf("delete key %s: %w", k, err)
+func (s *dsObjectStore) DeleteDetails(ids ...string) error {
+	for _, chunk := range lo.Chunk(ids, 100) {
+		err := s.updateTxn(func(txn *badger.Txn) error {
+			for _, id := range chunk {
+				s.cache.Del(pagesDetailsBase.ChildString(id).Bytes())
+
+				for _, k := range []ds.Key{
```
deff7, line 25: maybe use longer var names? `key`

PR #569, opened 2023-10-30
GO-2249: add basic snapshot validation

AnastasiaShemyakinskaya:
---

---

### Description

Some users now use PB import to import their own snapshots (for example to test own imports)
Add snapshot validation to avoid errors during its creation. It also can be extensible 

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

https://linear.app/anytype/issue/GO-2249/panic-during-objectimport-in-latest-version
### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/core/block/import/pb/converter.go
+++ b/core/block/import/pb/converter.go
@@ -243,7 +243,7 @@ func (p *Pb) makeSnapshot(name, profileID, path string, file io.ReadCloser, isMi
 	if errGS != nil {
 		return nil, errGS
 	}
-	if valid := p.validateSnapshot(snapshot); !valid {
+	if valid := p.isSnapshotValid(snapshot); !valid {
 		return nil, fmt.Errorf("snapshot is not valid: %s", name)
 	}
 	id := uuid.New().String()
@@ -542,6 +542,6 @@ func (p *Pb) filterObjects(objectTypesToImport widgets.ImportWidgetFlags, object
 	return rootObjects
 }
 
-func (p *Pb) validateSnapshot(snapshot *pb.SnapshotWithType) bool {
+func (p *Pb) isSnapshotValid(snapshot *pb.SnapshotWithType) bool {
 	return !(snapshot == nil || snapshot.Snapshot == nil || snapshot.Snapshot.Data == nil)
 }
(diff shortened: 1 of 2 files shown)


core/block/import/pb/converter.go
```diff
@@ -541,3 +541,7 @@ func (p *Pb) filterObjects(objectTypesToImport widgets.ImportWidgetFlags, object
 	}
 	return rootObjects
 }
+
+func (p *Pb) validateSnapshot(snapshot *pb.SnapshotWithType) bool {
```
KirillSto, line 545: Maybe rename it to `isSnapshotEmpty`?
1. More specific name for check
2. Starting with `is`, so we understand it returns boolean
3. We do not need to perform negation twice

KirillSto (changes requested):

core/block/import/pb/converter.go
AnastasiaShemyakinskaya: I called it validateSnapshot, because we can add other rules in the future

AnastasiaShemyakinskaya: 1. Rename to isSnapshotValid 
2. `We do not need to perform negation twice`. What do you mean?

KirillSto: Inside isSnapshotValid we perform negation on result of disjunction. It is the first time
Second time we perform it on method's return value (!valid)

fat-fellow (approved):

core/block/import/pb/converter.go
AnastasiaShemyakinskaya: It is correct, I suppose
If disjunction is positive, it means, that snapshot is not valid. 
If snapshot is empty, the result of function is false - snapshot is not valid. if !valid, then we return error.
Or may be I missed something?

KirillSto: Yes, it is correct. Just redundant computing :slightly_smiling_face: 
If validation function will grow and will be reused somewhere else, it is ok to leave it as is

KirillSto (approved):

core/block/import/pb/converter.go
AnastasiaShemyakinskaya: I think it's more clear for function, which checks, if snapshot is valid. We can see it from the code, that if snapshot is not valid (!valid), then we return error.

PR #576, opened 2023-10-31
GO-1883: Disable nested subscription, fix dependent subscription

deff7:


core/subscription/service.go
```diff
@@ -175,7 +175,9 @@ func (s *service) subscribeForQuery(req pb.RpcObjectSearchSubscribeRequest, f *d
 		sub.forceSubIds = filterDepIds
 	}
 
-	if withNested, ok := f.FilterObj.(database.WithNestedFilter); ok {
+	// FIXME Nested subscriptions disabled now. We should enable them only by client's request
```
KirillSto, line 178: Maybe we should create task on that?

KirillSto (approved):

PR #594, opened 2023-11-03
GO-2121: static builtin templates do not removed from objectstore

AnastasiaShemyakinskaya:
---

---

### Description

* Extract common struct VirtualSpace to use it further for import preview
* Cleanup static resources from ObjectStore, when virtual space is closed

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [x] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md


space/virtual.go
```diff
@@ -0,0 +1,118 @@
...
+			loadMandatoryObjectsCh: make(chan struct{}),
+		},
+	}
+	return vs
+}
+
+func (vs *VirtualSpace) Close(_ context.Context) error {
+	staticObjects := vs.service.sourceService.GetStaticObjectsBySpaceID(vs.Id())
+	for _, id := range staticObjects {
+		vs.service.sourceService.RemoveStaticSource(id)
```
deff7, line 39: it will work only on graceful shutdown

core/block/source/service.go
```diff
@@ -203,4 +204,21 @@ func (s *service) RemoveStaticSource(id string) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	delete(s.staticIds, id)
+	err := s.objectStore.DeleteDetails(id)
```
deff7, line 207: Maybe it will be more reliable to query all objects with specific space id (i.e. marketplace) and remove them? Or create new key in badger for virtual objects' registry

space/virtual.go
AnastasiaShemyakinskaya: May be we should register virtual spaces in object store and query them on Anytype start? So if virtual space exists in object store, we should check its objects and remove them

core/block/source/service.go
```diff
@@ -203,4 +204,21 @@ func (s *service) RemoveStaticSource(id string) {
...
+	if err != nil {
+		log.Errorf("failed to delete objects details &s, err: %s", id, err)
+	}
+}
+
+func (s *service) GetStaticObjectsBySpaceID(spaceID string) []string {
```
deff7, line 213: seems like it's not used

space/virtual.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/anyproto/any-sync/commonspace/objecttreebuilder"
+	"github.com/anyproto/any-sync/commonspace/spacestorage"
+	"github.com/anyproto/any-sync/commonspace/spacesyncproto"
+	"github.com/anyproto/any-sync/commonspace/syncstatus"
+	"github.com/anyproto/any-sync/net/peer"
+
+	"github.com/anyproto/anytype-heart/pkg/lib/localstore/objectstore"
+)
+
+const name = "virtualSpaceService"
```
deff7, line 22: maybe name it more descriptive?

AnastasiaShemyakinskaya: Hmm, may be `virtualSpaceRegister`?

deff7: I mean const name

AnastasiaShemyakinskaya: afais we usually call it `CName`, is it ok? May be we can call it something like `componentName` or `serviceName`?

deff7: CName is already taken

maybe just virtualSpaceServiceName?

AnastasiaShemyakinskaya: Fixed

deff7 (approved):

PR #610, opened 2023-11-08
GO-1882 creation date and last modified date not set to the created

requilence:
- introduce originalCreatedTime to the state and snapshot. It's purpose is to store the user-side creation time. E.g. when you import the file and want to see the file creation date instead of object tree creation date. I decided to use the real state field instead of one more detail in order to make sure we have a way to make sure that this field is set Only in the first snapshot
- for imported objects now addedDate relation is applied with the underlying objectTree creation date

So the logic is this:
-  in case originalCreatedTime is set in the last snapshot then we use  it for the creationDate relation, and objectTree creation date for relation addedDate
- in case originalCreatedTime is not set, we use objectTree creation date for creationDate relation (like before)

Also this PR fixes the migration of personal space's workspace creationDate into the new spaceView by utilising this new mechanis,


core/block/editor/spaceview.go
```diff
@@ -163,6 +163,18 @@ func (s *SpaceView) SetSpaceData(details *types.Struct) error {
 	}
 
 	if changed {
+		if st.ParentState().ParentState() == nil {
```
deff7, line 166: maybe add check that the first ParentState is not nil?

core/block/source/source.go
```diff
@@ -364,12 +370,13 @@ func (s *source) buildChange(params PushChangeParams) (c *pb.Change) {
...
-				Collections:   params.State.Store(),
-				RelationLinks: params.State.PickRelationLinks(),
-				Key:           params.State.UniqueKeyInternal(),
+				Blocks:                   params.State.BlocksToSave(),
+				Details:                  params.State.Details(),
+				ObjectTypes:              domain.MarshalTypeKeys(params.State.ObjectTypeKeys()),
+				Collections:              params.State.Store(),
+				RelationLinks:            params.State.PickRelationLinks(),
+				Key:                      params.State.UniqueKeyInternal(),
+				OriginalCreatedTimestamp: params.State.OriginalCreatedTimestamp(),
```
deff7, line 379: we have a bunch of places where snapshot is filled (export for example), maybe add this there too?

core/block/editor/spaceview.go
requilence: it's not possible because we do st := s.NewState() 10 lines above :)

core/block/source/source.go
requilence: When we do import we use the createdDate from the details instead. Because it may be both "original date" and "object tree date". Also this way it backward compatible and less logic to handle. At the same time these 2 places may be ambiguous if someone want to operate with snapshots models.

Maybe we can have it unexported?

requilence: Ah, we can't make it unexported here.. Only via creating separate message struct with omitted field only for the export/import, which is too much for this.
I've added the comment to the models.proto

deff7 (approved):

PR #615, opened 2023-11-08
GO-2345 migrate subobjects in realtime changes

requilence:
In case we got some realtime changes from the old pre-migration client they still may reference the subobjects. We need to migrate the state in this case as well.

### DIFF
--- a/core/block/source/source.go
+++ b/core/block/source/source.go
@@ -206,6 +206,10 @@ func (s *source) Update(ot objecttree.ObjectTree) {
 		} else {
 			s.changesSinceSnapshot += sinceSnapshot
 		}
+		// we may receive the update from the pre-migration anytype client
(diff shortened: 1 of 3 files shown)


core/block/source/source.go
```diff
@@ -319,6 +327,9 @@ func (s *source) PushChange(params PushChangeParams) (id string, err error) {
 	}
 	change := s.buildChange(params)
 
+	if s.id == "bafyreigv4y4jcfnxk4cb5iax3ko6mft3tgd5ptdwhnjwyrqbexb6mmgclu" {
```
deff7, line 330: debug code

deff7 (changes requested):

requilence:
closed because this changs can introduce problems on old clients

PR #637, opened 2023-11-14
GO-2390 fix subobject migration

requilence:
- handle wrong bundled objectTypes
- migrate state on Init so it will handle object creation
- migrate state on ResetToVersion so it will handle history version revert operation


core/block/editor/state/change.go
```diff
@@ -293,13 +293,21 @@ func (s *State) changeRelationRemove(rem *pb.ChangeRelationRemove) error {
 	s.RemoveRelation(rem.RelationKey...)
 	return nil
 }
+func migrateObjectTypeIDToKey(old string) (new string, migrated bool) {
```
deff7, line 296: seems like `migrated` is not used

deff7 (approved):

PR #665, opened 2023-11-22
GO-1420: Files as objects

deff7:


mcrakhman (approved):

core/block/object/objectcreator/set.go
```diff
@@ -49,3 +49,18 @@ func (s *service) createSet(ctx context.Context, space clientspace.Space, req *p
 
 	return s.CreateSmartBlockFromStateInSpace(ctx, space, []domain.TypeKey{bundle.TypeKeySet}, newState)
 }
+
+func getDefaultViewRelations(rels []*model.Relation) []*model.BlockContentDataviewRelation {
+	var viewRels = make([]*model.BlockContentDataviewRelation, 0, len(rels))
```
AnastasiaShemyakinskaya, line 54: Seems like var is redundant?

fat-fellow (approved):

core/block/object/objectcreator/set.go
deff7: absolutely, thank you

AnastasiaShemyakinskaya (approved):

PR #672, opened 2023-11-28
GO-2432: implement mechanism to send notification events to client (local notifications)

AnastasiaShemyakinskaya:
---

---

### Description

1. Add pb model Notification with payload for different types of notifications (import, invite and etc)
2. Add Notification events to send them to client
3. Add Notification service, which accept Notification payload, send it to client using event sender and save it to badger notification store
4. Add NotificationList request, which return list of notification depending on given criteria (limit, includeRead)
5. Add NotificationReply request, which update notification status, based on action by client

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [x] 👍 yes

### Added to documentation?

- [ ] 📜 README.md


pkg/lib/localstore/objectstore/notification_store.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/anyproto/anytype-heart/pkg/lib/localstore"
+	"github.com/anyproto/anytype-heart/pkg/lib/pb/model"
+	"github.com/anyproto/anytype-heart/util/badgerhelper"
+)
+
+const notificationsPrefix = "notifications"
+
+var notificationsInfo = ds.NewKey("/" + notificationsPrefix + "/info")
+
+type NotificationStore interface {
```
deff7, line 17: maybe move this to notification service?

deff7: i mean extract notification-related code to service. Methods in objectstore that don't related to objects are legacy

AnastasiaShemyakinskaya: Extract it to separate component

fat-fellow (approved):

core/notifications/notification_store.go
```diff
@@ -0,0 +1,81 @@
...
+
+const (
+	notificationsPrefix   = "notifications"
+	notificationStoreName = "notification_store"
+)
+
+var notificationsInfo = ds.NewKey("/" + notificationsPrefix + "/info")
+
+type NotificationStore interface {
+	app.Component
```
deff7, line 23: seems like notification store could be just an ordinary object without properties of a service, it could be initialized inside notification service

core/notifications/notifications.go
```diff
@@ -0,0 +1,126 @@
...
+	n.eventSender = app.MustComponent[event.Sender](a)
+	return nil
+}
+
+func (n *notificationService) Name() (name string) {
+	return CName
+}
+
+func (n *notificationService) CreateAndSendLocal(notification *model.Notification) error {
+	n.eventSender.Broadcast(&pb.Event{
```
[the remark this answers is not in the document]
deff7: broadcast works locally, it just sends event to all registered sessions

core/notifications/notifications.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/anyproto/anytype-heart/core/event"
+	"github.com/anyproto/anytype-heart/pb"
+	"github.com/anyproto/anytype-heart/pkg/lib/pb/model"
+)
+
+const CName = "notificationService"
+
+type Notifications interface {
+	app.Component
+	CreateAndSendLocal(notification *model.Notification) error
```
deff7, line 17: consider increasing data integrity and encapsulation by assigning id and timestamp to notification inside service

core/notifications/notifications.go
```diff
@@ -0,0 +1,126 @@
...
+	UpdateAndSend(notification *model.Notification) error
```
deff7, line 18: could client code update notification id and timestamp?

AnastasiaShemyakinskaya: - Well, I planned to use it in middleware in case when notification objects is updated. And in this case we didn't update these fields in notification. We can add logic to keep id and timestamp, but now this function is supposed to be used inside middleware in only one case

KirillSto (approved):

PR #679, opened 2023-11-29
GO-2433: add derived notification object for cross device notifications

AnastasiaShemyakinskaya:
---

---

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [x] 👍 yes

### Added to documentation?

- [ ] 📜 README.md


fat-fellow (approved):

core/notifications/notifications.go
```diff
@@ -1,45 +1,135 @@
...
+			}
+			n.updateNotificationsInLocalStore()
+			return
+		}
+	}
+}
+
+func (n *notificationService) updateNotificationsInLocalStore() {
+	var notifications map[string]*model.Notification
+	err := block.DoState(n.picker, n.notificationID, func(s *state.State, sb smartblock.SmartBlock) error {
```
KirillSto, line 92: Maybe it is enough to call `block.Do` generic here? 
State could be retrived from sb.NewState()
`DoState` invokes sb.Apply(s), while `Do` does not

pkg/lib/pb/model/protos/models.proto
```diff
@@ -593,6 +594,7 @@ message Account {
     string timeZone = 104; // time zone from config
     string analyticsId = 105;
     string networkId = 106; // network id to which anytype is connected
+    string notificationId = 107; // network id to which anytype is connected
```
KirillSto, line 597: Comment should be changed

core/notifications/notifications.go
AnastasiaShemyakinskaya: We need to call apply to push changes of notifications

pkg/lib/pb/model/protos/models.proto
AnastasiaShemyakinskaya: remove this parameter, it is not needed

AnastasiaShemyakinskaya: My bad, thank you  - fixed it

core/notifications/notifications.go
AnastasiaShemyakinskaya: My bad, thank you - fixed it

core/block/editor/state/change.go
```diff
@@ -759,6 +781,42 @@ func (s *State) makeOriginalCreatedChanges() (ch []*pb.ChangeContent) {
 	return
 }
 
+func (s *State) makeNotificationChanges() []*pb.ChangeContent {
+	var changes []*pb.ChangeContent
+	if s.parent == nil || s.parent.notifications == nil {
```
deff7, line 786: I think you should use getter for `notifications`, because parent's state could not have initialized `notifications`. Maybe I'm wrong

core/notifications/notifications.go
```diff
@@ -135,3 +236,44 @@ func (n *notificationService) List(limit int64, includeRead bool) ([]*model.Noti
 func (n *notificationService) isNotificationRead(notification *model.Notification) bool {
 	return notification.GetStatus() == model.Notification_Read || notification.GetStatus() == model.Notification_Replied
 }
+
+func (n *notificationService) loadNotificationObject(ctx context.Context) {
+	defer close(n.notificationCh)
```
deff7, line 241: maybe call `n.indexNotifications(notificationContext)` directly here? Channel creates a point of synchronization and linearize order of calling `loadNotificationObject -> indexNotifications`

core/block/editor/state/state.go
```diff
@@ -164,11 +165,11 @@ func (s *State) RootId() string {
 }
 
 func (s *State) NewState() *State {
-	return &State{parent: s, blocks: make(map[string]simple.Block), rootId: s.rootId, noObjectType: s.noObjectType, migrationVersion: s.migrationVersion, uniqueKeyInternal: s.uniqueKeyInternal, originalCreatedTimestamp: s.originalCreatedTimestamp}
+	return &State{parent: s, blocks: make(map[string]simple.Block), rootId: s.rootId, noObjectType: s.noObjectType, migrationVersion: s.migrationVersion, uniqueKeyInternal: s.uniqueKeyInternal, originalCreatedTimestamp: s.originalCreatedTimestamp, notifications: s.CopyNotifications()}
```
deff7, line 168: I think you can optimize things by not copying notifications. If you will use getter for notification so all should be okay. This getter should traverse parents to find the first non-nil notifications map like it did in other parts of state

AnastasiaShemyakinskaya: I think the problem here is that the parent state and the state can refer to the same map, which can lead to unexpected behavior or changes, so that's why I use copy here.

core/block/editor/state/change.go
AnastasiaShemyakinskaya: Fixed

core/notifications/notifications.go
```diff
@@ -40,17 +63,86 @@ func (n *notificationService) Init(a *app.App) (err error) {
...
+func (n *notificationService) Run(_ context.Context) (err error) {
+	notificationContext, notificationCancel := context.WithCancel(context.Background())
+	n.notificationCancel = notificationCancel
+	go n.loadNotificationObject(notificationContext)
+	go n.indexNotifications(notificationContext)
+	return nil
+}
+
+func (n *notificationService) indexNotifications(ctx context.Context) {
+	for {
```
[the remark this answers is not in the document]
AnastasiaShemyakinskaya: Fixed

core/block/editor/state/state.go
deff7: it's common pattern in state, for example:
```
func (s *State) Get(id string) (b simple.Block) {
	if b = s.blocks[id]; b != nil {
		return
	}
	if s.parent != nil {
		if b = s.Pick(id); b != nil {
			b = b.Copy()
			s.blocks[id] = b
			return
		}
	}
	return
}
```

we defer copying to the moment of write

AnastasiaShemyakinskaya: Fixed

core/notifications/notifications.go
```diff
@@ -135,3 +224,44 @@ func (n *notificationService) List(limit int64, includeRead bool) ([]*model.Noti
...
+		return
+	}
+	n.notificationId, err = spc.DeriveObjectID(ctx, uk)
+	if err != nil {
+		log.Errorf("failed to get notification object: %v", err)
+		return
+	}
+	ctxWithPeer := peer.CtxWithPeerId(ctx, peer.CtxResponsiblePeers)
+	_, err = spc.GetObject(ctxWithPeer, n.notificationId)
+	if err != nil && errors.Is(err, net.ErrUnableToConnect) {
```
mcrakhman, line 246: Are you sure that this is what it will return now? I think it is possible that we will just hang here, because we wait forever for some peers to connect (see `GetResponsiblePeers`) 

This can happen especially if we upgrade from prev version and start the app with no connection to nodes. I guess some long timeout will be working here (like 20 sec maybe?)

Also even in this case it is still possible that we will miss the document and it will load. We need to see that we won't get overloaded with a lot of notifications in case the account is large

mcrakhman: But again maybe I overestimate the problem tbh

AnastasiaShemyakinskaya: I rewrite so now it works like this
1. We get tech space
2. We try to derive notification object in tech space
3. If we error TreeAlreadyExist, we try to derive id

I think we have GetObject() call in `DeriveTreeObject`, probably it should be called with timeout?

PR #687, opened 2023-12-04
GO-1938 network config change for existing account

requilence:
Commands changed: AccountSelect, AccountCreate
- NetworkMode field added enum{DefaultConfig, LocalOnly, CustomConfig}.
- In case of CustomConfig NetworkCustomConfigFilePath field is expected
- For AccountSelect, in case you already have account locally but network id is mismatch you will receive the error

Command added: AccountChangeNetworkConfigAndRestart
- It has the same fields: NetworkMode and NetworkCustomConfigFilePath. It can change the param for existing account and restart it
- Make sure to refresh current space because all existing subscriptions will be canceled

We assume that client persist NetworkMode and NetworkCustomConfigFilePath locally and send it on every AccountSelect, AccountCreate  for every account. Even in case the account is already logged in. Explanations below.
- if we log in to the account with new network id we had 2 options
   - do implicit push to another network
   - ignore provided networkId
- in both cases the behaviour is not clear for user and can lead to misunderstanding
- if we will introduce multi-accounts later we can show the list of accounts with the corresponding network. In this case it will be more clear


cmd/grpcserver/anytypeHelperDebug
deff7, line 1: remove

deff7 (approved):

PR #744, opened 2023-12-21
backlinks watcher accumulator

requilence:


core/block/source/source.go
```diff
@@ -353,10 +360,10 @@ func (s *source) PushChange(params PushChangeParams) (id string, err error) {
 	if change.Snapshot != nil {
 		s.lastSnapshotId = id
 		s.changesSinceSnapshot = 0
-		log.Infof("%s: pushed snapshot", s.id)
+		log.Errorf("%s: pushed snapshot", s.id)
```
KirillSto, line 363: Level could be lowered back

core/block/source/source.go
```diff
@@ -353,10 +360,10 @@ func (s *source) PushChange(params PushChangeParams) (id string, err error) {
...
 	} else {
 		s.changesSinceSnapshot++
-		log.Debugf("%s: pushed %d changes", s.id, len(change.Content))
+		log.Errorf("%s: pushed %d changes", s.id, len(change.Content))
```
KirillSto, line 366: Level could be lowered back

KirillSto (approved):

core/block/backlinks/watcher.go
```diff
@@ -67,109 +75,169 @@ func (uw *UpdateWatcher) Close(context.Context) error {
...
+				process()
+				return
+			case <-time.After(aggregationInterval):
+				l.Lock()
+				if time.Since(lastReceivedUpdates) < aggregationInterval || len(accumulatedBacklinks) == 0 {
+					l.Unlock()
+					continue
+				}
+
+				log.Warnf("updating backlinks for %d objects", len(accumulatedBacklinks))
```
mcrakhman, line 150: Fix here warn also pls

core/block/backlinks/watcher.go
```diff
@@ -67,109 +75,169 @@ func (uw *UpdateWatcher) Close(context.Context) error {
...
+		l                    sync.Mutex
+		lastReceivedUpdates  time.Time
+		closedCh             = make(chan struct{})
+		aggregationInterval  = time.Second * 5
+	)
+	defer close(closedCh)
+
+	go func() {
+		process := func() {
+			log.Infof("updating backlinks for %d objects", len(accumulatedBacklinks))
```
mcrakhman, line 130: It should be debug I guess

core/block/backlinks/watcher.go
```diff
@@ -67,109 +75,169 @@ func (uw *UpdateWatcher) Close(context.Context) error {
...
-		for _, id := range modification.ids {
-			err = spc.DoLockedIfNotExists(id, func() error {
-				return uw.store.ModifyObjectDetails(id, modification.modifier)
+	err = spc.DoLockedIfNotExists(id, func() error {
+		return uw.store.ModifyObjectDetails(id, func(details *types.Struct) (*types.Struct, error) {
+			return updateBacklinks(details, backlinksUpdate)
+		})
+	})
+
+	if !errors.Is(err, ocache.ErrExists) {
```
mcrakhman, line 214: If there is no error shouldn't you return from func, i.e. if you applied the changes when the object was not in cache?

KirillSto: BTW yesterday I found many cases when error is DetailsNotChanged. That is absolutely harmless error, so should be considered as normal

PR #775, opened 2024-01-08
Go 2643 fix tz on android and ios

requilence:
https://github.com/golang/go/issues/20455
https://github.com/golang/go/issues/20797

Tested on both iOS and Android. Fixes problems with relative filters, e.g. Today

Android seems to have tzdata on some devices and I was not able to fine a way to read it on iOS. I have embed [tzdata](https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/time/tzdata/tzdata.go) on both of them and it should add 450kb to the binary/heap size


requilence:
We may also later move the responsibility to set the TZ name/offset to clients and save some storage

fat-fellow (approved):

PR #782, opened 2024-01-10
GO-2625: mark relation options uninstall if according relation was deleted

AnastasiaShemyakinskaya:
---

---

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

https://linear.app/anytype/issue/GO-2625/mark-relation-options-uninstall-if-according-relation-was-deleted
### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md


core/block/delete.go
```diff
@@ -48,6 +53,12 @@ func (s *Service) DeleteObjectByFullID(id domain.FullID) (err error) {
 		if err != nil {
 			return fmt.Errorf("on delete: %w", err)
 		}
+		if sbType == coresb.SmartBlockTypeRelation {
+			err := s.deleteRelationOptions(relationKey)
```
deff7, line 57: maybe assign relationKey here?

core/relationoptions/relationoptions.go
```diff
@@ -0,0 +1,98 @@
...
+							Value:       pbtypes.Int64(int64(model.ObjectType_relationOption)),
+						},
+					},
+				})
+				if err != nil {
+					log.Errorf("failed to get options: %s", err)
+					return
+				}
+				for _, option := range relationOptions {
+					err = r.deleter.DeleteObject(option)
```
deff7, line 80: DeleteObject for relation option just set relation isUninstalled=true

AnastasiaShemyakinskaya: AFAIS it also calls `DeleteObject` on `ObjectStore`

deff7: DeleteObject causes relation option to have isUninstalled=true, therefore DeleteObject is already called for the option. So what exactly this code should do?

core/block/editor/page.go
```diff
@@ -72,9 +73,61 @@ func (p *Page) Init(ctx *smartblock.InitContext) (err error) {
...
+	return p.Type() == coresb.SmartBlockTypeRelation &&
+		pbtypes.GetBool(ctx.State.Details(), bundle.RelationKeyIsUninstalled.String())
+}
+
+func (p *Page) deleteRelationOptions(ctx *smartblock.InitContext) error {
+	relationKey := pbtypes.GetString(ctx.State.Details(), bundle.RelationKeyRelationKey.String())
+	relationOptions, _, err := p.objectStore.QueryObjectIDs(database.Query{
+		Filters: []*model.BlockContentDataviewFilter{
+			{
+				RelationKey: bundle.RelationKeyIsArchived.String(),
```
deff7, line 95: maybe ignore isArchived? Restoring from archive will no make sense

core/block/editor/page.go
```diff
@@ -72,9 +73,61 @@ func (p *Page) Init(ctx *smartblock.InitContext) (err error) {
...
+			},
+		},
+	})
+	if err != nil {
+		return err
+	}
+	for _, id := range relationOptions {
+		err = p.Space().Do(id, func(b smartblock.SmartBlock) error {
+			st := b.NewState()
+			st.SetDetailAndBundledRelation(bundle.RelationKeyIsUninstalled, pbtypes.Bool(true))
```
deff7, line 122: Call deleter.DeleteObject instead

core/relationoptions/relationoptions.go
AnastasiaShemyakinskaya: It should remove options, which already exist in account, but their relation was already deleted. So this logic is for already deleted relations in account. I think we need, because we have a case, when user have around 700 not needed relation options in account, but relations are not existed

deff7: 1) you should fix query then
2) it's not necessary to do this periodically, seems like executing of this routine on service start is enough

AnastasiaShemyakinskaya: Discussed

core/block/editor/page.go
AnastasiaShemyakinskaya: Fix

core/block/editor/page.go
AnastasiaShemyakinskaya: Fixed

deff7 (approved):

PR #788, opened 2024-01-11
GO-2658 add syncer, switch ristretto

requilence:
- switch badger to include corrupted db check (with indexLen = 0 ) https://github.com/anyproto/badger/tree/fix-corrupted-db-checksum-check
- switch to ristretto fork to have msync impl on windows  https://github.com/anyproto/ristretto/tree/msync-win
- add regular mmap sync with configurable interval of inactivity (every 1 min)
- add fallback every 500 db versions (in case of active db writes)
- remove and rebuild localstore db(not spacestore) in case of checksum errors

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,6 +31,7 @@ require (
 	github.com/gogo/status v1.1.1
 	github.com/golang-jwt/jwt v3.2.2+incompatible
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da
+	github.com/golang/mock v1.6.0
 	github.com/golang/snappy v0.0.4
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -31,6 +31,7 @@ require (
 	github.com/gogo/status v1.1.1
 	github.com/golang-jwt/jwt v3.2.2+incompatible
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da
+	github.com/golang/mock v1.6.0
```
deff7, line 34: we use uber's implementation

mcrakhman (approved):

PR #791, opened 2024-01-13
Set SbType in importFromExternalSource

balazsgrill:
---

---

### Description

[`importFromExternalSource`](https://github.com/anyproto/anytype-heart/blob/main/core/block/import/importer.go#L186C13-L186C13) leaves `SbType` unsetted in the created Snapshot instances, which is checked in [`IDProvider.GetIDAndPayload`](https://github.com/anyproto/anytype-heart/blob/main/core/block/import/common/objectid/provider.go#L48C20-L48C35), causing it to always fail with "unsupported smartblock to import".

I added a non-null SbType value, and managed to use the external import API, although I'm unsure about the semantics of this field and its relation to the Object type set in the snapshot contents.

The [test of this function](https://github.com/anyproto/anytype-heart/blob/main/core/block/import/importer_test.go#L250C21-L250C21) doesn't catch this issue as it uses a mock ID provider without such check.

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [ ] 👍 yes
- [x] 🙋 no, because I need help

### Added to documentation?

- [ ] 📜 README.md


balazsgrill:

PR #820, opened 2024-01-22
Files as objects: fix export-import

deff7:


core/block/import/common/common.go
```diff
@@ -163,6 +166,15 @@ func handleLinkBlock(oldIDtoNew map[string]string, block simple.Block, st *state
 	st.Set(simple.New(block.Model()))
 }
 
+func handleFileBlock(oldIDtoNew map[string]string, block simple.Block, st *state.State) {
```
fat-fellow, line 169: lowercase id

PR #824, opened 2024-01-22
GO-2583: fix some bugs

AnastasiaShemyakinskaya:
---

---

### Description

1. In case of exporting one object, we take its files objects from object store and save them with object, so files can be restored during export
2. Fix saving file keys: before we use function from state.Doc interface, now we use function from smartBlock interface
3. In case of non-anytype export, just save files for backup purpose

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md


core/block/export/export.go
```diff
@@ -451,6 +526,39 @@ func (e *export) handleFileObject(ctx context.Context, keys *files.FileKeys, spa
...
+		if relLink.Format == model.RelationFormat_file {
+			if relLink.GetKey() == bundle.RelationKeyCoverId.String() {
+				v := pbtypes.GetString(st.Details(), relLink.GetKey())
+				_, err := cid.Decode(v)
+				if err != nil {
+					// this is an exception cause coverId can contain not a file hash but color
+					continue
+				}
+			}
+			if fileHash := pbtypes.GetString(st.Details(), relLink.GetKey()); fileHash != "" {
```
deff7, line 550: GetStringList covers the case with String value, it returns `[]string{stringValue}`

core/block/export/export.go
```diff
@@ -451,6 +526,40 @@ func (e *export) handleFileObject(ctx context.Context, keys *files.FileKeys, spa
...
+		if _, werr := e.saveFile(ctx, wr, fh); werr != nil {
+			log.With("hash", fh).Warnf("can't save file: %v", werr)
+		}
+	}
+}
+
+func (e *export) getFilesFromRelations(st *state.State, fileHashes []string) []string {
+	for _, relLink := range st.GetRelationLinks() {
+		if relLink.Format == model.RelationFormat_file {
+			if e.isCoverRelationColor(st, relLink) {
```
deff7, line 542: coverId has RelationFormat_longtext, you should check it independent of file relation format

core/block/import/common/objectcreator/objectcreator.go
```diff
@@ -525,6 +528,19 @@ func (oc *ObjectCreator) setFileImportedFlagAndOrigin(st *state.State, origin mo
...
+func (oc *ObjectCreator) saveFileKeys(st *state.State, newHash string, oldHash string) bool {
+	fileKeys, err := oc.fileStore.GetFileKeys(newHash)
+	if err != nil {
+		return true
+	}
+	st.AddFileKeys(&pb.ChangeFileKeys{
+		Hash: newHash,
+		Keys: fileKeys,
+	})
+	st.RemoveFileKeys(oldHash)
```
deff7, line 540: maybe don't delete keys? Just in case

core/block/export/export.go

deff7 (approved):

PR #852, opened 2024-02-01
GO-2668 Add details of objects in dataview filters

KirillSto:
https://linear.app/anytype/issue/GO-2668/collections-filters-are-not-working

Objects set in filters of Collection views were not shown up, because desktop client did not recieve their details on ObjectSearchSubscribe.

This PR contains two fixes: 

1. For desktop, that takes details from dependencies in ObjecSearchSubscribe response. Collection subscription was not taking filterObjectIds into account, while initialising dependency subscription.
2. For mobile clients, that take details from smartblock view on ObjectOpen. We did not analyze filters on collecting dependent ids from dataview block.

### DIFF
--- a/core/block/simple/dataview/dataview.go
+++ b/core/block/simple/dataview/dataview.go
@@ -221,6 +221,33 @@ func (l *Dataview) FillSmartIds(ids []string) []string {
 		if view.DefaultTemplateId != "" {
 			ids = append(ids, view.DefaultTemplateId)
 		}
+		ids = append(ids, getIdsFromFilters(view.Filters)...)
+	}
+
+	return ids
+}
+
+func getIdsFromFilters(filters []*model.BlockContentDataviewFilter) (ids []string) {
+	for _, filter := range filters {
+		if filter.Format != model.RelationFormat_object &&
+			filter.Format != model.RelationFormat_status &&
+			filter.Format != model.RelationFormat_tag {
+			continue
+		}
+
+		id := filter.Value.GetStringValue()
+		if id != "" {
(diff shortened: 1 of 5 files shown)


core/block/simple/dataview/views.go
```diff
@@ -207,3 +209,11 @@ func (l *Dataview) ReorderViewRelations(viewID string, relationKeys []string) er
 	}
 	return nil
 }
+
+func (l *Dataview) setRelationFormat(filter *model.BlockContentDataviewFilter) {
```
KirillSto, line 213: We are setting relation format explicitly, because format is not obligatory field in protocol and longtext=0 was set by default

core/block/simple/dataview/dataview.go
```diff
@@ -221,6 +221,33 @@ func (l *Dataview) FillSmartIds(ids []string) []string {
...
 		}
+		ids = append(ids, getIdsFromFilters(view.Filters)...)
+	}
+
+	return ids
+}
+
+func getIdsFromFilters(filters []*model.BlockContentDataviewFilter) (ids []string) {
+	for _, filter := range filters {
+		if filter.Format != model.RelationFormat_object &&
```
deff7, line 232: do we need file objects here? (and special RelationKeyCoverId)

deff7 (approved):

core/block/simple/dataview/dataview.go
KirillSto: It seems our clients do not support filters by relations with File format, so we cannot meet such objects in dataview block

PR #853, opened 2024-02-01
GO-2489: Multiplayer

deff7:


space/internal/spaceprocess/joiner/statuschanger.go
```diff
@@ -0,0 +1,39 @@
...
+	s.status = a.MustComponent(spacestatus.CName).(spacestatus.SpaceStatus)
+	return nil
+}
+
+func (s *statusChanger) Name() (name string) {
+	return statusChangerCName
+}
+
+func (s *statusChanger) Run(ctx context.Context) (err error) {
+	s.status.Lock()
```
deff7, line 32: why create separate component only to do one thing that could be done in?

```
func (i *joiner) Start(ctx context.Context) error {
	return i.app.Start(ctx)
}
```

core/identity/identity.go
```diff
@@ -218,30 +244,321 @@ func (s *service) runLocalProfileSubscriptions(ctx context.Context) (err error)
...
+}
+
+func (s *service) indexIconImage(profile *model.IdentityProfile) error {
+	if len(profile.IconEncryptionKeys) > 0 {
+		return s.fileAclService.StoreFileKeys(domain.FileId(profile.IconCid), profile.IconEncryptionKeys)
+	}
+	return nil
+}
+
+func (s *service) handleIdentityData(identityData *identityrepoproto.DataWithIdentity) error {
```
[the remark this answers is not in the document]
deff7: sure, thanks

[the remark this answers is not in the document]
deff7: divided and gave more descriptive name

core/block/editor/participant.go
```diff
@@ -0,0 +1,47 @@
...
+	ctx.State.SetDetailAndBundledRelation(bundle.RelationKeyIsArchived, pbtypes.Bool(false))
+	ctx.State.SetDetailAndBundledRelation(bundle.RelationKeyIsHidden, pbtypes.Bool(false))
+	ctx.State.SetDetailAndBundledRelation(bundle.RelationKeyLayout, pbtypes.Int64(int64(model.ObjectType_participant)))
+	ctx.State.SetDetailAndBundledRelation(bundle.RelationKeyLayoutAlign, pbtypes.Int64(int64(model.Block_AlignCenter)))
+
+	template.InitTemplate(ctx.State,
+		template.WithEmpty,
+		template.WithTitle,
+		template.WithFeaturedRelations,
+		template.WithAddedFeaturedRelation(bundle.RelationKeyType),
```
requilence, line 40: we can include backlinks?

core/acl/aclservice.go
```diff
@@ -0,0 +1,477 @@
...
+	err = cl.ChangePermissions(ctx, list.PermissionChangesPayload{
+		Changes: listPerms,
+	})
+	if err != nil {
+		return fmt.Errorf("%w, %w", ErrAclRequestFailed, err)
+	}
+	return nil
+}
+
+func (a *aclService) Exit(ctx context.Context, spaceId string) error {
```
requilence, line 179: Maybe Join/Leave can be a better naming pair?

core/acl/aclservice.go
```diff
@@ -0,0 +1,477 @@
...
+			ReadKey:     crypto.NewAES(),
+		},
+	})
+	if err != nil {
+		return fmt.Errorf("%w, %w", ErrAclRequestFailed, err)
+	}
+	return nil
+}
+
+func (a *aclService) Cancel(ctx context.Context, spaceId string) (err error) {
```
requilence, line 109: CancelJoin?

core/acl/aclservice.go
```diff
@@ -0,0 +1,478 @@
...
+		return err
+	}
+	acl := acceptSpace.CommonSpace().Acl()
+	acl.RLock()
+	recs, err := acl.AclState().JoinRecords(false)
+	if err != nil {
+		acl.RUnlock()
+		return err
+	}
+	// TODO: change this logic to use RequestJoin objects
```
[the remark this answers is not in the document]
requilence: from mc: we can delete this todo

core/acl/aclservice.go
```diff
@@ -0,0 +1,477 @@
...
+			iconCid, iconEncryptionKeys, err := a.fileAcl.GetInfoForFileSharing(ctx, iconObjectId)
+			if err != nil {
+				return fmt.Errorf("get icon info: %w", err)
+			}
+			invitePayload.SpaceIconCid = iconCid
+			invitePayload.SpaceIconEncryptionKeys = iconEncryptionKeys
+		}
+		return nil
+	})
+	if err != nil {
```
requilence, line 363: what if icon broken and not possible to load? maybe we can create the invite without it in this case (just an idea, need confirmation)

core/block/service.go
```diff
@@ -205,7 +205,7 @@ func (s *Service) OpenBlock(sctx session.Context, id domain.FullID, includeRelat
 		st := ob.NewState()
 
 		st.SetLocalDetail(bundle.RelationKeyLastOpenedDate.String(), pbtypes.Int64(time.Now().Unix()))
-		if err = ob.Apply(st, smartblock.NoHistory, smartblock.NoEvent, smartblock.SkipIfNoChanges, smartblock.KeepInternalFlags); err != nil {
+		if err = ob.Apply(st, smartblock.NoHistory, smartblock.NoEvent, smartblock.SkipIfNoChanges, smartblock.KeepInternalFlags, smartblock.IgnoreNoPermissions); err != nil {
```
requilence, line 208: need to add it also in ShowBlock

core/block/source/service.go
```diff
@@ -133,6 +136,19 @@ func (s *service) newSource(ctx context.Context, space Space, id string, buildOp
 			return NewBundledRelation(id), nil
 		case smartblock.SmartBlockTypeIdentity:
 			return NewIdentity(s.identityService, id), nil
+		case smartblock.SmartBlockTypeParticipant:
+			participantState := state.NewDoc(id, nil).(*state.State)
+			participantState.SetObjectTypeKey(bundle.TypeKeyParticipant)
```
requilence, line 141: why not in editor/participant.go? just interesting

core/indexer/indexer.go
```diff
@@ -110,6 +114,33 @@ func (i *indexer) Close(ctx context.Context) (err error) {
...
+		Filters: []*model.BlockContentDataviewFilter{
+			{
+				RelationKey: bundle.RelationKeySpaceId.String(),
+				Condition:   model.BlockContentDataviewFilter_Equal,
+				Value:       pbtypes.String(spaceId),
+			},
+			{
+				RelationKey: bundle.RelationKeyType.String(),
+				Condition:   model.BlockContentDataviewFilter_Equal,
+				Value:       pbtypes.String(bundle.TypeKeyParticipant.String()),
```
requilence, line 128: we store type id here, not type key. easier to compare with layout participant here

requilence (approved):

space/internal/spaceprocess/joiner/statuschanger.go
mcrakhman: No this couldn't be done like this, because this status thing might not be initialised at that point (it sets techSpace on init). I could've done it better by presumably making status a component with empty init and initialising it via constructor essentially. But yeah now it works like this :-)

mcrakhman: So you are right that this is weird, I will move techspace to a status constructor init at some point, just need to check that this doesn't break anything

core/block/source/service.go
deff7: because of deriving type id, I'll write a comment

core/block/editor/participant.go
deff7: I'll try

core/acl/aclservice.go
deff7: sure

deff7: sure, but then we have to rename api method

core/acl/aclservice.go
```diff
@@ -0,0 +1,477 @@
...
+func (a *aclService) Remove(ctx context.Context, spaceId string, identities []crypto.PubKey) error {
+	removeSpace, err := a.spaceService.Get(ctx, spaceId)
+	if err != nil {
+		return err
+	}
+	newPrivKey, _, err := crypto.GenerateRandomEd25519KeyPair()
+	if err != nil {
+		return err
+	}
+	cl := removeSpace.CommonSpace().AclClient()
```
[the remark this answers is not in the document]
deff7: added todo

core/acl/aclservice.go
deff7: done

core/acl/aclservice.go
deff7: fixed, thanks

core/block/service.go
deff7: ShowBlock doesn't call Apply

core/indexer/indexer.go
deff7: fixed

PR #896, opened 2024-02-11
GO-2516: create notifications for space invitation

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-2516/create-notifications-for-space-invitation

1. Add notification for owner, who get request to join space. (function `sendJoinRequest`)
2. Add notification for approvw/decline join request in `SpaceRequestApprove`/`SpaceRequestDecline`
3. Add notification for user, who get approved by owner (function `sendParticipantRequestApprove`)
4. Add function `SendNotification`, which works with acl notifications. The algorithm is:
* In case of account starting we iterate through all Acl records and send according notifications
* In case we have last notification id for given Acl in cache, we iterate Acl records from particular id
* In case we update acl in real time, we just look at last record (Acl.Head())


core/block/editor/state/state.go
```diff
@@ -1837,6 +1846,45 @@ func (s *State) AddBundledRelations(keys ...domain.RelationKey) {
...
+		for _, n := range s.parent.ListNotifications() {
+			if _, ok := s.notifications[n.Id]; !ok {
+				s.notifications[n.Id] = pbtypes.CopyNotification(n)
+			}
+		}
+	}
+	s.notifications[notification.Id] = notification
+}
+
+func (s *State) ListNotifications() map[string]*model.Notification {
```
fat-fellow, line 1877: Seems like possible DRY with  GetNotificationById

util/pbtypes/copy.go
```diff
@@ -158,3 +161,24 @@ func CopyFilter(in *model.BlockContentDataviewFilter) (out *model.BlockContentDa
 	bytesPool.Put(buf)
 	return
 }
+
+func CopyNotification(in *model.Notification) (out *model.Notification) {
```
fat-fellow, line 165: dunno if it is possible to have a generic version of the method, don't do if too tedious

core/identity/identity.go
```diff
@@ -89,7 +89,8 @@ type service struct {
 	// identity => spaceId => observer
 	identityObservers      map[string]map[string]*observer
 	identityEncryptionKeys map[string]crypto.SymKey
-	identityProfileCache   map[string]*model.IdentityProfile
+	sync.RWMutex
```
deff7, line 92: there is `lock` already

core/identity/identity.go
```diff
@@ -262,6 +263,22 @@ func (s *service) GetMyProfileDetails() (identity string, metadataKey crypto.Sym
 	return s.myIdentity, s.spaceService.AccountMetadataSymKey(), s.currentProfileDetails
 }
 
+func (s *service) WaitProfile(ctx context.Context, identity string) *model.IdentityProfile {
+	for {
```
deff7, line 267: you should use ticker here, also add case to select for s.closing

core/identity/identity.go
```diff
@@ -262,6 +263,22 @@ func (s *service) GetMyProfileDetails() (identity string, metadataKey crypto.Sym
...
+		select {
+		case <-ctx.Done():
+			return nil
+		default:
+			s.RLock()
```
deff7, line 272: s.lock.RLock()

core/identity/identity.go
```diff
@@ -456,7 +473,9 @@ func (s *service) broadcastIdentityProfile(identityData *identityrepoproto.DataW
 		return fmt.Errorf("find profile: %w", err)
 	}
 
+	s.RLock()
```
deff7, line 476: no need to lock here, it's already under the lock

space/internal/components/aclobjectmanager/aclobjectmanager.go
```diff
@@ -281,22 +303,22 @@ func (a *aclObjectManager) updateParticipantFromIdentity(ctx context.Context, id
...
-						Identity:     identity,
-						IdentityName: profile.Name,
-						IdentityIcon: profile.IconCid,
-					},
-				},
-			})
-			if err != nil {
-				return nil, err
-			}
+			// err := a.notificationService.(&model.Notification{
```
deff7, line 306: clean up

space/internal/components/aclnotifications/notifications.go
```diff
@@ -0,0 +1,197 @@
...
+		Space: space.Id(),
+		Acl:   aclId,
+	})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (n *AclNotificationSender) getProfileData(ctx context.Context, reqJoin *aclrecordproto.AclAccountRequestJoin,
```
deff7, line 169: you could return *model.IdentityProfile, it contains pubkey.Account() already

pkg/lib/pb/model/protos/models.proto
```diff
@@ -962,8 +964,11 @@ message Notification {
         Export export = 8;
         GalleryImport galleryImport = 9;
         RequestToJoin requestToJoin = 10;
+        RequestResponse requestResponse = 11;
+        ParticipantRequestApproved participantRequestApproved = 12;
     }
     string space = 7;
+    string acl = 13;
```
deff7, line 971: what is the purpose of this field?

AnastasiaShemyakinskaya: So we can use for `lastNotificationByAcl` in notification service for indexing purpose. So we don't need to iterate each time from the beginning `Acl.Records()`

core/identity/identity.go
AnastasiaShemyakinskaya: Thank you, didn't notice!

pkg/lib/pb/model/protos/models.proto
deff7: maybe name it `AclHeadId`? And in other places too

AnastasiaShemyakinskaya: Do you mean rename `acl` field in `Notificaion` model?

deff7: yes, it's more descriptive

space/internal/components/aclnotifications/notifications.go
AnastasiaShemyakinskaya: I suppose, `WaitProfile` function can return nil profile

space/internal/components/aclnotifications/notifications.go
```diff
@@ -0,0 +1,229 @@
...
+		AclHeadId: aclId,
+	})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (n *AclNotificationSender) getProfileData(ctx context.Context, account string) *model.IdentityProfile {
+	ctxWithTimeout, _ := context.WithTimeout(ctx, time.Second*30)
```
deff7, line 182: I think we can't show notification without participant name, so we need to wait it until it's loaded

core/block/editor/smartblock/smartblock.go
```diff
@@ -699,7 +699,7 @@ func (sb *smartBlock) Apply(s *state.State, flags ...ApplyFlag) (err error) {
 				sb.undo.Add(act)
 			}
 		}
-	} else if hasStoreChanges(changes) || migrationVersionUpdated { // TODO: change to len(changes) > 0
+	} else if hasStoreChanges(changes) || migrationVersionUpdated || hasNotificationChanges(changes) { // TODO: change to len(changes) > 0
```
mcrakhman, line 702: Do we need two functions here btw? Or they are used only here? If that is the case we can have only function hasChanges

core/block/editor/state/change.go
```diff
@@ -433,6 +437,26 @@ func (s *State) changeOriginalCreatedTimestampSet(set *pb.ChangeOriginalCreatedT
...
+	if s.notifications == nil {
+		s.notifications = map[string]*model.Notification{}
+	}
+	if _, ok := s.notifications[notification.Id]; ok {
+		return
+	}
+	s.notifications[notification.Id] = notification
+}
+
+func (s *State) updateNotification(update *pb.ChangeNotificationUpdate) {
```
mcrakhman, line 450: Can there be a case where we incorrectly update status, where for example the changes has been made concurrently? What I mean here is that imagine some change with status "Unread" comes after "Read" and marks the notification as not seen. Or this can't happen?

core/notifications/notifications.go
```diff
@@ -40,17 +65,92 @@ func (n *notificationService) Init(a *app.App) (err error) {
...
+
+func (n *notificationService) Close(_ context.Context) (err error) {
+	if n.notificationCancel != nil {
+		n.notificationCancel()
+	}
+	return nil
+}
+
+func (n *notificationService) CreateAndSend(notification *model.Notification) error {
+	if notification.Id == "" {
```
mcrakhman, line 129: Is this a valid scenario? I mean I would probably just return an error and make choosing an id a responsibility of whoever created the notification

core/notifications/notifications.go
```diff
@@ -132,6 +241,55 @@ func (n *notificationService) List(limit int64, includeRead bool) ([]*model.Noti
...
 	return notification.GetStatus() == model.Notification_Read || notification.GetStatus() == model.Notification_Replied
 }
+
+func (n *notificationService) loadNotificationObject(ctx context.Context) {
+	uk, err := domain.NewUniqueKey(sb.SmartBlockTypeNotificationObject, "")
+	if err != nil {
+		log.Errorf("failed to get notification object unique key: %v", err)
+		return
+	}
+	techSpaceID, err := n.spaceCore.DeriveID(ctx, spacecore.TechSpaceType)
```
mcrakhman, line 260: I guess we shouldn't get tech space like this, maybe it is better to expose a method that returns a tech space from the space service (if it is not exposed yet).

core/block/editor/state/change.go
mcrakhman: If this is the case we may prevent changing status from "Read" back to "Shown" etc

AnastasiaShemyakinskaya: I think, it's not possible for acl notifications, because we add notification in NotificationObject only once. So it can't be Unread again.

deff7 (approved):

mcrakhman (approved):

PR #907, opened 2024-02-15
GO-2797: make error about wrong JSON/pb format in PB import universal

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-2797/make-error-about-wrong-jsonpb-format-in-pb-import-universal

Currently we have a lot of errors in sentry, when people try to import JSON/Pb file, which doesn't match our `Snapshot` format. So I add more general error for such cases to avoid lot of similar errors in Sentry


core/block/import/pb/converter.go
```diff
@@ -288,18 +289,18 @@ func (p *Pb) getSnapshotFromFile(rd io.ReadCloser, name string) (*pb.SnapshotWit
...
+			return nil, ErrWrongFormat
 		}
 		return snapshot, nil
 	}
 	if filepath.Ext(name) == ".pb" {
 		snapshot := &pb.SnapshotWithType{}
 		data, err := io.ReadAll(rd)
 		if err != nil {
-			return nil, fmt.Errorf("PB:GetSnapshot %w", err)
+			return nil, ErrWrongFormat
```
KirillSto, line 300: It seems that an error on reading bytes from file is not connected with wrong format of PB file. I could be wrong :see_no_evil:

KirillSto (approved):

PR #912, opened 2024-02-15
GO-1394 Add new focused block id in response

KirillSto:
https://linear.app/anytype/issue/GO-1394/change-the-logic-for-saving-cursor-focus-on-paste

This PR include two fixes. 

1. First - on BlockCopy of range of blocks, where first one is text block. Text from this block was empty in response. 
The reason - special logic on copying range of text symbols of first text block. We did not check if client provides zero rangeTo and rangeFrom in request.
The fix - I have added check on rangeTo and rangeFrom. If these parameters are both equal to zero, than we are facing multirange copy and should copy entire text block as is.

2. Second bug was connected with cursor loss on pasting a bunch of text blocks. Previous solution used to substitute id of old block to last inserted text block. But it was bad solution.
The fix is pretty simple. The client already has logic on selecting block to put focus on. It takes last block returned in blockIds field of BlockPaste response. So I have just added this block id to response


core/block/editor/clipboard/clipboard_test.go
```diff
@@ -149,11 +149,11 @@ func TestCommonSmart_pasteAny(t *testing.T) {
...
 	})
 
-	t.Run("9. Save id of focused block", func(t *testing.T) {
+	t.Run("9. Return id of focused block", func(t *testing.T) {
 		sb := createPage(t, createBlocks([]string{}, []string{"11111", "22222", "33333", "44444", "55555"}, emptyMarks))
-		pasteAny(t, sb, "4", model.Range{}, []string{}, createBlocks([]string{"new1", "new2"}, []string{"aaaaa", "bbbbb"}, emptyMarks))
+		ids := pasteAny(t, sb, "4", model.Range{}, []string{}, createBlocks([]string{"new1", "new2"}, []string{"aaaaa", "bbbbb"}, emptyMarks))
 		checkBlockText(t, sb, []string{"11111", "22222", "33333", "aaaaa", "bbbbb", "44444", "55555"})
-		assert.Equal(t, sb.Blocks()[5].Id, "4")
+		assert.NotEmpty(t, ids)
```
deff7, line 156: what does this test actually check?

KirillSto: I replaced this test with another one. New unit test also checks length of returned list of new blocks. Last block in this list in case isSameBlockCaret=false will be chosen as focused on client side

fat-fellow (approved):

PR #930, opened 2024-02-20
GO-2889 Set UniqueKyes to relation options

KirillSto:
https://linear.app/anytype/issue/GO-2889/default-status-options-are-duplicated-after-import

We should set similar ids and uniqueKeys to all realtion options in our usecases


ra3orblade:
I do not understand why do you change only usecases, when user creates an option and makes export/import it gets duplicated as well as I suppose.

KirillSto:
> I do not understand why do you change only usecases, when user creates an option and makes export/import it gets duplicated as well as I suppose.

If user imports option in the same space, it is not duplicated, because options are derived via uniqueKey. The bug was connected with import of default tag and status options because they had different ids and no unique keys set

fat-fellow (approved):

PR #943, opened 2024-02-21
GO-2956 add link block to the profile object in the participant object

requilence:
as a temp measure before the next release when we will migrate the profile object blocks into the separate object


requilence:
discarded in favor of #952

PR #948, opened 2024-02-22
GO-2673 Improve UsecaseValidation tool

KirillSto:
https://linear.app/anytype/issue/GO-2673/implement-validator-on-usecase-addition-in-gallery

- Add flag to specify spaceDashboardId in profile
- Fix treatment of .pb.json suffix
- Add relaton links from lib


cmd/usecasevalidator/main.go
```diff
@@ -164,6 +164,7 @@ func getFlags() (*cliFlags, error) {
 	rules := flag.String("rules", "", "Path to file with processing rules")
 	exclude := flag.Bool("exclude", false, "Exclude objects that did not pass validation")
 	custom := flag.Bool("c", false, "Collect usage information about custom types and relations")
+	spaceDBId := flag.String("s", "", "Id of object to be set as Space Dashboard")
```
AnastasiaShemyakinskaya, line 167: May be use the full spaceDashBoardId? We use camel case afaiu. And DB is usually associated with database I guess.

KirillSto: Yeah, thank you for mentioning that. Fixed!

PR #950, opened 2024-02-22
GO-2892: not export set or collections during markdown import

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-2892/set-of-any-type-become-page-after-export-of-a-space-and-import-to

I decided to not export `Sets` or `Collections` during markdown import, as they are imported as `Page` object type during `Markdown` import. We can't get information about their object type during `Markdown` import as markdown file doesn't contain information about object type or layout. So it can confuse users. 
Also they doesn't contain useful information in markdown file (it's empty for sets or collections).


core/block/export/export.go
```diff
@@ -608,12 +608,20 @@ func validType(sbType smartblock.SmartBlockType) bool {
...
+func validTypeForNonProtobuf(sbType smartblock.SmartBlockType, details *types.Struct) bool {
+	if !validLayout(details) {
+		return false
+	}
 	return sbType == smartblock.SmartBlockTypeProfilePage ||
 		sbType == smartblock.SmartBlockTypePage ||
 		sbType == smartblock.SmartBlockTypeFileObject
 }
 
+func validLayout(details *types.Struct) bool {
```
KirillSto, line 620: Maybe we should rename this func as validLayoutForNonProtobuf and move its invokation to line 541:

`if !isProtobuf && !validTypeForNonProtobuf(sbType) && !validLayoutForNonProtobuf(r.Details) {`

KirillSto (approved):

PR #952, opened 2024-02-23
GO-2945 add old profile migration

requilence:
we deprecate the ability to store any custom state inside the personal space's derived Profile object. From now on it only supposed to store name/image.

Because some user may have some data within it this migration extract this into the new object. But only in case there are any non-empty blocks within it.

### DIFF
--- a/core/block/editor/profile.go
+++ b/core/block/editor/profile.go
@@ -82,10 +82,7 @@ func (p *Profile) CreationStateMigration(ctx *smartblock.InitContext) migration.
 			template.InitTemplate(st,
 				template.WithObjectTypesAndLayout([]domain.TypeKey{bundle.TypeKeyProfile}, model.ObjectType_profile),
 				template.WithDetail(bundle.RelationKeyLayoutAlign, pbtypes.Float64(float64(model.Block_AlignCenter))),
-				template.WithTitle,
-				template.WithFeaturedRelations,
 				template.WithRequiredRelations(),
-				migrationWithIdentityBlock,
 				migrationSetHidden,
 			)
 		},
@@ -94,6 +91,28 @@ func (p *Profile) CreationStateMigration(ctx *smartblock.InitContext) migration.
 
 func migrationSetHidden(st *state.State) {
 	st.SetDetail(bundle.RelationKeyIsHidden.String(), pbtypes.Bool(true))
+}
+
+func migrationExtractBlocksIfExists() template.StateTransformer {
+	return func(st *state.State) {
+		blockId := "identity"
+		st.Set(simple.New(&model.Block{
+			Id: blockId,
+			Content: &model.BlockContentOfRelation{
+				Relation: &model.BlockContentRelation{
+					Key: bundle.RelationKeyProfileOwnerIdentity.String(),
+				},
+			},
+			Restrictions: &model.BlockRestrictions{
+				Edit:   true,
+				Remove: true,
+				Drag:   true,
+				DropOn: true,
+			},
+		}))
+
+		st.InsertTo(state.TitleBlockID, model.Block_Bottom, blockId)
+	}
 }
 
 func migrationWithIdentityBlock(st *state.State) {
@@ -125,6 +144,12 @@ func (p *Profile) StateMigrations() migration.Migrations {
 		{
 			Version: 3,
 			Proc:    migrationSetHidden,
+		},
+		{
+			Version: 4,
+			Proc: func(s *state.State) {
+				// skip it. it will be done separately via ProfileMigrationExtractUserStat in space.migrationProfileObject
+			},
 		},
 	})
 }
--- a/core/block/editor/profile.go
+++ b/core/block/editor/profile.go
@@ -93,28 +93,6 @@ func migrationSetHidden(st *state.State) {
 	st.SetDetail(bundle.RelationKeyIsHidden.String(), pbtypes.Bool(true))
 }
 
-func migrationExtractBlocksIfExists() template.StateTransformer {
-	return func(st *state.State) {
-		blockId := "identity"
-		st.Set(simple.New(&model.Block{
-			Id: blockId,
-			Content: &model.BlockContentOfRelation{
-				Relation: &model.BlockContentRelation{
-					Key: bundle.RelationKeyProfileOwnerIdentity.String(),
-				},
-			},
-			Restrictions: &model.BlockRestrictions{
-				Edit:   true,
-				Remove: true,
-				Drag:   true,
-				DropOn: true,
-			},
-		}))
-
-		st.InsertTo(state.TitleBlockID, model.Block_Bottom, blockId)
-	}
-}
-
 func migrationWithIdentityBlock(st *state.State) {
 	blockId := "identity"
 	st.Set(simple.New(&model.Block{
--- a/core/block/editor/profile.go
+++ b/core/block/editor/profile.go
@@ -126,7 +126,6 @@ func (p *Profile) StateMigrations() migration.Migrations {
 		{
 			Version: 4,
 			Proc: func(s *state.State) {
-				// skip it. it will be done separately via ProfileMigrationExtractUserStat in space.migrationProfileObject
 			},
 		},
 	})
--- a/core/block/editor/profilemigration/profilemigration.go
+++ b/core/block/editor/profilemigration/profilemigration.go
@@ -64,7 +64,8 @@ func ExtractCustomState(st *state.State) (userState *state.State, err error) {
 	newState.SetDetails(newStateDetails)
 	// remove the identity block
 	newState.Unlink(identityBlockId)
-
+	newState.CleanupBlock(identityBlockId)
+	
 	rootBlock := st.Get(st.RootId())
 	rootBlock.Model().ChildrenIds = slices.DeleteFunc(rootBlock.Model().ChildrenIds, func(s string) bool {
 		return !slices.Contains(whitelistBlocks, s)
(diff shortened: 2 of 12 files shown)


core/block/editor/profilemigration/profilemigration.go
```diff
@@ -0,0 +1,97 @@
...
+	})
+
+	uk, err := domain.NewUniqueKey(smartblock2.SmartBlockTypePage, InternalKeyOldProfileData)
+	if err != nil {
+		return nil, err
+	}
+	newState := state.NewDocWithUniqueKey(st.RootId(), blocksMap, uk).(*state.State)
+	newState.SetDetails(pbtypes.CopyStruct(st.CombinedDetails()))
+	newName := pbtypes.GetString(newState.Details(), bundle.RelationKeyName.String()) + " [migrated]"
+	newState.SetDetail(bundle.RelationKeyName.String(), pbtypes.String(newName))
```
deff7, line 62: SetDetailAndBundledRelation?

core/block/editor/profile.go
```diff
@@ -126,6 +145,12 @@ func (p *Profile) StateMigrations() migration.Migrations {
 			Version: 3,
 			Proc:    migrationSetHidden,
 		},
+		{
+			Version: 4,
+			Proc: func(s *state.State) {
+				// skip it. it will be done separately via ProfileMigrationExtractUserStat in space.migrationProfileObject
```
deff7, line 151: why do you need to increase the version?

core/block/editor/profilemigration/profilemigration.go
```diff
@@ -0,0 +1,97 @@
...
+	newState.SetDetail(bundle.RelationKeyIsHidden.String(), pbtypes.Bool(false))
+	newState.SetDetail(bundle.RelationKeyIsReadonly.String(), pbtypes.Bool(false))
+	newState.SetDetail(bundle.RelationKeyLayout.String(), pbtypes.Int64(int64(model.ObjectType_profile)))
+	newState.CleanupBlock(identityBlockId)
+
+	rootBlock := st.Pick(st.RootId())
```
deff7, line 68: st.Get

deff7, line 68: st.Get

core/block/editor/profilemigration/profilemigration.go
```diff
@@ -0,0 +1,97 @@
...
+	slices.DeleteFunc(rootBlock.Model().ChildrenIds, func(s string) bool {
```
deff7, line 69: `rootBlock.Model().ChildrenIds = slices.DeleteFunc(...`

core/block/editor/profilemigration/profilemigration.go
```diff
@@ -0,0 +1,97 @@
...
+func ExtractCustomState(st *state.State) (userState *state.State, err error) {
+	identityBlockId := "identity"
+	// we leave identity and other blocks in the original object to avoid them being re-added by old clients
+	whitelistBlocks := []string{state.HeaderLayoutID, state.DescriptionBlockID, state.FeaturedRelationsID, state.TitleBlockID, identityBlockId}
+	hasCustomState := false
+	st.Iterate(func(b simple.Block) (isContinue bool) {
+		if slices.Contains(whitelistBlocks, b.Model().Id) {
+			return true
+		}
+		if textBlock, ok := b.(text.Block); ok {
```
deff7, line 33: maybe use IsEmpty for text block too? You can have empty text and colored background, for example

core/block/editor/profilemigration/profilemigration.go
```diff
@@ -0,0 +1,97 @@
...
+		return nil, err
+	}
+	newState := state.NewDocWithUniqueKey(st.RootId(), blocksMap, uk).(*state.State)
+	newState.SetDetails(pbtypes.CopyStruct(st.CombinedDetails()))
+	newName := pbtypes.GetString(newState.Details(), bundle.RelationKeyName.String()) + " [migrated]"
+	newState.SetDetail(bundle.RelationKeyName.String(), pbtypes.String(newName))
+	newState.SetDetail(bundle.RelationKeyIsHidden.String(), pbtypes.Bool(false))
+	newState.SetDetail(bundle.RelationKeyIsReadonly.String(), pbtypes.Bool(false))
+	newState.SetDetail(bundle.RelationKeyLayout.String(), pbtypes.Int64(int64(model.ObjectType_profile)))
+	newState.CleanupBlock(identityBlockId)
```
deff7, line 66: why do you need this?

core/block/editor/profilemigration/profilemigration.go
```diff
@@ -0,0 +1,97 @@
...
+
+	rootBlock := st.Pick(st.RootId())
+	slices.DeleteFunc(rootBlock.Model().ChildrenIds, func(s string) bool {
+		return !slices.Contains(whitelistBlocks, s)
+	})
+
+	var whitelistDetailKeys = []string{
```
deff7, line 73: `:=`

core/block/editor/profilemigration/profilemigration.go
```diff
@@ -0,0 +1,97 @@
...
+		"iconEmoji",
+		"name",
+		"isHidden",
+		"featuredRelations",
+		"layout",
+		"layoutAlign",
+		"iconImage",
+		"iconOption",
+	}
+	keysToRemove := []string{}
```
deff7, line 83: `var ... []string` :)

core/block/editor/profilemigration/profilemigration.go
```diff
@@ -0,0 +1,97 @@
...
+		blocksMap[b.Model().Id] = b.Copy()
+		return true
+	})
+
+	uk, err := domain.NewUniqueKey(smartblock2.SmartBlockTypePage, InternalKeyOldProfileData)
+	if err != nil {
+		return nil, err
+	}
+	newState := state.NewDocWithUniqueKey(st.RootId(), blocksMap, uk).(*state.State)
+	newState.SetDetails(pbtypes.CopyStruct(st.CombinedDetails()))
```
deff7, line 60: you can filter details here, before adding them to state. Also you should add relation links for details to state (`st.AddRelationLinks(keys...)`)

space/clientspace/space.go
```diff
@@ -248,6 +255,51 @@ func (s *space) InstallBundledObjects(ctx context.Context) error {
 	return nil
 }
 
+func (s *space) migrationProfileObject(ctx context.Context) error {
```
deff7, line 258: maybe move this as function with `clientspace.Space` parameter to `profilemigration` package?

core/block/simple/dataview/dataview.go
```diff
@@ -516,3 +516,7 @@ func (s *Dataview) UpdateRelationOld(relationKey string, rel model.Relation) err
 
 	return nil
 }
+
+func (d *Dataview) IsEmpty() bool {
```
deff7, line 520: consider adding IsEmpty() to `simple.Block` interface if every block should implement it

also do not forget about simple table block, it's accidentally put in `core/block/editor/table` package, maybe because of some kind of import cycles

core/block/editor/profile.go
```diff
@@ -96,6 +93,28 @@ func migrationSetHidden(st *state.State) {
 	st.SetDetail(bundle.RelationKeyIsHidden.String(), pbtypes.Bool(true))
 }
 
+func migrationExtractBlocksIfExists() template.StateTransformer {
```
deff7, line 96: unused

core/block/editor/profilemigration/profilemigration.go
requilence: main point was to avoid migration if there is no any valuable information.
IsEmpty() method has a lot of checks inside, including align != default. If the text is empty it doesn't make any sense.

space/clientspace/space.go
requilence: I tried initially, but it's really hard because of cyclic deps. even with interfaces

core/block/editor/profile.go
requilence: removed

core/block/simple/dataview/dataview.go
requilence: we may think about it separately. Table is the layout block. It means it in a way always empty itself. We don't consider amount of children for the IsEmpty(). 

For this particular task we need to check if there is any block with it's own non-empty content. If it will be table without any blocks with data inside, we will ignore it

core/block/editor/profilemigration/profilemigration.go
requilence: done

core/block/editor/profilemigration/profilemigration.go
requilence: we need to reset isHidden because it will be set to true when initing the profile sb

core/block/editor/profilemigration/profilemigration.go
requilence: fixed

core/block/editor/profilemigration/profilemigration.go
requilence: thanks, fixed

core/block/editor/profilemigration/profilemigration.go
requilence: added relations

core/block/editor/profilemigration/profilemigration.go
deff7: I could have background color for example, is it valuable information? It could be used for decorations. And layout blocks could be valuable as well. Maybe just copy everything?

core/block/editor/profile.go
deff7: then maybe remove comment?

deff7 (approved):

PR #955, opened 2024-02-26
GO-2955 Fix cutting text range

KirillSto:
https://linear.app/anytype/issue/GO-2955/pressing-cmd-x-cuts-the-entire-block-instead-of-the-selected-text-when

We should do paste of blocks range instead of text range only when both rageTo and rangeFrom equal zero


core/block/editor/clipboard/clipboard.go
```diff
@@ -226,7 +226,7 @@ func isRangeSelect(firstTextBlock *model.Block, lastTextBlock *model.Block, rang
 		lastTextBlock == nil &&
 		rang != nil &&
 		rang.To-rang.From != int32(textutil.UTF16RuneCountString(firstTextBlock.GetText().Text)) &&
-		rang.To != 0 && rang.From != 0
+		(rang.To != 0 || rang.From != 0)
```
ra3orblade, line 229: isRangeSelected is true when `to > 0`, from can be whatever

PR #957, opened 2024-02-26
fix profile link migration

requilence:


deff7 (approved):

core/block/source/sub_object_links_migration.go
```diff
@@ -138,10 +138,9 @@ func (m *subObjectsAndProfileLinksMigration) migrateId(oldId string) (newId stri
 	if m.profileID != "" && m.identityObjectID != "" {
 		// we substitute all links to profile object with identity object EXCEPT the case with
 		// widget to identity in Personal space, we must substitute identity with profile to show links correctly
```
KirillSto, line 140: It seems this comment could be deleted, as we do not need logic with widget check anymore

PR #962, opened 2024-02-28
GO-2747: export all spaces

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-2747/export-all-spaces

In case we have `spaceId` = 0 for `ObjectListExport`, we try to get all objects from all spaces and store it in archive/folder in such way

```
spaces
|
| _  <spaceId>
     |
     |_ <objectId1>.json
```


core/block/export/export.go
```diff
@@ -534,8 +564,8 @@ func (e *export) createProfileFile(spaceID string, wr writer) error {
 	return nil
 }
 
-func (e *export) objectValid(sbType smartblock.SmartBlockType, id string, r *model.ObjectInfo, includeArchived bool, isProtobuf bool) bool {
-	if r.Id == addr.AnytypeProfileId {
+func (e *export) objectValid(sbType smartblock.SmartBlockType, id string, details *types.Struct, includeArchived bool, isProtobuf bool) bool {
```
KirillSto, line 567: We can accept model.ObjectInfo here except separatind object's details and id

KirillSto (approved):

PR #963, opened 2024-02-28
GO-2788: remove actual file path from sourceFilePath relation

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-2788/remove-actual-file-path-from-sourcefilepath-relation

We change sourcePathRelation logic for import. Now we don't store path to file with object, because it's sensitive information for shared spaces. So now it works in such way

1. Anyblock import : store spaceId/objectId of source object
2. Notion import: id of document in Notion
3. Markdown, txt, html and etc: fila path hash


core/block/import/pb/converter.go
```diff
@@ -434,22 +434,27 @@ func (p *Pb) cleanupEmptyBlock(snapshot *pb.SnapshotWithType) {
...
 	if createdDate == 0 {
-		mo.Snapshot.Data.Details.Fields[bundle.RelationKeyCreatedDate.String()] = pbtypes.Int64(time.Now().Unix())
+		sn.Snapshot.Data.Details.Fields[bundle.RelationKeyCreatedDate.String()] = pbtypes.Int64(time.Now().Unix())
 	}
 }
 
+func (p *Pb) setSourceFilePath(sn *pb.SnapshotWithType) {
+	spaceId := pbtypes.GetString(sn.Snapshot.Data.Details, bundle.RelationKeySpaceId.String())
+	id := pbtypes.GetString(sn.Snapshot.Data.Details, bundle.RelationKeyId.String())
+	sourceFilePath := filepath.Join(spaceId, "/", id)
```
deff7, line 454: "/" is redundant

core/block/import/common/common.go
```diff
@@ -49,9 +40,11 @@ func GetCommonDetails(sourcePath, name, emoji string, layout model.ObjectTypeLay
 	if emoji == "" {
 		emoji = slice.GetRandomString(randomIcons, name)
 	}
+	h := md5.Sum([]byte(sourcePath)) //nolint:gosec
```
deff7, line 43: maybe use sha256 here? Just to be safer

deff7 (approved):

PR #994, opened 2024-03-07
GO-3018: Replace space id inside paricipant ids during import

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-3018/replace-space-id-inside-paricipant-ids-during-import


core/block/import/common/objectcreator/objectcreator.go
```diff
@@ -160,6 +162,24 @@ func (oc *ObjectCreator) Create(dataObject *DataObject, sn *common.Snapshot) (*t
 	return respDetails, newID, nil
 }
 
+func (oc *ObjectCreator) createStaticObject(ctx context.Context, spaceID string, newID string, st *state.State) (*types.Struct, string, error) {
```
deff7, line 165: maybe don't create participants at all? They are created in aclobjectmanager. I think it is sufficient only to fix ids in links and details

AnastasiaShemyakinskaya: I suppose we can have a case, when we import archive in account without imported space (and acl). So participants are not creates in that case afaiu 🤔

AnastasiaShemyakinskaya: But participant are created in acl object manager, right? If we don't have imported space in our account, we don't have acl. So participant can't be created in this case 🤔
I guess we can export shared space and import it in any account

deff7: paritcipant is created in acl obj manager, right. So the only sensible thing that we can do is to replace SpaceId. But there is a funny corner case where link targets identity that doesn't exist in current space and it's make no sense to use it. Maybe we should substitute identity too, but it seems weird as well. Maybe you should discuss it with Yura or Kostya

AnastasiaShemyakinskaya: I guess for backup purpose we can just change id without creation. I will fix it

AnastasiaShemyakinskaya: Fixed

deff7 (approved):

PR #999, opened 2024-03-12
GO-3041 CopyStruct add copyVals arg

requilence:


util/pbtypes/copy.go
```diff
@@ -30,7 +30,7 @@ func CopyBlock(in *model.Block) (out *model.Block) {
 	return
 }
 
-func CopyStruct(s *types.Struct) *types.Struct {
+func CopyStruct(s *types.Struct, copyVals bool) *types.Struct {
```
fat-fellow, line 33: would be nice to have a test for the flag

fat-fellow (approved):

fat-fellow (changes requested):
Tests

util/pbtypes/copy.go
requilence: @fat-fellow added some tests

fat-fellow (approved):

requilence:
rebased to fix unsigned commit that somehow appeared in the history

PR #1013, opened 2024-03-18
GO-2930: Store device names

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-2930/store-device-names

1. Add new device object  to store devices names in tech space
2. Add RPC `DeviceSetName` to change device name
3. Add RPC `DeviceList` to get devices list


pkg/lib/pb/model/protos/models.proto
```diff
@@ -50,6 +50,7 @@ enum SmartBlockType {
     FileObject = 0x215;
 
     NotificationObject = 0x217;
+    DeviceObject = 0x218;
```
deff7, line 53: I think Page type is sufficient for this task, as you probably don't need special behavior for smartblock

fat-fellow (dismissed):
Tests

pkg/lib/pb/model/protos/models.proto
AnastasiaShemyakinskaya: Fixed

core/block/editor/smartblock/smartblock.go
```diff
@@ -1328,19 +1328,21 @@ func ObjectApplyTemplate(sb SmartBlock, s *state.State, templates ...template.St
 
 func hasChanges(changes []*pb.ChangeContent) bool {
 	for _, ch := range changes {
-		if isStoreOrNotificationChanges(ch) {
+		if isSuitableChanges(ch) {
 			return true
 		}
```
requilence, line 1333: It's not clear from the name of the func what hasChanges actually does. It's in the context of the sb package, not notifications

core/block/editor/state/change.go
```diff
@@ -823,6 +845,34 @@ func (s *State) makeNotificationChanges() []*pb.ChangeContent {
 	return changes
 }
 
+func (s *State) makeDeviceInfoChanges() []*pb.ChangeContent {
+	if s.parent == nil && len(s.parent.ListDevices()) == 0 {
```
requilence, line 849: 1. Nil deref here(in case s.parent is nil)
2. Also for the first change should it work fine if the parent is nil but we have devices in the current state?

requilence (changes requested):

core/block/editor/smartblock/smartblock.go
AnastasiaShemyakinskaya: We can rename `hasChanges` to something like `hasChangesToPush`, because we can check such changes in that function: store, notifications, devices changes

core/block/editor/state/change.go
AnastasiaShemyakinskaya: Fixed and add test

requilence:
lets disable indexation for device object

requilence (approved):

PR #1021, opened 2024-03-19
GO-3096: add space name to notifications

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-3096/add-space-name-to-all-notifications-which-has-spaceid


deff7 (approved):

space/internal/components/aclnotifications/notifications.go
```diff
@@ -99,13 +108,34 @@ func (n *aclNotificationSender) AddRecords(acl list.AclList,
...
+			spaceName:     spaceName,
 		})
 		if err != nil {
 			logger.Errorf("failed to add acl record, %s", err)
 		}
 	}
 }
 
+func (n *aclNotificationSender) getSpaceName(spaceId string) string {
+	records, _, err := n.objectStore.Query(database.Query{
```
mcrakhman, line 120: This should work, though I would also make sure that the layout is "spaceView" (add another check), so we would not get some other object which in theory can use the same relation

mcrakhman (approved):

PR #1035, opened 2024-03-25
GO-3112: "Untitled" in the name of the space in the notification

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-3112/untitled-in-the-name-of-the-space-in-the-notification

Extract space name from object store and add it to
1. Import notification
2. Gallery notification
3. Notification for space deletion


fat-fellow (changes requested):
tests

mcrakhman (approved):

PR #1036, opened 2024-03-26
GO-3127: Leave request approve method

mcrakhman:
Add SpaceLeaveApprove request:
- check that we do have a remove request from such participants
- if yes delete them
- if no send an error


core/acl/aclservice.go
```diff
@@ -195,6 +197,37 @@ func (a *aclService) ChangePermissions(ctx context.Context, spaceId string, perm
...
+			if rec.RequestIdentity.Equals(identity) {
+				delete(identitiesMap, identity.Account())
+			}
+		}
+	}
+	if len(identitiesMap) != 0 {
+		acl.RUnlock()
+		var identitiesString string
+		for identity := range identitiesMap {
+			identitiesString += identity + ", "
```
fat-fellow, line 223: Stringbuffer?

mcrakhman: This is just an error to print the identities :-) It doesn't matter

mcrakhman: But it is better to just do strings with separator though

mcrakhman: Changed that

PR #1038, opened 2024-03-26
GO-2969 reorder relation options

requilence:
This PR adds an ability for clients to reorder the relation's options.
It's done in a way that default order is the date of creation (based on the BSON ID sorting)

RPC added: `RelationMoveOption`.
- in case you are using sorting `relationOptionOrder ASC`(for select), set the Sort to 0(OlderFirst).
- in case you are using sorting `relationOptionOrder DESC`(for multi-select tags), set the Sort to 1(NewerFirst).

Relations added: `relationOptionOrder`, `relationOptionInternalOrder`

Client should use `relationOptionOrder` for sorting because it contains the fallback to the creation date

Under the hood it is using the lexicographical sorting. It allows to change the position of a single element without having a centralised list of options. There is no need for any migration, because we already use BSON ID for option ids, and bson ids are sorted in the chronological way.
When you add the new option it's automatically will be added to the begin or the end (depends on the sort) because of bsons


fat-fellow (changes requested):
Tests

core/block/service.go
```diff
@@ -685,6 +686,51 @@ func (s *Service) RemoveListOption(optionIds []string, checkInObjects bool) erro
 	return nil
 }
```
fat-fellow, line 687: Need to extract this code from the block service

PR #1062, opened 2024-04-01
GO-2734 Add globalName directly

KirillSto:
https://linear.app/anytype/issue/GO-2734/add-cache-for-client-to-resolve-id-name

Save globalName of myIdentity directly, as NS need some time to process newly registered name

### DIFF
--- a/core/payments/payments.go
+++ b/core/payments/payments.go
@@ -42,7 +42,7 @@ var (
 )
 
 type globalNamesUpdater interface {
-	UpdateGlobalNames()
+	UpdateGlobalNames(myIdentityGlobalName string)
 }
 
 /*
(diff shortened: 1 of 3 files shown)


AnthonyAkentiev (approved):

core/payments/payments.go
```diff
@@ -273,7 +273,7 @@ func (s *service) GetSubscriptionStatus(ctx context.Context, req *pb.RpcMembersh
 
 	// 5 - if requested any name has changed, then we need to update details of local identity
 	if isDiffRequestedName {
-		s.profileUpdater.UpdateGlobalNames()
+		s.profileUpdater.UpdateGlobalNames(status.RequestedAnyName)
```
AnthonyAkentiev, line 276: Can be empty for some tiers that feature NO any name, but I think it will work here...

AnthonyAkentiev (approved):

PR #1065, opened 2024-04-02
GO-3027: Error Importing Object to Another Space

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-3027/error-importing-object-to-another-space

Deleted linked objects were also exported. This leads to export error during single object export. 
So I skip exporting of deleted objects

### DIFF
--- a/core/block/export/export.go
+++ b/core/block/export/export.go
@@ -47,7 +47,6 @@ import (
 	"github.com/anyproto/anytype-heart/space"
 	"github.com/anyproto/anytype-heart/space/clientspace"
 	"github.com/anyproto/anytype-heart/space/spacecore/typeprovider"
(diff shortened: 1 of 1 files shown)


core/block/export/export.go
```diff
@@ -372,6 +373,10 @@ func (e *export) getNested(spaceID string, id string, docs map[string]*types.Str
 	return nestedDocsIds
 }
 
+func (e *export) isLinkedObjectExist(rec []database.Record) bool {
```
KirillSto, line 376: We can make isLinkedObjectExist package-level func, as no fields or methods of export are used

KirillSto (approved):

fat-fellow (changes requested):
Tests

fat-fellow (approved):

PR #1079, opened 2024-04-04
GO-3173: send to client number of objects which was imported within one import 

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-3173/send-to-client-number-of-objects-which-was-imported-within-one-import

1. Add new field in `ObjectImport` request - `objectCount`. It represents number of created objects during one import
2. Small refactoring - extract return value from import service to separate data structure `ImportResponse`

### DIFF
--- a/core/block/import/importer.go
+++ b/core/block/import/importer.go
@@ -167,7 +167,12 @@ func (i *Import) sendFileEvents(returnedErr error) {
 	i.fileSync.ClearImportEvents()
 }
 
-func (i *Import) importFromBuiltinConverter(ctx context.Context, req *pb.RpcObjectImportRequest, c common.Converter, progress process.Progress, origin objectorigin.ObjectOrigin) (string, int64, error) {
+func (i *Import) importFromBuiltinConverter(ctx context.Context,
+	req *pb.RpcObjectImportRequest,
+	c common.Converter,
+	progress process.Progress,
+	origin objectorigin.ObjectOrigin,
+) (string, int64, error) {
 	allErrors := common.NewError(req.Mode)
 	res, err := c.GetSnapshots(ctx, req, progress)
 	if !err.IsEmpty() {
--- a/core/block/import/importer_test.go
+++ b/core/block/import/importer_test.go
@@ -249,7 +249,6 @@ func Test_ImportIgnoreErrorModeWithTwoErrorsPerFile(t *testing.T) {
 
 	assert.NotNil(t, res.Err)
 	assert.Contains(t, res.Err.Error(), "converter error")
-	assert.Contains(t, res.Err.Error(), "converter error", "creator error")
 }
 
 func Test_ImportExternalPlugin(t *testing.T) {
--- a/core/object.go
+++ b/core/object.go
@@ -785,10 +785,16 @@ func (mw *Middleware) ObjectSetInternalFlags(cctx context.Context, req *pb.RpcOb
 }
 
 func (mw *Middleware) ObjectImport(cctx context.Context, req *pb.RpcObjectImportRequest) *pb.RpcObjectImportResponse {
-	response := func(code pb.RpcObjectImportResponseErrorCode, rootCollectionID string, objectCount int64, err error) *pb.RpcObjectImportResponse {
-		m := &pb.RpcObjectImportResponse{Error: &pb.RpcObjectImportResponseError{Code: code}, CollectionId: rootCollectionID, ObjectsCount: objectCount}
-		if err != nil {
-			m.Error.Description = err.Error()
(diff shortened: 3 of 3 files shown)


core/block/import/importer.go
```diff
@@ -136,17 +141,23 @@ func (i *Import) Import(ctx context.Context,
...
-	var rootCollectionId string
+	var (
+		rootCollectionId string
+		objectsCount     int64
+	)
+	returnedErr = fmt.Errorf("unknown import type %s", req.Type)
 	if c, ok := i.converters[req.Type.String()]; ok {
-		rootCollectionId, returnedErr = i.importFromBuiltinConverter(ctx, req, c, progress, origin)
-		return rootCollectionId, progress.Id(), returnedErr
+		rootCollectionId, objectsCount, returnedErr = i.importFromBuiltinConverter(ctx, req, c, progress, origin)
```
KirillSto, line 150: Maybe we can return `ImportResponse` from `importFromBuiltinConverter` and than enrich it with progress.ID ?

core/block/import/importer.go
```diff
@@ -136,17 +141,23 @@ func (i *Import) Import(ctx context.Context,
...
 	}
 	if req.Type == model.Import_External {
-		returnedErr = i.importFromExternalSource(ctx, req, progress)
-		return rootCollectionId, "", returnedErr
+		objectsCount, returnedErr = i.importFromExternalSource(ctx, req, progress)
```
KirillSto, line 153: Same here

core/block/import/importer.go
```diff
@@ -156,41 +167,36 @@ func (i *Import) sendFileEvents(returnedErr error) {
 	i.fileSync.ClearImportEvents()
 }
 
-func (i *Import) importFromBuiltinConverter(ctx context.Context,
-	req *pb.RpcObjectImportRequest,
-	c common.Converter,
-	progress process.Progress,
-	origin objectorigin.ObjectOrigin,
-) (string, error) {
+func (i *Import) importFromBuiltinConverter(ctx context.Context, req *pb.RpcObjectImportRequest, c common.Converter, progress process.Progress, origin objectorigin.ObjectOrigin) (string, int64, error) {
```
KirillSto, line 170: Method's parameters blow up in single line again :sweat_smile:

core/block/import/importer_test.go
```diff
@@ -235,17 +239,17 @@ func Test_ImportIgnoreErrorModeWithTwoErrorsPerFile(t *testing.T) {
...
 		Mode:                  1,
 		SpaceId:               "space1",
 	}, objectorigin.Import(model.Import_Notion), nil)
 
-	assert.NotNil(t, res)
-	assert.Contains(t, res.Error(), "converter error")
-	assert.Contains(t, res.Error(), "converter error", "creator error")
+	assert.NotNil(t, res.Err)
+	assert.Contains(t, res.Err.Error(), "converter error")
+	assert.Contains(t, res.Err.Error(), "converter error", "creator error")
```
KirillSto, line 252: It seems that "converter error" should be removed

core/block/import/types.go
```diff
@@ -14,14 +14,21 @@ import (
 	"github.com/anyproto/anytype-heart/pb"
 )
 
+type ImportResponse struct {
```
KirillSto, line 17: That is cool!

core/object.go
```diff
@@ -785,25 +785,25 @@ func (mw *Middleware) ObjectSetInternalFlags(cctx context.Context, req *pb.RpcOb
 }
 
 func (mw *Middleware) ObjectImport(cctx context.Context, req *pb.RpcObjectImportRequest) *pb.RpcObjectImportResponse {
-	response := func(code pb.RpcObjectImportResponseErrorCode, rootCollectionID string, err error) *pb.RpcObjectImportResponse {
-		m := &pb.RpcObjectImportResponse{Error: &pb.RpcObjectImportResponseError{Code: code}, CollectionId: rootCollectionID}
+	response := func(code pb.RpcObjectImportResponseErrorCode, rootCollectionID string, objectCount int64, err error) *pb.RpcObjectImportResponse {
```
KirillSto, line 788: `response` could also be `func(pb.RpcCode, ImportResponse) *pb.RpcResponse`

KirillSto (changes requested):

core/block/import/importer.go
AnastasiaShemyakinskaya: I don't think this is a good idea, because it's not obvious mutate of ImportResponse structure

core/block/import/importer.go
AnastasiaShemyakinskaya: Response above

KirillSto (approved):

core/block/import/importer.go
KirillSto: Yeah, agree

PR #1080, opened 2024-04-04
GO-2673 Exclude deleted objects

KirillSto:
https://linear.app/anytype/issue/GO-2673/implement-validator-on-usecase-addition-in-gallery

- exclude deleted object from archive
- exclude templates with no target type presented in the archive
- delete recommendedRelations of type objects in case these relations are not in the archive

### DIFF
--- a/cmd/usecasevalidator/main.go
+++ b/cmd/usecasevalidator/main.go
@@ -75,6 +75,7 @@ const anytypeProfileFilename = addr.AnytypeProfileId + ".pb"
 var (
 	errIncorrectFileFound = fmt.Errorf("incorrect protobuf file was found")
 	errValidationFailed   = fmt.Errorf("validation failed")
+	errSkipObject         = fmt.Errorf("validation failed, but object can be skipped")
 )
 
 func main() {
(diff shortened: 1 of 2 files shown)


cmd/usecasevalidator/main.go
```diff
@@ -75,6 +75,7 @@ const anytypeProfileFilename = addr.AnytypeProfileId + ".pb"
 var (
 	errIncorrectFileFound = fmt.Errorf("incorrect protobuf file was found")
 	errValidationFailed   = fmt.Errorf("validation failed")
+	errSkipObject         = fmt.Errorf("validation failed, but object can be skipped")
```
AnastasiaShemyakinskaya, line 78: Choose another error message please. Something like "object is invalid, skip it".
It's just misleading a bit imho. Why don't we failed validation if we have error about failed validation 😅

KirillSto: Yeah, thank you, changed the error message

fat-fellow (approved):

KirillSto:
Tests are also needed for validators. I will design them ubder https://linear.app/anytype/issue/GO-3207/implement-user-readable-validator-output

PR #1084, opened 2024-04-04
GO-3194 payments/nameservice protocol refactoring

AnthonyAkentiev:
---

---

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [x] 👍 yes

### Added to documentation?

- [ ] 📜 README.md


pkg/lib/pb/model/protos/models.proto
```diff
@@ -1218,30 +1208,38 @@ message MembershipTierData {
...
     string colorStr = 11;
+
+    // Stripe platform-specific data:
+    string stripeProductId = 12;
+    string stripeManageUrl = 13;
+    uint32 stripePriceUsdCents = 14;
+
+    // iOS platform-specific data:
+    string iosProductId = 15;
+    string iosManageUrl = 16;
```
AnthonyAkentiev, line 1240: Or this field is not PER-tier?

pkg/lib/pb/model/protos/models.proto
```diff
@@ -1218,30 +1208,38 @@ message MembershipTierData {
...
+
+    // Android platform-specific data:
+    string androidProductId = 17;
+    string androidManageUrl = 18;
```
AnthonyAkentiev, line 1244: Or this field is not PER-tier?

pkg/lib/pb/model/protos/models.proto
```diff
@@ -1157,21 +1157,7 @@ message ManifestInfo {
...
-        // it will still create an active subscription, but with NO features
-        TierBuilder1WeekTEST = 2;
-        // this tier can be used just for testing in debug mode
-        // it will still create an active subscription, but with NO features
-        TierCoCreator1WeekTEST = 3;
-        TierBuilder = 4;
-        TierCoCreator = 5;
-    }
-    
+message Membership {
```
KirillSto, line 1160: Where are all these tiers identifiers are going? To any-sync protocol?

pkg/lib/pb/model/protos/models.proto
```diff
@@ -1157,21 +1157,7 @@ message ManifestInfo {
...
-        // it will still create an active subscription, but with NO features
-        TierCoCreator1WeekTEST = 3;
-        TierBuilder = 4;
-        TierCoCreator = 5;
-    }
-    
+message Membership {    
     enum Status {
         StatusUnknown = 0;
         // please wait a bit more
```
konstantiniiv, line 1163: I don’t understand what it means "please wait a bit more"? wait on the client side for changing the status from StatusUnknown to another? there will be other statuses, I suppose?

pkg/lib/pb/model/protos/models.proto
```diff
@@ -1186,12 +1172,16 @@ message Membership {
     }
     
     enum PaymentMethod {
-        MethodCard = 0; // Stripe
```
konstantiniiv, line 1189: maybe it's worth adding 'None' as the default value, for a situation when you can buy this Tier?

pkg/lib/pb/model/protos/models.proto
```diff
@@ -1186,12 +1172,16 @@ message Membership {
...
-        MethodGooglePay = 3;
-        MethodAppleInapp = 4;
-        MethodGoogleInapp = 5;
+        MethodInappApple = 2;
+        MethodInappGoogle = 3;
+    }
+
+    enum EmailVerificationStatus {
+        NotVerified = 0;
+        CodeSent = 1;
```
konstantiniiv, line 1183: after sending the erroneous code, there will be a status "NotVerified"?

pkg/lib/pb/model/protos/models.proto
ignatovv: it is better to keep it per tier

pkg/lib/pb/model/protos/models.proto
AnthonyAkentiev: Yes, If you want/need -> I can keep it here also and roll this change back

pkg/lib/pb/model/protos/models.proto
AnthonyAkentiev: How subscription purchase works (in Stripe case):

* User requests a new subscription.
* User is redirected to Stripe.
* User makes the payment.
* Stripe immediately returns "success" and redirects you via a deeplink <-- at this point, the user may still not have the subscription.
* (Uncertain when) Within a certain period of time, Stripe sends us a webhook <-- at this point, the subscription will be created and will enter the **pending** state.
* (Uncertain when) We process it and load various settings: names, restrictions, etc.
Once everything is completed, the subscription transitions to the active state.

added more comments to the .proto file

pkg/lib/pb/model/protos/models.proto
AnthonyAkentiev: No. I have added more comments, please review.

pkg/lib/pb/model/protos/models.proto
KirillSto: No, it is better to keep it at any-sync, you are right

fat-fellow (approved):

PR #1104, opened 2024-04-10
GO-3251 indexer: finish the ft queue gracefully on close

requilence:


core/indexer/fulltext.go
```diff
@@ -31,9 +32,12 @@ func (i *indexer) ForceFTIndex() {
 	}
 }
 
-func (i *indexer) ftLoop() {
+// ftLoop runs full-text indexer
+// MUST NOT be called more than once
```
deff7, line 36: it's better to use context with cancel, it could be closed multiple times

requilence: this method existing only to extract the code for the goroutine and can must called only once. Also there is a comment

pkg/lib/localstore/ftsearch/ftsearch.go
```diff
@@ -133,6 +133,9 @@ func (f *ftSearch) BatchIndex(docs []SearchDoc) (err error) {
 		}
 	}()
 	for _, doc := range docs {
+		if ctx.Err() == context.Canceled {
```
AnastasiaShemyakinskaya, line 136: I suppose, should be `error.Is`

requilence: for the context error it makes no sense because it has the defined finite list of constant errors: context.Canceled or context.DeadlineExceeded

fat-fellow (approved):

AnastasiaShemyakinskaya (approved):

PR #1123, opened 2024-04-16
GO-3286: fix not working remove and delete notifications from member side

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-3286/fix-not-working-remove-and-delete-notifications-from-member-side

## Summary by CodeRabbit

- **New Features**
	- Enhanced security and stability in notification handling through improved synchronization.
	- Updated account status management to reflect current data accuracy.

- **Bug Fixes**
	- Corrected account status parameters in testing scenarios to align with actual application behavior.

- **Tests**
	- Added comprehensive tests for new notification functionalities and account status updates.

- **Documentation**
	- Introduced interfaces for better modularity and testing in notification and identity services.


space/internal/components/aclnotifications/notifications.go
```diff
@@ -71,9 +75,13 @@ func (n *aclNotificationSender) Run(ctx context.Context) (err error) {
 }
 
 func (n *aclNotificationSender) Close(ctx context.Context) (err error) {
-	if err := n.batcher.Close(); err != nil {
-		logger.Errorf("failed to close batcher, %s", err)
-	}
+	go func() {
+		n.mx.Lock()
```
fat-fellow, line 79: batcher is thread safe

space/internal/components/aclnotifications/notifications.go
```diff
@@ -71,9 +75,13 @@ func (n *aclNotificationSender) Run(ctx context.Context) (err error) {
 }
 
 func (n *aclNotificationSender) Close(ctx context.Context) (err error) {
-	if err := n.batcher.Close(); err != nil {
```
mcrakhman, line 74: Why was this needed? It looks weird, we should wait when batcher closes synchronously

space/internal/components/aclobjectmanager/aclobjectmanager.go
```diff
@@ -157,14 +157,19 @@ func (a *aclObjectManager) process() {
 
 func (a *aclObjectManager) processAcl() (err error) {
 	var (
-		common   = a.sp.CommonSpace()
-		acl      = common.Acl()
-		aclState = acl.AclState()
+		common         = a.sp.CommonSpace()
```
mcrakhman, line 160: Tbh looking at this now, I would've done it like this:
- remove checking for self status from process states, to make it simpler
- create another method `isAccountRemoved` which would be very simple: call `aclState.Permissions(aclState.Identity())`check if the account was removed (you can also add there the logic with upToDate)

space/internal/components/aclnotifications/notifications.go
AnastasiaShemyakinskaya: It doesn't relate to badger, notification can be send in two different processes (when space `loader` is inited and in `UpdateAcl`) in acl object manager. Space loader can close components like acl notification service, while another process (`UpdateAcl`) is still executed. So we avoid racing here

space/internal/components/aclnotifications/notifications.go
```diff
@@ -71,6 +75,8 @@ func (n *aclNotificationSender) Run(ctx context.Context) (err error) {
 }
 
 func (n *aclNotificationSender) Close(ctx context.Context) (err error) {
+	n.mx.Lock()
```
mcrakhman, line 78: I guess we don't need a mutex here

fat-fellow: I've said that before also

mcrakhman: We already revised approach a little bit, so some changes will follow

mcrakhman (approved):

PR #1128, opened 2024-04-17
Go 2331 fix readonly for tag status rels

KirillSto:
https://linear.app/anytype/issue/GO-2331/all-relations-with-tag-or-status-which-i-created-are-locked-and-unable

Set relationReadonlyValue=false for all relations with tag or status format


fat-fellow (changes requested):
tests

PR #1142, opened 2024-04-18
GO-3258 temp disable blocks parent cache

requilence:
In some cases the cache state is invalid and leads to error when apply changes "target without parent"

## Summary by CodeRabbit

- **Refactor**
	- Temporarily disabled the parent IDs cache in the editor to enhance testing. Cache will be re-enabled after test coverage is verified.

### DIFF
--- a/core/block/editor/state/state.go
+++ b/core/block/editor/state/state.go
@@ -346,6 +346,9 @@ func (s *State) ResetParentIdsCache() {
 }
 
 func (s *State) EnableParentIdsCache() bool {
+	// temporary disable the cache
+	// todo: enable after we cover everything with tests
+	return false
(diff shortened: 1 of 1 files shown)


core/block/editor/state/state.go
```diff
@@ -346,6 +346,9 @@ func (s *State) ResetParentIdsCache() {
 }
 
 func (s *State) EnableParentIdsCache() bool {
+	// temporary disable the cache
+	// todo: enable after we cover everything with tests
+	return false
```
deff7, line 351: maybe `return true`?

requilence: yep, changed

fat-fellow (approved):

PR #1220, opened 2024-05-07
GO-2696 Check networkId on SpaceJoin

KirillSto:
https://linear.app/anytype/issue/GO-2696/space-joining-validate-network-id

On SpaceJoin we should check if networkId is the same


mcrakhman (changes requested):
Please move everything to acl service level and write tests, config should also be inside acl service. config should be mocked when testing

also the error should also be in acl service I guess

PR #1225, opened 2024-05-09
GO-2331 Space Migrations

KirillSto:
https://linear.app/anytype/issue/GO-2331/all-relations-with-tag-or-status-which-i-created-are-locked-and-unable

- Set relationReadonlyValue=false for all relations with tag or status format
- Introduce migration.Runner component to handle readonlyRelationsFixer and systemObjectReviser migrations
- QueryWithCtx method is added to ObjectStore


core/migration/relationsfixer.go
```diff
@@ -0,0 +1,88 @@
...
+
+type readonlyRelationsFixer struct{}
+
+func (readonlyRelationsFixer) Name() string {
+	return "ReadonlyRelationsFixer"
+}
+
+func (readonlyRelationsFixer) Run(ctx context.Context, store QueryableStore, space DoableViaContext) (toMigrate, migrated int, err error) {
+	spaceId := space.Id()
+	var relations []database.Record
```
AnastasiaShemyakinskaya, line 25: Unnecessary var here?

core/migration/relationsfixer.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	for _, r := range relations {
+		var (
+			name = pbtypes.GetString(r.Details, bundle.RelationKeyName.String())
+			uk   = pbtypes.GetString(r.Details, bundle.RelationKeyUniqueKey.String())
+		)
+
+		format, ok := model.RelationFormat_name[int32(pbtypes.GetInt64(r.Details, bundle.RelationKeyRelationFormat.String()))]
+		if !ok {
```
AnastasiaShemyakinskaya, line 45: I don't think it is possible scenario according to your code

core/migration/runner.go
```diff
@@ -0,0 +1,71 @@
...
+	); err != nil {
+		log.Errorf("failed to run default migrations: %v", err)
+	}
+}
+
+func run(ctx context.Context, store objectstore.ObjectStore, space clientspace.Space, migrations ...Migration) (mErr error) {
+	var (
+		spaceId = space.Id()
+		finish  = make(chan struct{})
+		sstore  = &safeStore{store: store}
```
AnastasiaShemyakinskaya, line 40: Maybe rename sstore to something more meaningful?

core/migration/runner.go
```diff
@@ -0,0 +1,71 @@
...
+			log.Error(fErr)
+			if errors.Is(err, context.Canceled) {
+				return
+			}
+			continue
+		}
+		log.Debugf("migration '%s' in space '%s' is successful. %d out of %d objects were migrated",
+			m.Name(), spaceId, migrated, toMigrate)
+	}
+	finish <- struct{}{}
```
AnastasiaShemyakinskaya, line 69: Should be in defer statement?

core/migration/safestore.go
```diff
@@ -0,0 +1,36 @@
...
+	m      sync.RWMutex
+}
+
+func (s *safeStore) Query(q database.Query) (records []database.Record, err error) {
+	s.m.RLock()
+	if s.locked {
+		s.m.RUnlock()
+		return nil, context.Canceled
+	}
+	s.m.RUnlock()
```
AnastasiaShemyakinskaya, line 28: Can be in defer statement

core/block/object/objectcreator/migration.go
```diff
@@ -0,0 +1,25 @@
+package objectcreator
+
+import (
+	"context"
+
+	"github.com/anyproto/anytype-heart/core/migration"
+	"github.com/anyproto/anytype-heart/space/clientspace"
+)
+
+func (s *service) RunMigrations(ctx context.Context, space clientspace.Space) {
```
requilence, line 10: I think you can have a single context for all RunMigration calls here. Because you still have a single Close. RunMigrations now is not thread-safe (because of appending to the slice). Just create the ctx and cancel in service Init.

KirillSto: Thank you! I forgot that appending to slice is not thread-safe, but your solution requires storing context inside ObjectCreator structure, that is a bad practice. Moreover, we pass contexts from space builder, so we have context for each space.

Maybe it is better to modify CancelFunc slice under lock?

core/migration/runner.go
KirillSto: If we put this statement in defer, we will send msg to finish channel even in case context was cancelled, so listenning goroutine would not be able to read this message. In this case program will stuck

core/migration/safestore.go
KirillSto: I did not put Unlock to defer statement, because I don't want to wait until Query will be done. Lock is used here only to access bool value, but Query could be called by multiple threads simulteneously

core/migration/relationsfixer.go
KirillSto: Thank you! Removed this check

core/migration/relationsfixer.go
KirillSto: Done

core/migration/runner.go
KirillSto: Done!

core/block/object/objectcreator/migration.go
KirillSto: However, use of single ctx is pretty accurate. I did it acc to your suggestion

AnastasiaShemyakinskaya (approved):

core/migration/runner.go
```diff
@@ -0,0 +1,77 @@
...
```
deff7, line 75: what's a purpose of this channel?

core/migration/runner.go
```diff
@@ -0,0 +1,77 @@
...
+	DoCtx(ctx context.Context, objectId string, apply func(sb smartblock.SmartBlock) error) error
+	Id() string
+}
+
+type Migration interface {
+	Run(context.Context, queryableStore, doableViaContext) (toMigrate, migrated int, err error)
+	Name() string
+}
+
+func Run(ctx context.Context, store objectstore.ObjectStore, space clientspace.Space) {
```
deff7, line 33: maybe create component for it?

core/block/object/objectcreator/creator.go
```diff
@@ -47,7 +47,9 @@ type Service interface {
 	CreateSmartBlockFromStateInSpace(ctx context.Context, space clientspace.Space, objectTypeKeys []domain.TypeKey, createState *state.State) (id string, newDetails *types.Struct, err error)
 
 	InstallBundledObjects(ctx context.Context, space clientspace.Space, sourceObjectIds []string, isNewSpace bool) (ids []string, objects []*types.Struct, err error)
-	app.Component
+
+	RunMigrations(space clientspace.Space)
```
deff7, line 51: why it's here?

core/migration/runner.go
KirillSto: By sending message to this channel we notify [goroutine](https://github.com/anyproto/anytype-heart/pull/1225/files#diff-86953f03b4d14f9d8d39d649c5ce980b2e4bccbf6af96d562566bf0b0f7684baR49-R59) waiting for context cancellation that migration is over, so goroutine could be finished. Otherwise this waiting goroutine will remain waiting for context cancellation. As we define common context for all space migrations that is stored in objectCreator struct, these waiting goroutines will end their work only on objectCreator component closing

core/migration/runner.go
KirillSto: Done!

core/block/object/objectcreator/creator.go
KirillSto: Migration logic is removed from migrator

space/internal/components/migration/runner.go
```diff
@@ -0,0 +1,116 @@
...
+func (r *Runner) run(migrations ...Migration) (mErr error) {
+	spaceId := r.spc.Id()
+
+	for _, m := range migrations {
+		toMigrate, migrated, err := m.Run(r.ctx, r.store, r.spc)
+		if err != nil {
+			if errors.Is(err, context.Canceled) {
+				return err
+			}
+			mErr = multierror.Append(mErr, fmt.Errorf(errFormat, m.Name(), spaceId, err, migrated, toMigrate))
```
mcrakhman, line 109: I understand that you wanted to use same string errFormat, but it would be better to create a separate function for that

mcrakhman: check aclService maybe to also return typed error

space/internal/components/migration/runner.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/anyproto/anytype-heart/pkg/lib/localstore/objectstore"
+	"github.com/anyproto/anytype-heart/space/clientspace"
+	"github.com/anyproto/anytype-heart/space/internal/components/migration/common"
+	"github.com/anyproto/anytype-heart/space/internal/components/migration/readonlyfixer"
+	"github.com/anyproto/anytype-heart/space/internal/components/migration/systemobjectreviser"
+	"github.com/anyproto/anytype-heart/space/internal/components/spaceloader"
+)
+
+const (
+	CName     = "common.components.migration-runner"
```
mcrakhman, line 22: lets name the component with the path (e.g. without - runner)

space/internal/components/migration/readonlyfixer/relationsfixer.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/anyproto/anytype-heart/core/block/editor/smartblock"
+	"github.com/anyproto/anytype-heart/pkg/lib/bundle"
+	"github.com/anyproto/anytype-heart/pkg/lib/database"
+	"github.com/anyproto/anytype-heart/pkg/lib/logging"
+	"github.com/anyproto/anytype-heart/pkg/lib/pb/model"
+	"github.com/anyproto/anytype-heart/space/internal/components/migration/common"
+	"github.com/anyproto/anytype-heart/util/pbtypes"
+)
+
+const name = "ReadonlyRelationsFixer"
```
mcrakhman, line 19: better to have names like CName, or the other approach - you can actually pass log to migration from parent component

space/internal/components/migration/common/common.go
```diff
@@ -0,0 +1,17 @@
+package common
+
+import (
+	"context"
+
+	"github.com/anyproto/anytype-heart/core/block/editor/smartblock"
+	"github.com/anyproto/anytype-heart/pkg/lib/database"
+)
+
+type SpaceWithCtx interface {
```
mcrakhman, line 10: Maybe put all common interfaces in dependencies thus we will at least have same approach inside the space module

pkg/lib/localstore/objectstore/queries.go
```diff
@@ -23,6 +24,32 @@ func (s *dsObjectStore) Query(q database.Query) ([]database.Record, error) {
...
+		recs, err := s.Query(q)
+		if err != nil {
+			errCh <- err
+		} else {
+			recordsCh <- recs
+		}
+	}()
+
+	select {
+	case <-ctx.Done():
```
mcrakhman, line 44: I would actually do it synchronously, I dont like hanging goroutine especially if you return when context is done.

pkg/lib/localstore/objectstore/queries.go
```diff
@@ -23,6 +24,32 @@ func (s *dsObjectStore) Query(q database.Query) ([]database.Record, error) {
...
+	if err = ctx.Err(); err != nil {
+		return
+	}
+
+	recordsCh := make(chan []database.Record, 1)
+	errCh := make(chan error, 1)
+	go func() {
+		recs, err := s.Query(q)
+		if err != nil {
+			errCh <- err
```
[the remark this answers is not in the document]
mcrakhman: lets not use channels and make it synchronous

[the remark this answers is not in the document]
KirillSto: I have removed QueryWithCtx implementation, because it does not solve the problem with ctx cancellation

space/internal/components/migration/runner.go
```diff
@@ -0,0 +1,116 @@
...
+	if err := r.run(systemobjectreviser.Migration{}, readonlyfixer.Migration{}); err != nil {
+		log.Error("failed to run default migrations", zap.String("spaceId", r.spc.Id()), zap.Error(err))
+	}
+}
+
+func (r *Runner) run(migrations ...Migration) (mErr error) {
+	spaceId := r.spc.Id()
+
+	for _, m := range migrations {
+		toMigrate, migrated, err := m.Run(r.ctx, r.store, r.spc)
```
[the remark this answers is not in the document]
KirillSto: Done

space/internal/components/migration/runner.go
KirillSto: I have added wrapError function to apply format

space/internal/components/migration/runner.go
KirillSto: Thank you! Changed the name

space/internal/components/migration/readonlyfixer/relationsfixer.go
KirillSto: I have added log as input parameter of Migration.Run, so migration name is displayed as zap string field of log message

pkg/lib/localstore/objectstore/queries.go
KirillSto: Done!

space/internal/components/migration/readonlyfixer/relationsfixer.go
```diff
@@ -0,0 +1,90 @@
...
+			Value: pbtypes.Bool(false),
+		}}
+		e := space.DoCtx(ctx, pbtypes.GetString(r.Details, bundle.RelationKeyId.String()), func(sb smartblock.SmartBlock) error {
+			if ds, ok := sb.(basic.DetailsSettable); ok {
+				return ds.SetDetails(nil, det, false)
+			}
+			return nil
+		})
+		if e != nil {
+			err = multierror.Append(err, fmt.Errorf("failed to set readOnlyValue=true to relation %s in space %s: %w", uk, spaceId, e))
```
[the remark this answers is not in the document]
KirillSto: Thank you! Done

mcrakhman (approved):

space/internal/components/migration/systemobjectreviser/systemobjectreviser.go
```diff
@@ -12,26 +18,52 @@ import (
...
 	}
+
+	for _, details := range spaceObjects {
+		shouldBeRevised, e := reviseSystemObject(ctx, log, space, details, marketObjects)
+		if !shouldBeRevised {
+			continue
+		}
+		toMigrate++
+		if e != nil {
+			err = multierror.Append(err, fmt.Errorf("failed to revise object: %w", e))
```
cheggaaa, line 57: errors.Join

cheggaaa (approved):

PR #1261, opened 2024-05-21
GO-3171 check p2p status and send event

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-3171/check-p2p-status-and-send-event

1. During fetch responsible peers in `clientPeerManager`, we check if peer were connected and send update in case of it was removed
2. Send update of p2p status in case of new discovered peer in `localdiscovery` service
3. In case of no network interfaces with multicast in system we send event that p2p is not possible during local discovery


space/spacecore/localdiscovery/localdiscovery.go
```diff
@@ -194,6 +210,7 @@ func (l *localDiscovery) readAnswers(ch chan *zeroconf.ServiceEntry) {
 	for entry := range ch {
 		if entry.Instance == l.peerId {
 			log.Debug("discovered self")
+			l.peerToPeerStatusUpdater.BroadcastPeerUpdate()
```
mcrakhman, line 213: why do you call this here? does it mean that p2p works if we discovered self?

mcrakhman: why dont u do it with other peers

requilence (approved):

PR #1264, opened 2024-05-23
GO-3427 Fix restrictions of sets and collections

KirillSto:
https://linear.app/anytype/issue/GO-3427/fix-collectionssets-restrictions

We did not set restrictions to sets and collections correctly, because these types of objects do have internal keys, so their layout was not taken into account.
Small refactor of restriction service is also presented


core/block/restriction/fixture_test.go
```diff
@@ -1,61 +1,62 @@
...
+	}
+	uk, _ := domain.NewUniqueKey(sbType, "")
+	return &restrictionHolder{
+		sbType:    sbType,
+		layout:    layout,
+		uniqueKey: uk,
+	}
+}
+
+func givenTemplate(layout model.ObjectTypeLayout) {
```
AnastasiaShemyakinskaya, line 60: Do we need this function?

KirillSto: Oh, forgot to delete it! Thank you!

AnastasiaShemyakinskaya (approved):

KirillSto:
@fat-fellow https://github.com/anyproto/anytype-heart/pull/1264/files#diff-f7a6130a2772d1a2bc016ea4b4708b43adb3f729e3027881786f09b9f2e2bb85R188 
this line conains the very fix

fat-fellow (approved):

PR #1267, opened 2024-05-23
GO-2361 Save activeView to DB

KirillSto:
https://linear.app/anytype/issue/GO-2361/back-to-the-last-view-of-the-set-collection

Save activeView to badger and inject it to dataview blocks on smartblock Init/Apply


core/block/editor/smartblock/smartblock.go
```diff
@@ -1346,6 +1351,40 @@ func hasDetailsMsgs(msgs []simple.EventMessage) bool {
...
+			i++
+		}
+	}
+	// we exclude BlockDataViewActiveViewIsSet messages, because we do not need to send events on it
+	return msgs[:i]
+}
+
+func (sb *smartBlock) injectActiveViews(s *state.State) {
+	if err := s.Iterate(func(b simple.Block) (isContinue bool) {
+		if d := b.Model().GetDataview(); d != nil {
```
requilence, line 1374: it feels like, instead of iterating all blocks for all smartblocks when opening them the reverse approach may be more efficient and also utilise the objectstore indexes.
- get all saved activeViews from objectstore for specific objectId
- then pick specific blocks and modify them

pb/protos/events.proto
```diff
@@ -902,6 +903,11 @@ message Event {
                 string id = 1; // dataview block's id
                 bool value = 2;
             }
+
+            message ActiveViewIsSet {
```
requilence, line 907: better be ActiveViewSet? because the last in our naming is just a verb

requilence (changes requested):

pb/protos/events.proto
KirillSto: Thank you, done!

core/block/editor/smartblock/smartblock.go
KirillSto: Yes, I have rewritten this logic, so we store separate map-like string for each object

cheggaaa (changes requested):
1. Save active view in dataview.SetActiveView
2. Move the injectViews method to editor/dataview

cheggaaa (approved):

requilence (approved):

PR #1271, opened 2024-05-24
GO-2140 Event objectClose on logout and object deletion

KirillSto:
https://linear.app/anytype/issue/GO-2140/events-for-closed-subscriptions-and-documents

Send objectClose event on deletion and logout


core/block/delete.go
```diff
@@ -78,6 +78,7 @@ func (s *Service) DeleteObjectByFullID(id domain.FullID) (err error) {
 	}
 
 	sendOnRemoveEvent(s.eventSender, id.ObjectID)
+	sendOnCloseEvent(s.eventSender, id.ObjectID)
```
fat-fellow, line 81: can we test it?

KirillSto: Yeah, I have added unit-test on event broadcasting and also moved deletion logics of derived objects to separate func

core/block/delete.go
```diff
@@ -33,33 +34,11 @@ func (s *Service) DeleteObjectByFullID(id domain.FullID) (err error) {
...
-		if err != nil {
-			return fmt.Errorf("on delete: %w", err)
-		}
-		if sbType == coresb.SmartBlockTypeRelation {
-			err := s.deleteRelationOptions(relationKey)
-			if err != nil {
-				return fmt.Errorf("failed to delete relation options of deleted relation: %w", err)
-			}
-		}
+	case coresb.SmartBlockTypeObjectType, coresb.SmartBlockTypeRelation,
```
fat-fellow, line 39: align in a column for readability

fat-fellow (approved):

core/block/delete.go
KirillSto: Done!

cheggaaa (changes requested):
We should make it more reliable. 
Good candidates to use:
```
smartblock.Close()
smartblock.ObjectClose()
smartblock.ObjectCloseAllSessions()
```

cheggaaa (approved):

PR #1274, opened 2024-05-27
GO-3453: progress bars protocol

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-3453/[next]-progress-bars-protocol


core/device/networkstate.go
```diff
@@ -45,14 +45,13 @@ func (n *networkState) GetNetworkState() model.DeviceNetworkType {
...
-	defer n.networkMu.Unlock()
-
 	if n.networkState == networkState {
 		// to avoid unnecessary hook calls
 		return
 	}
 	n.networkState = networkState
-	n.runOnNetworkUpdateHook()
+	n.networkMu.Unlock()
+	n.runOnNetworkUpdateHook(networkState)
```
requilence, line 54: deadlock in case of networkState has not changed

core/block/process/service.go
```diff
@@ -85,13 +97,33 @@ func (s *service) monitor(p Process) {
...
+	}
 	var prevInfo = info
 	for {
 		select {
+		case <-updated:
+			info := p.Info()
+			if !infoEquals(info, prevInfo) {
+				s.eventSender.BroadcastExceptSessions(&pb.Event{
+					Messages: []*pb.EventMessage{
+						{
```
requilence, line 112: maybe we can avoid reading update in this loop at all? It will block the progress.Update call with some chance, even in case of buffered chan. Especially for the operations that call Update very frequently.

Instead you can leave the ticker only and update the info under the lock?

requilence (changes requested):

core/device/networkstate.go
AnastasiaShemyakinskaya: I'll remove this code, it was just a test and doesn't relate to progress bar. May be I committed it by mistake

core/block/process/service.go
AnastasiaShemyakinskaya: I added new channel because sometimes `done` event can be send before the update event from ticker, so progress bar is finished with wrong information. So we can avoid such problems if during finish with assume, that totalCount == doneCount and put this information to `Info`

requilence (approved):

PR #1275, opened 2024-05-27
GO-3442: web-clipper doesn't save full page

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-3442/web-clipper-doesnt-save-full-page

1. Web clipper doesn't save all images. That's because some web sites use `data-src` attributes instead of `src`
2. So I added support for `data-src` in html converter


fat-fellow (approved):

core/block/import/markdown/anymark/markdown.go
```diff
@@ -233,6 +233,15 @@ func getCustomHTMLRules() []html2md.Rule {
 		simpleText, blockquote, italic, code, bdo, div, img, table}
 }
 
+func extractImageSource(src string, ok bool, selec *goquery.Selection) string {
```
KirillSto, line 236: It seems we don't need **ok** as input parameter here

KirillSto (changes requested):

core/block/import/markdown/anymark/markdown.go
AnastasiaShemyakinskaya: Fixed

core/block/import/markdown/anymark/markdown.go
```diff
@@ -233,6 +230,19 @@ func getCustomHTMLRules() []html2md.Rule {
 		simpleText, blockquote, italic, code, bdo, div, img, table}
 }
 
+func extractImageSource(selec *goquery.Selection) string {
+	var (
+		src string
+		ok  bool
+	)
```
KirillSto, line 237: These declarations are redundunt as we initialize both of these variables in line 238

KirillSto (approved):

core/block/import/markdown/anymark/markdown.go
AnastasiaShemyakinskaya: We need it, because we return src in the end of the function

PR #1288, opened 2024-05-31
GO-3515: cannot export object with linked objects

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-3515/cannot-export-object-with-linked-objects
1. Skip export of recommended relation if it's not exist in object store
2. Add optimization - now we try to get derived objects only from object types or templates, because currently for relations it's useless as they have only system relations and type

### DIFF
--- a/core/block/export/export.go
+++ b/core/block/export/export.go
@@ -688,29 +688,32 @@ func (e *export) cleanupFile(wr writer) {
 }
 
 func (e *export) getRelatedDerivedObjects(objects map[string]*types.Struct) ([]database.Record, error) {
-	derivedObjects, err := e.iterateObjects(objects)
+	derivedObjects, typesAndTemplates, err := e.iterateObjects(objects)
 	if err != nil {
 		return nil, err
 	}
-	if len(derivedObjects) > 0 {
+	// get derived objects only from types and templates,
+	// because relations currently have only system relations and object type
+	if len(typesAndTemplates) > 0 {
 		derivedObjectsMap := make(map[string]*types.Struct, 0)
-		for _, object := range derivedObjects {
+		for _, object := range typesAndTemplates {
 			id := object.Get(bundle.RelationKeyId.String()).GetStringValue()
 			derivedObjectsMap[id] = object.Details
 		}
-		iteratedObjects, err := e.iterateObjects(derivedObjectsMap)
+		iteratedObjects, typesAndTemplates, err := e.iterateObjects(derivedObjectsMap)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


core/block/export/export.go
```diff
@@ -688,29 +688,32 @@ func (e *export) cleanupFile(wr writer) {
...
 			return nil, err
 		}
 		derivedObjects = append(derivedObjects, iteratedObjects...)
+		derivedObjects = append(derivedObjects, typesAndTemplates...)
 	}
 	return derivedObjects, nil
 }
 
-func (e *export) iterateObjects(objects map[string]*types.Struct) ([]database.Record, error) {
+func (e *export) iterateObjects(objects map[string]*types.Struct) ([]database.Record, []database.Record, error) {
```
KirillSto, line 713: It seems we can use named return values for readability

core/block/export/export.go
```diff
@@ -767,68 +770,78 @@ func (e *export) isObjectWithDataview(details *types.Struct) bool {
 
 func (e *export) processObject(object *types.Struct,
 	derivedObjects []database.Record,
+	typesAndTemplates []database.Record,
 	relations []string,
-) ([]database.Record, error) {
+) ([]database.Record, []database.Record, error) {
```
KirillSto, line 775: And here as well

core/block/export/export.go
```diff
@@ -767,68 +770,78 @@ func (e *export) isObjectWithDataview(details *types.Struct) bool {
...
+	derivedObjects, err = e.handleSetOfRelation(object, derivedObjects)
+	if err != nil {
+		return nil, nil, err
 	}
-	return e.handleSetOfRelation(object, derivedObjects)
+	return derivedObjects, typesAndTemplates, nil
 }
 
-func (e *export) addObjectType(objectTypeId string, derivedObjects []database.Record) ([]database.Record, error) {
+func (e *export) addObjectType(objectTypeId string, derivedObjects []database.Record, typesAndTemplates []database.Record) ([]database.Record, []database.Record, error) {
```
KirillSto, line 807: And here

core/block/export/export.go
```diff
@@ -918,7 +931,7 @@ func (e *export) getRelationOptions(relationKey string) ([]database.Record, erro
 	return relationOptionsDetails, nil
 }
 
-func (e *export) addTemplates(id string, derivedObjects []database.Record) ([]database.Record, error) {
+func (e *export) addTemplates(id string, derivedObjects []database.Record, typesAndTemplates []database.Record) ([]database.Record, []database.Record, error) {
```
KirillSto, line 934: And here I think

KirillSto (approved):

PR #1289, opened 2024-05-31
GO-3554 refactor ft indexation

requilence:
- check existing doc content before writing the update to the index to avoid unnecessary memory and disk usage
- use ID with prefix query to find all docs for objects instead of a separate field DocID
- refactor batcher to batch doc deletions together with updates


pkg/lib/localstore/ftsearch/ftsearch.go
```diff
@@ -118,10 +120,26 @@ func (f *ftSearch) Index(doc SearchDoc) (err error) {
...
+
+	return f.index.Batch(batch)
+}
+
+func (f *ftSearch) BatchIndex(ctx context.Context, docs []SearchDoc, deletedDocs []string) (err error) {
 	if len(docs) == 0 {
 		return nil
 	}
+	defer func() {
+		b, _ := f.index.Stats().MarshalJSON()
```
fat-fellow, line 140: fmt.Printf?

fat-fellow (changes requested):
fmt.printf

fat-fellow (approved):

PR #1299, opened 2024-06-05
GO-3555: payments refactoring

AnthonyAkentiev:
---

---

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [x] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [x] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/core/payments/payments.go
+++ b/core/payments/payments.go
@@ -221,7 +221,7 @@ func getCacheExpireTime(dateEnds time.Time) time.Time {
 // 1. Check in cache. if req.NoCache -> do not check in cache.
 // 2. If found in cache -> return it
 // 3. Ask PP node
-// 4a. If PP node didn't answer and we are have membership -> return error
+// 4a. If PP node didn't answer and we have membership -> return it
 // 4b. If PP node didn't answer -> create empty response
 // 5. Save to cache. Lifetime - min(subscription ends, 24h)
 // 6. If tier or status has changed -> send event
@@ -238,7 +238,7 @@ func (s *service) GetSubscriptionStatus(ctx context.Context, req *pb.RpcMembersh
 	// 1 - check in cache
 	// if cache is disabled -> will return objects and ErrCacheDisabled
 	// if cache is expired -> will return objects and ErrCacheExpired
-	cachedStatus, tiers, err := s.cache.CacheGet()
+	cachedStatus, _, err := s.cache.CacheGet()
 
 	// if NoCache flag -> skip returning from cache
 	if !req.NoCache && (err == nil) && (cachedStatus != nil) && (cachedStatus.Data != nil) {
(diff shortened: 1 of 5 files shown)


core/payments/payments.go
```diff
@@ -192,6 +192,42 @@ func (s *service) sendEvent(status *pb.RpcMembershipGetStatusResponse) {
...
+	// sub end < now OR no sub end provided (unlimited)
+	if isExpired {
+		log.Debug("setting cache to +1 day because subscription is isExpired")
+		return timeNext
+	}
+
+	// sub end >= now
+	// return min(sub end, now + 24h)
+	if dateEnds.Before(timeNext) {
+		log.Debug("setting cache to +1 day because subscription ends soon")
```
KirillSto, line 213: It seems that this message is incorrect, as we are setting expiration time to less than one day. Moreover, cacheLifetimeDur can be changed, so we should consider different lifetime here

core/payments/payments.go
```diff
@@ -192,6 +192,42 @@ func (s *service) sendEvent(status *pb.RpcMembershipGetStatusResponse) {
 	})
 }
 
+func getCacheExpireTime(dateEnds time.Time) time.Time {
```
KirillSto, line 195: Maybe it is better to move this logic to separate func of cacheservice?
CacheSet could accept dateEnds except of expireTime, for example, and all internal logics of expireTime calculation will be handled by the very cache

core/payments/payments.go
```diff
@@ -192,6 +192,42 @@ func (s *service) sendEvent(status *pb.RpcMembershipGetStatusResponse) {
...
+	const cacheLifetimeDur = 24 * time.Hour
```
KirillSto, line 196: Should not we define this constant in the beginning of the file

core/payments/payments.go
```diff
@@ -236,18 +274,25 @@ func (s *service) GetSubscriptionStatus(ctx context.Context, req *pb.RpcMembersh
...
+		/*
+			// 4a. If PP node didn't answer and we have membership -> return error
+			if tiers != nil && tiers.Tiers != nil && len(tiers.Tiers) > 0 {
+				if tiers.Tiers[0].Id != uint32(proto.SubscriptionTier_TierExplorer) {
+					// return error
+					log.Error("returning error in get status", zap.Error(err))
+					return nil, err
+				}
 			}
+		*/
```
KirillSto, line 286: This code could be deleted?

KirillSto (approved):

PR #1309, opened 2024-06-12
GO-3388: New offload all files logic

deff7:


core/files/fileoffloader/offloader.go
```diff
@@ -108,30 +110,41 @@ func (s *service) FilesOffload(ctx context.Context, objectIds []string, includeN
...
+			if err != nil {
+				return 0, 0, fmt.Errorf("not pinned file: collect cids: %w", err)
+			}
+			gc.MarkAsUsing(cids)
 		}
 	}
-	return filesOffloaded, totalSize, nil
+
+	err = gc.CollectGarbage(ctx)
+	return 0, 0, err
```
AnastasiaShemyakinskaya, line 147: Is it okay that we always return 0 and 0?

deff7: Thank you. Afaik, clients don't use these fields. I'd better to remove them

AnastasiaShemyakinskaya (approved):

PR #1314, opened 2024-06-13
GO-3192 | GO-3629 - Refactor table block normalization

KirillSto:
https://linear.app/anytype/issue/GO-3192/sometimes-clicking-an-object-doesnt-open-it-instead-it-leads-to-the
https://linear.app/anytype/issue/GO-3629/fix-tables-normalization

Some objects containing tables are broken, as Rows object became the parent of table cell object.
This PR contains changes on table block normalization method, that exclude any errors on incorrect state.

- If table root is missed or contains incorrect children -> drop the table
- If some row or column block is missed -> it is recreated
- If some block is found among children of incorrect block (e.g. cell is in rows) -> it is moved under the table
- If some cell has incorrect id -> it is moved under the table


fat-fellow (approved):

core/block/editor/table/table.go
```diff
@@ -261,6 +278,22 @@ func (tb Table) Iterate(f func(b simple.Block, pos CellPosition) bool) error {
 	return nil
 }
 
+func (tb Table) MoveBlocksUnderTheTable(ids ...string) {
+	parent := tb.s.PickParentOf(tb.block.Model().Id)
```
deff7, line 282: GetParentOf?

KirillSto: Fixed!

deff7 (approved):

PR #1321, opened 2024-06-17
GO-3564 Performance: Run profiler method

deff7:


core/debug/profiler/service.go
```diff
@@ -41,6 +53,83 @@ func (s *service) Run(ctx context.Context) (err error) {
 	return nil
 }
 
+func (s *service) RunProfiler(ctx context.Context, seconds int) (string, error) {
+	var tracerBuf bytes.Buffer
```
mcrakhman, line 57: Maybe we should add debug stats data

PR #1328, opened 2024-06-20
GO-3564 Login trace

deff7:


fat-fellow (changes requested):
Need to add the ability to get login trace after login finishes.

fat-fellow (approved):

PR #1330, opened 2024-06-20
GO-3543 add new sync status for files that doesn't fit limit

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-3543/add-new-sync-status-for-files-that-doesnt-fit-limit

1. During file indexing I evaluate left bytes in user account storage and update space and object status according to amount of left bytes
2. If 99% is occupied, then we send error StorageLimitExceed to space and reset it, so it will not be shown always
3. Refactoring of File and Object state for space sync service


core/filestorage/filesync/upload.go
```diff
@@ -77,7 +77,8 @@ func (s *fileSync) handleLimitReachedError(err error, it *QueueItem) *errLimitRe
 			log.Error("set limit reached error logged", zap.String("objectId", it.ObjectId), zap.Error(setErr))
 		}
 
-		s.runOnLimitedHook(it.ObjectId, it.FullFileId())
+		bytesLeftPercentage := float64(limitReachedErr.accountLimit-limitReachedErr.totalBytesUsage) / float64(limitReachedErr.accountLimit)
```
deff7, line 80: I'm concerned that accountLimit could be zero due to some network error or bug in limit calculations, maybe add zero value check?

core/filestorage/filesync/upload.go
```diff
@@ -236,9 +237,9 @@ func (s *fileSync) runOnUploadStartedHook(fileObjectId string, fileId domain.Ful
 	return nil
 }
 
-func (s *fileSync) runOnLimitedHook(fileObjectId string, fileId domain.FullFileId) {
+func (s *fileSync) runOnLimitedHook(fileObjectId string, fileId domain.FullFileId, bytesLeft float64) {
```
deff7, line 240: maybe rename bytesLeft to bytesLeftPercentage here and further in code?

deff7 (approved):

PR #1332, opened 2024-06-20
GO-3514 custom types that are deleted are impossible to bring back

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-3514/custom-types-that-are-deleted-are-impossible-to-bring-back-through

1. Check during creation of derived objects that it was deleted before
2. If so, I create new unique key and according new object
3. Then I update object type or relations with new key


core/block/import/common/objectcreator/objectcreator.go
```diff
@@ -111,6 +111,7 @@ func (oc *ObjectCreator) Create(dataObject *DataObject, sn *common.Snapshot) (*t
 		log.With("objectID", newID).Errorf("failed to update objects ids: %s", err)
 	}
 
+	oc.updateKeys(st, oldIDtoNew)
```
KirillSto, line 114: Just to confirm. Does not oldIDtoNew map is getting filled by object type ids and not unique keys?

KirillSto (changes requested):

KirillSto (approved):

PR #1361, opened 2024-07-02
GO-3192 Add check on moving table blocks

KirillSto:
https://linear.app/anytype/issue/GO-3192/sometimes-clicking-an-object-doesnt-open-it-instead-it-leads-to-the

We should add check on moving table blocks. Right now only these operations are allowed:
- we could move table rows between each other
- we could move table root block whereever we want

Other moving operations on table blocks are prohibited, so:
- in case of moving some blocks TO some table block that is not root block, we should move these blocks under the table
- we should return error, if client is trying to move some table block except root and row

---

PR also includes:
- table package unit tests refactoring. Now coverage is >90%
- fix of panic on creation of table with 0 rows and `withHeaderRow=true`
- check of parameters on RowDuplicate, as client could insert dup of some row in wrong place


core/block/editor/basic/basic.go
```diff
@@ -278,6 +287,48 @@ func (bs *basic) Move(srcState, destState *state.State, targetBlockId string, po
...
+			t, err := table.NewTable(st, root.Model().Id)
+			if err != nil {
+				return "", 0, ErrCannotMoveTableBlocks
+			}
+
+			// we allow moving rows between each other
+			rows := t.RowIDs()
+			if slices.Contains(rows, id) && slices.Contains(rows, target) {
+				if pos == model.Block_Bottom || pos == model.Block_Top {
+					return target, pos, nil
```
requilence, line 303: Probably you don't want to return here(and other places)  in case of success? Because you are iterating the slice of BlockIds

requilence (dismissed):

core/block/editor/table/table.go
```diff
@@ -901,6 +147,19 @@ func NewTable(s *state.State, id string) (*Table, error) {
 	return &tb, nil
 }
 
+// GetTableRootBlock iterates over parents of block. Returns nil in case root table block is not found
+func GetTableRootBlock(s *state.State, id string) (block simple.Block) {
```
deff7, line 151: PickTableRootBlock as Pick in state

core/block/editor/table/table.go
```diff
@@ -1006,3 +265,38 @@ func (tb Table) Iterate(f func(b simple.Block, pos CellPosition) bool) error {
 	}
 	return nil
 }
+
+// CheckTableBlocksMove checks if Insert operation is allowed in case table blocks are affected
+func CheckTableBlocksMove(st *state.State, target string, pos model.BlockPosition, blockIds []string) (string, model.BlockPosition, error) {
+	// nolint:errcheck
+	if t, _ := NewTable(st, target); t != nil {
```
deff7, line 272: you could just check error (err == nil && t != nil)

core/block/editor/table/table.go
```diff
@@ -1006,3 +265,38 @@ func (tb Table) Iterate(f func(b simple.Block, pos CellPosition) bool) error {
...
+		t := GetTableRootBlock(st, id)
+		if t != nil && t.Model().Id != id {
+			// we should not move table blocks except table root block
+			return "", 0, ErrCannotMoveTableBlocks
+		}
+	}
+
+	t := GetTableRootBlock(st, target)
+	if t != nil && t.Model().Id != target {
+		// we allow inserting blocks into table cell
```
deff7, line 292: we don't allow this, table cell have to be text block without children. But it's a client limitation, so if client devs agreed to support nested block we can support this as well

core/block/editor/table/table.go
KirillSto: Done!

core/block/editor/table/table.go
KirillSto: Thank you! Done

core/block/editor/table/table.go
KirillSto: Yes, u r right! Client only are allowed to c/p blocks into table cell, but any way it is being pasted as plaintext.
So in case we are trying to insert some blocks into the table cell, we should insert them under the table

core/block/editor/basic/basic.go
KirillSto: Fixed

core/block/editor/table/editor.go
```diff
@@ -0,0 +1,781 @@
...
+	}
+
+	if !s.Unlink(req.TargetId) {
+		return fmt.Errorf("unlink row block")
+	}
+	return nil
+}
+
+func (t *editor) RowDuplicate(s *state.State, req pb.RpcBlockTableRowDuplicateRequest) (newRowID string, err error) {
+	if req.Position != model.Block_Top && req.Position != model.Block_Bottom {
```
KirillSto, line 236: I have added this check to exclude insertion of new Row block in some wrong place relatively to old row

core/block/editor/table/editor.go
```diff
@@ -0,0 +1,781 @@
...
+		return "", fmt.Errorf("position %s is not supported", model.BlockPosition_name[int32(req.Position)])
+	}
+	srcRow, err := pickRow(s, req.BlockId)
+	if err != nil {
+		return "", fmt.Errorf("pick source row: %w", err)
+	}
+
+	if _, err = pickRow(s, req.TargetId); err != nil {
```
KirillSto, line 244: This check guarantees we insert new row under or above another row

core/block/editor/table/editor.go
```diff
@@ -0,0 +1,781 @@
...
+		},
+	})
+	if !s.Add(rowsLayout) {
+		return "", fmt.Errorf("add rows block")
+	}
+
+	tableBlock.Model().ChildrenIds = []string{columnsLayout.Model().Id, rowsLayout.Model().Id}
+
+	if req.WithHeaderRow {
+		if len(rowIDs) == 0 {
```
KirillSto, line 134: I have added this check to exclude panic, if `rows==0`

core/block/editor/table/table.go
```diff
@@ -1006,3 +259,32 @@ func (tb Table) Iterate(f func(b simple.Block, pos CellPosition) bool) error {
 	}
 	return nil
 }
+
+// CheckTableBlocksMove checks if Insert operation is allowed in case table blocks are affected
+func CheckTableBlocksMove(st *state.State, target string, pos model.BlockPosition, blockIds []string) (string, model.BlockPosition, error) {
+	if t, err := NewTable(st, target); err == nil && t != nil {
+		// we allow moving rows between each other
+		if slice.ContainsAll(t.RowIDs(), append(blockIds, target)...) {
```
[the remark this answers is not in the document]
KirillSto: Thank you! Did not find it in lo at first glance :see_no_evil:

fat-fellow (approved):

deff7 (approved):

PR #1365, opened 2024-07-03
feat: save grpc ports to listening.json

hellodword:
---

---

### Description

Currently, anytype-ts [listens on dynamic ports](https://github.com/anyproto/anytype-ts/blob/ab64f57309161f27d917b1f88414aa366ff7701f/electron/js/server.js#L30). This is probably beneficial as it can avoid some weird port conflict issues. However, it makes scenarios that need to retrieve these ports more difficult, such as [this code](https://github.com/anyproto/anytype-ts/blob/ab64f57309161f27d917b1f88414aa366ff7701f/go/nativeMessagingHost.go#L225). I find it complex and prone to failure, for example:

```
$ lsof -i -P -n | grep LISTEN | grep -i anytype

$ sudo lsof -i -P -n | grep LISTEN | grep -i anytype
anytypeHe 120603          ubuntu  12u  IPv4 1631878      0t0  TCP 127.0.0.1:44855 (LISTEN)
anytypeHe 120603          ubuntu  13u  IPv4 1631882      0t0  TCP 127.0.0.1:39355 (LISTEN)
anytypeHe 120603          ubuntu  43u  IPv4 1634872      0t0  TCP 127.0.0.1:33357 (LISTEN)
```

I think a mechanism could be added to handle this better. In the meantime, I believe writing a file to the userdata directory is acceptable.

```
$ ANYTYPE_DATA_PATH=/tmp go run ./cmd/grpcserver/ 127.0.0.1:0 127.0.0.1:0
mw grpc: build on 0001-01-01 00:00:00 +0000 UTC at #
gRPC server started at: 127.0.0.1:44285
gRPC Web proxy started at: 127.0.0.1:45713

$ cat /tmp/listening.json 
{"grpc":"127.0.0.1:44285","updated":1720004324,"web":"127.0.0.1:45713"}
```

If this pull request can eventually be accepted, I would be happy to modify the previously mentioned `nativeMessagingHost.go`. 

More use case, see: https://github.com/orgs/anyproto/discussions/211

What do you think?

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/cmd/grpcserver/grpc.go
+++ b/cmd/grpcserver/grpc.go
@@ -5,10 +5,13 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net"
 	"net/http"
+	"path"
+
 	// nolint: gosec
 	_ "net/http/pprof"
 	"os"
@@ -49,12 +52,16 @@ var commonOSSignals = []os.Signal{os.Interrupt, syscall.SIGTERM, syscall.SIGINT}
 func main() {
 	var addr string
 	var webaddr string
+	var dataPath string
 	app.StartWarningAfter = time.Second * 5
 	fmt.Printf("mw grpc: %s\n", vcs.GetVCSInfo().Description())
 	if len(os.Args) > 1 {
 		addr = os.Args[1]
 		if len(os.Args) > 2 {
 			webaddr = os.Args[2]
+			if len(os.Args) > 3 {
+				dataPath = os.Args[3]
+			}
 		}
 	}
 
@@ -71,6 +78,16 @@ func main() {
 			webaddr = env
 		} else {
 			webaddr = defaultWebAddr
+		}
+	}
+
+	if dataPath == "" {
+		if env := os.Getenv("ANYTYPE_DATA_PATH"); env != "" {
+			dataPath = env
+		} else {
+			if appData, err := os.UserConfigDir(); err == nil {
+				dataPath = path.Join(appData, "anytype")
+			}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


deff7:
Hi! Thank you for contribution. There is a way to select specific port via ENV variable, example for macos:

```
open --env ANYTYPE_GRPCWEB_ADDR=localhost:31008 --env ANYTYPE_USE_SIDE_SERVER=http://localhost:31008 /Applications/Anytype.app
```

`ANYTYPE_USE_SIDE_SERVER` is variable for desktop app and should be the same as `ANYTYPE_GRPCWEB_ADDR`

all options for anytype-heart:
- ANYTYPE_GRPC_ADDR -- this port is actually not is use, as desktop app uses grpc-web gateway
- ANYTYPE_GRPCWEB_ADDR -- this port for desktop application
- ANYTYPE_GATEWAY_ADDR -- this port for serving images and files to desktop application

deff7:
Oh, I see. Then maybe prioritize ENV variables? Something like this:

```
	if env := os.Getenv("ANYTYPE_GRPCWEB_ADDR"); env != "" {
		webaddr = env
	} else if webaddr == "" {
		webaddr = defaultWebAddr
	}
```

requilence:
I think the overall idea of saving RPC ports to the json file is good and reliable. 
On windows port ranges got blocked from time to time, so it makes sense to try to reuse the port but in case we are not able to do it, we can take another ones.

Some comments regarding the PR:
- it's better to use the go struct so this config will be typed
- you need to add ANYTYPE_GATEWAY_ADDR as well. We are using it to run http gateway to serve files to the client
- call the config more general "config.json" so we can extend it later

PR #1370, opened 2024-07-04
GO-3699: sync details change lastModifiedDate/lastModifiedBy values and create a change

AnastasiaShemyakinskaya:
1. Don't add change if relation link for local relation added/removed
2. Move relation links to separate package to avoid import cycle

### DIFF
--- a/core/block/editor/state/change.go
+++ b/core/block/editor/state/change.go
@@ -599,24 +599,24 @@ func (s *State) fillChanges(msgs []simple.EventMessage) {
 		})
 	}
 	if len(newRelLinks) > 0 {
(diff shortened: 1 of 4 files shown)


pkg/lib/bundle/init.go
```diff
@@ -225,3 +225,7 @@ func RelationKeyFromID(id string) (domain.RelationKey, error) {
 
 	return "", fmt.Errorf("invalid type url: no prefix found")
 }
+
+func IsLocalRelations(relationKey string) bool {
+	return slices.Contains(LocalRelationsKeys, relationKey)
```
deff7, line 230: probably derived relation should be skipped too

deff7 (approved):

deff7 (approved):

PR #1383, opened 2024-07-10
GO-3747 stacktrace logger improvements

requilence:
- send most actual stacktrace when request is finished
- double check if it really contains the method goroutine

### DIFF
--- a/metrics/interceptors.go
+++ b/metrics/interceptors.go
@@ -65,10 +65,6 @@ var excludedLongExecutionMethods = []string{
 	"DebugRunProfiler",
 }
 
-func stackTraceHasMethod(method string, stackTrace []byte) bool {
-	return bytes.Contains(stackTrace, []byte("core.(*Middleware)."+method+"("))
-}
-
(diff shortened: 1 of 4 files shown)


metrics/interceptors.go
```diff
@@ -1,23 +1,30 @@
...
-	unexpectedErrorCode      = -1
-	parsingErrorCode         = -2
-	defaultUnaryWarningAfter = time.Second * 3
+	unexpectedErrorCode = -1
+	parsingErrorCode    = -2
 )
 
+var (
+	// every duration will be added to the previous ones
+	UnaryTraceCollections = []time.Duration{time.Second * 3, time.Second * 7, time.Second * 10, time.Second * 30, time.Second * 50} // 3, 10, 20, 50, 100
```
fat-fellow, line 24: misleading comment
// 3, 10, 20, 50, 100
this is not true

requilence: @fat-fellow because it runs after each other, the durations are cumulative

fat-fellow (approved):

PR #1386, opened 2024-07-11
GO-3756 add required internal relations to all sbtype of objects

requilence:
We should add requiredInternalRelations to all objects, even a system ones. Because we set some local details such as "syncStatus" or "spaceId" to them. Currently this leads to ignoring of changes and creating a snapshot


fat-fellow (changes requested):
tests

fat-fellow (approved):

fat-fellow (approved):

core/block/editor/state/state.go
```diff
@@ -1934,12 +1934,23 @@ func (s *State) SelectRoots(ids []string) []string {
 }
 
 func (s *State) AddBundledRelations(keys ...domain.RelationKey) {
-	links := make([]*model.RelationLink, 0, len(keys))
+	if len(keys) == 0 {
+		return
+	}
+	added := false
+	relLinks := s.GetRelationLinks()
```
deff7, line 1941: GetRelationLinks copies from parent state making optimization not so useful

PR #1393, opened 2024-07-14
GO-3674 Create file object on copying block to other space

KirillSto:
https://linear.app/anytype/issue/GO-3674/copy-file-blocks-between-spaces

On copying file blocks from one space to another we should create new file object with same hash and keys

### DIFF
--- a/core/block/editor/clipboard/clipboard.go
+++ b/core/block/editor/clipboard/clipboard.go
@@ -19,6 +19,7 @@ import (
 	"github.com/anyproto/anytype-heart/core/block/simple"
 	"github.com/anyproto/anytype-heart/core/block/simple/text"
 	"github.com/anyproto/anytype-heart/core/converter/html"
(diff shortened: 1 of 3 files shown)


core/block/editor/clipboard/clipboard.go
```diff
@@ -585,6 +589,34 @@ func (cb *clipboard) newHTMLConverter(s *state.State) *html.HTML {
 	return html.NewHTMLConverter(cb.fileService, s, cb.fileObjectService)
 }
 
+func (cb *clipboard) processFileBlock(f *model.BlockContentOfFile) {
+	if cb.fileObjectService == nil {
```
deff7, line 593: it have to be initialized at this moment

KirillSto: Removed this check

deff7 (approved):

PR #1395, opened 2024-07-15
GO-3765 membership: extra check for error + refactoring

AnthonyAkentiev:
Somehow cash ended up with returning error - 
cache_returned_error

---

---

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [ ] 👍 yes
- [x] 🙋 no, because I need help

### Added to documentation?

- [ ] 📜 README.md


core/payments/payments.go
```diff
@@ -769,8 +773,10 @@ func (s *service) getAllTiers(ctx context.Context, req *pb.RpcMembershipGetTiers
 
 	// if NoCache -> skip returning from cache
 	if !req.NoCache && (err == nil) && (cachedTiers != nil) && (cachedTiers.Tiers != nil) {
-		log.Debug("returning tiers from cache", zap.Error(err), zap.Any("cachedTiers", cachedTiers))
-		return cachedTiers, nil
+		if (cachedTiers.Error == nil) || (cachedTiers.Error.Code == pb.RpcMembershipGetTiersResponseError_NULL) {
```
KirillSto, line 776: It seems we can simplify the beginning of func a little bit, as cachedTiers is used only when NoCache is false. So lines 770-780 could be:

```
	// 1 - check in cache in case NoCache is False
	if !req.NoCache {
		_, cachedTiers, err := s.cache.CacheGet()
		if err == nil && cachedTiers != nil && cachedTiers.Tiers != nil &&
			(cachedTiers.Error == nil || cachedTiers.Error.Code == pb.RpcMembershipGetTiersResponseError_NULL) {
			log.Debug("returning tiers from cache", zap.Error(err), zap.Any("cachedTiers", cachedTiers))
			return cachedTiers, nil
		}
	}
```
It is safe that we have check **cache==nil** and checks on cache fields in one logical expression, because Go quaranties sequental evaluation

core/payments/payments.go
```diff
@@ -218,9 +218,11 @@ func (s *service) GetSubscriptionStatus(ctx context.Context, req *pb.RpcMembersh
 
 	// if NoCache flag -> skip returning from cache
 	if !req.NoCache && (err == nil) && (cachedStatus != nil) && (cachedStatus.Data != nil) {
-		// 2. If found in cache -> return it
-		log.Debug("returning subscription status from cache", zap.Error(err), zap.Any("cachedStatus", cachedStatus))
-		return cachedStatus, nil
+		if (cachedStatus.Error == nil) || (cachedStatus.Error.Code == pb.RpcMembershipGetStatusResponseError_NULL) {
```
KirillSto, line 221: Maybe we should not call **s.cache.CacheGet** on every request?
Imho we can call it in two cases: when NoCache is not set and when request to PP node is failed. Then code can be simplified a lil bit:

```
	canReturnCachedStatus := func(s *pb.RpcMembershipGetStatusResponse) bool {
		return s != nil && s.Data != nil && (s.Error == nil || s.Error.Code == pb.RpcMembershipGetStatusResponseError_NULL)
	}
	
	// 1 - check in cache
	if !req.NoCache {
		// if cache is disabled -> will return objects and ErrCacheDisabled
		// if cache is expired -> will return objects and ErrCacheExpired
		cachedStatus, _, err := s.cache.CacheGet()
		if err == nil && canReturnCachedStatus(cachedStatus) {
			log.Debug("returning subscription status from cache", zap.Error(err), zap.Any("cachedStatus", cachedStatus))
			return cachedStatus, nil
		}
	}

	...

	status, err := s.ppclient.GetSubscriptionStatus(ctx, &reqSigned)
	if err != nil {
		// 4a. try reading from cache again
		cachedStatus, _, err := s.cache.CacheGet()
		if err == nil && canReturnCachedStatus(cachedStatus) {
			log.Debug("returning subscription status from cache", zap.Error(err), zap.Any("cachedStatus", cachedStatus))
			return cachedStatus, nil
		}

		...
	}


```

core/payments/payments.go
```diff
@@ -268,39 +271,82 @@ func (s *service) GetSubscriptionStatus(ctx context.Context, req *pb.RpcMembersh
...
+		log.Debug("subscription status treated as changed because cache was empty/expired/NoCache flag was passed")
+		return true
+	}
+
+	// 2 - Check if tier or status has changed
+	isDiffTier := false
+	isDiffStatus := false
+	isEmailDiff := false
+
+	if canReturnCachedStatus(cachedStatus) {
```
KirillSto, line 318: It seems we have some biz logic problem here. 
If **canReturnCachedStatus** returns false (e.g. cached status is nil or contains some error), then **isUpdateRequired** returns false. Bit imho in this case status should be updated in cache as well

core/payments/payments.go
```diff
@@ -268,39 +271,91 @@ func (s *service) GetSubscriptionStatus(ctx context.Context, req *pb.RpcMembersh
...
+
+	// 5 - Send all messages to the client if needed
+	if !isUpdateRequired(cacheErr, cachedStatus, status) {
+		// no need to send events or enable/disable cache
+		return &out, nil
+	}
+	s.updateStatus(ctx, status)
+
+	// 6 - Enable or disable cache (only if status has changed)
+	// also automatically will enable cash in N minutes after DisableForNextMinutes() call
```
fat-fellow, line 292: These 2 lines together looks weird

deff7 (approved):

PR #1409, opened 2024-07-22
GO-3635 Inject links along with derived relations + some smartblock refactoring

KirillSto:
https://linear.app/anytype/issue/GO-3635/update-backlinks-reactively-in-the-objectstore

StateAppend did not update links of smartblock, as injectLinks was not called explicitly. Now injectLinks is called inside injectDerivedDetails, so changes are applied for objects on other devices even if they are not uploaded in cache.

Additionally some refactoring in smartblock pkg is made. Details are in the code comments

### DIFF
--- a/core/block/editor/smartblock/smartblock.go
+++ b/core/block/editor/smartblock/smartblock.go
@@ -162,7 +162,6 @@ type SmartBlock interface {
 	CheckSubscriptions() (changed bool)
 	GetDocInfo() DocInfo
 	Restrictions() restriction.Restrictions
-	SetRestrictions(r restriction.Restrictions)
 	ObjectClose(ctx session.Context)
 	ObjectCloseAllSessions()
 
@@ -386,11 +385,7 @@ func (sb *smartBlock) sendObjectCloseEvent(_ ApplyInfo) error {
 
 // updateRestrictions refetch restrictions from restriction service and update them in the smartblock
 func (sb *smartBlock) updateRestrictions() {
-	restrictions := sb.restrictionService.GetRestrictions(sb)
-	sb.SetRestrictions(restrictions)
-}
-
-func (sb *smartBlock) SetRestrictions(r restriction.Restrictions) {
+	r := sb.restrictionService.GetRestrictions(sb)
 	if sb.restrictions.Equal(r) {
 		return
 	}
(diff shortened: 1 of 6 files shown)


core/block/editor/smartblock/smartblock.go
```diff
@@ -162,7 +162,6 @@ type SmartBlock interface {
 	CheckSubscriptions() (changed bool)
 	GetDocInfo() DocInfo
 	Restrictions() restriction.Restrictions
-	SetRestrictions(r restriction.Restrictions)
```
KirillSto, line 165: SetRestrictions method is not used

core/block/editor/smartblock/smartblock.go
```diff
@@ -547,7 +542,7 @@ func (sb *smartBlock) onMetaChange(details *types.Struct) {
 }
 
 // dependentSmartIds returns list of dependent objects in this order: Simple blocks(Link, mentions in Text), Relations. Both of them are returned in the order of original blocks/relations
-func (sb *smartBlock) dependentSmartIds(includeRelations, includeObjTypes, includeCreatorModifier, _ bool) (ids []string) {
+func (sb *smartBlock) dependentSmartIds(includeRelations, includeObjTypes, includeCreatorModifier bool) (ids []string) {
```
KirillSto, line 545: 4th bool argument has no name and is not used

core/block/editor/smartblock/smartblock.go
```diff
@@ -784,7 +777,6 @@ func (sb *smartBlock) ResetToVersion(s *state.State) (err error) {
 	s.SetParent(sb.Doc.(*state.State))
 	sb.storeFileKeys(s)
 	sb.injectLocalDetails(s)
-	sb.injectDerivedDetails(s, sb.SpaceID(), sb.Type())
```
KirillSto, line 787: injectDerivedDetails is called inside Apply, so we do not need first call

core/block/editor/smartblock/smartblock.go
```diff
@@ -639,8 +634,6 @@ func (sb *smartBlock) Apply(s *state.State, flags ...ApplyFlag) (err error) {
 		}
 	}
 
-	sb.beforeStateApply(s)
```
KirillSto, line 642: This method calls sb.setRestrictionsDetail(s) and sb.injectLinksDetails(s). Both methods are also called in injectDerivedDetails in the beginning of function body

core/block/editor/smartblock/smartblock.go
```diff
@@ -1460,6 +1447,7 @@ func (sb *smartBlock) injectDerivedDetails(s *state.State, spaceID string, sbt s
 		s.SetDetailAndBundledRelation(bundle.RelationKeyIsDeleted, pbtypes.Bool(isDeleted))
 	}
 
+	sb.injectLinksDetails(s)
```
KirillSto, line 1450: The very bugfix

deff7 (approved):

PR #1451, opened 2024-08-06
GO-2746: add export structure

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-2746/export-structure

### DIFF
--- a/core/block/export/export.go
+++ b/core/block/export/export.go
@@ -55,8 +55,14 @@ import (
 const CName = "export"
 
 const (
-	tempFileName   = "temp_anytype_backup"
-	spaceDirectory = "spaces"
+	tempFileName              = "temp_anytype_backup"
+	spaceDirectory            = "spaces"
+	typesDirectory            = "types"
+	objectsDirectory          = "objects"
+	relationsDirectory        = "relations"
+	relationsOptionsDirectory = "relationsOptions"
+	templatesDirectory        = "templates"
+	filesObjects              = "filesObjects"
 )
 
 var log = logging.Logger("anytype-mw-export")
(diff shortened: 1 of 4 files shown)


core/block/export/export.go
```diff
@@ -519,15 +525,33 @@ func (e *export) provideMarkdownName(s *state.State, wr writer, docID string, co
...
+	dir := e.provideFileDirectory(blockType)
+	filename := filepath.Join(dir, docID+conv.Ext())
 	if spaceId == "" {
 		spaceId := pbtypes.GetString(st.LocalDetails(), bundle.RelationKeySpaceId.String())
 		filename = filepath.Join(spaceDirectory, spaceId, filename)
 	}
 	return filename
 }
 
+func (e *export) provideFileDirectory(blockType smartblock.SmartBlockType) string {
```
mcrakhman, line 538: call it just `fileDirectory` maybe? and then `fileName`?

core/block/export/export.go
```diff
@@ -519,15 +525,33 @@ func (e *export) provideMarkdownName(s *state.State, wr writer, docID string, co
 	return wr.Namer().Get(path, docID, name, conv.Ext())
 }
 
-func (e *export) provideFileName(docID, spaceId string, conv converter.Converter, st *state.State) string {
-	filename := docID + conv.Ext()
+func (e *export) provideFileName(docID, spaceId string, conv converter.Converter, st *state.State, blockType smartblock.SmartBlockType) string {
+	dir := e.provideFileDirectory(blockType)
+	filename := filepath.Join(dir, docID+conv.Ext())
 	if spaceId == "" {
 		spaceId := pbtypes.GetString(st.LocalDetails(), bundle.RelationKeySpaceId.String())
```
mcrakhman, line 532: Can it be the case that spaceId is nil? What happens here, why if spaceId == "" we do something completely different? Maybe It shouldn't be inside this function? Can you explain please?

mcrakhman (approved):

PR #1453, opened 2024-08-06
GO-3849 Fix IsEmpty for sets

KirillSto:
https://linear.app/anytype/issue/GO-3849/need-to-save-the-setcollection-if-you-have-not-entered-a-name

DeleteIfEmpty internal flag is unset on state.IsEmpty returning false. However this check analyzes only text block contents.
Sets and collections have only standard blocks, so separate logic to check object emptiness must be supported

### DIFF
--- a/core/block/editor/state/state.go
+++ b/core/block/editor/state/state.go
@@ -11,6 +11,7 @@ import (
 	"github.com/ipfs/go-cid"
 
 	"github.com/anyproto/anytype-heart/core/block/simple"
+	"github.com/anyproto/anytype-heart/core/block/simple/dataview"
 	"github.com/anyproto/anytype-heart/core/block/undo"
 	"github.com/anyproto/anytype-heart/core/domain"
 	"github.com/anyproto/anytype-heart/core/session"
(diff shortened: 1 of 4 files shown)


core/block/editor/state/state.go
```diff
@@ -1378,6 +1385,35 @@ func (s *State) IsEmpty(checkTitle bool) bool {
 	return true
 }
 
+// isEmptyDataview is a separate emptiness check for sets and collections, as
```
AnastasiaShemyakinskaya, line 1388: as.....?🤔

core/block/simple/dataview/dataview.go
```diff
@@ -520,3 +522,29 @@ func (s *Dataview) UpdateRelationOld(relationKey string, rel model.Relation) err
...
+	case 1:
+		view = d.content.Views[0]
+	default:
+		return false
+	}
+
+	if view.DefaultObjectTypeId != "" ||
+		view.DefaultTemplateId != "" ||
+		len(view.Filters) != 0 ||
+		len(view.Sorts) > 1 ||
```
AnastasiaShemyakinskaya, line 544: Why do we check `len(view.Sorts)` as more then 1? Just want to understand the logic 😃

core/block/editor/state/state.go
KirillSto: Sorry, started typing and did not finish :smile:

core/block/simple/dataview/dataview.go
KirillSto: On creation of new dataview block we add sorting by lastModified by default:
https://github.com/anyproto/anytype-heart/blob/main/core/block/editor/dataview/dataview.go#L528
so if we find some other sorts in the view, we can consider dataview object as modified

AnastasiaShemyakinskaya (approved):

PR #1461, opened 2024-08-08
[release 6] GO-3737 - Fix type reinstall

KirillSto:
https://linear.app/anytype/issue/GO-3737/after-deleting-the-type-it-is-not-installed

Fix type reinstallation. The error was connected with implicit addition of isDeleted and isArchived filters to db query

### DIFF
--- a/core/block/object/objectcreator/installer.go
+++ b/core/block/object/objectcreator/installer.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/anyproto/any-sync/commonspace/object/tree/treestorage"
 	"github.com/gogo/protobuf/types"
-	"github.com/samber/lo"
 	"go.uber.org/zap"
 
 	"github.com/anyproto/anytype-heart/core/block/editor/objecttype"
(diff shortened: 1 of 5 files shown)


core/block/object/objectcreator/installer.go
```diff
@@ -279,48 +269,32 @@ func (s *service) prepareDetailsForInstallingObject(
...
-				RelationKey: bundle.RelationKeySpaceId.String(),
-				Condition:   model.BlockContentDataviewFilter_Equal,
-				Value:       pbtypes.String(space.Id()),
-			},
-			{
-				RelationKey: bundle.RelationKeyIsDeleted.String(),
-				Condition:   model.BlockContentDataviewFilter_Equal,
-				Value:       pbtypes.Bool(true),
-			},
+func (s *service) queryDeletedObjects(space clientspace.Space, sourceObjectIDs []string) ([]database.Record, error) {
```
deff7, line 272: you need to test this, we don't usually use Or filter, so i'm not sure it will work properly

KirillSto: I have added a unit test on this

deff7 (approved):

PR #1463, opened 2024-08-09
GO-3870: HistoryGetVersions ignore limit param

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-3870/historygetversions-ignore-limit-param

### DIFF
--- a/core/history/history.go
+++ b/core/history/history.go
@@ -148,6 +148,9 @@ func (h *history) Versions(id domain.FullID, lastVersionId string, limit int) (r
 		if e != nil {
 			return nil, e
 		}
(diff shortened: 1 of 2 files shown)


core/history/history.go
```diff
@@ -148,6 +148,9 @@ func (h *history) Versions(id domain.FullID, lastVersionId string, limit int) (r
 		if e != nil {
 			return nil, e
 		}
+		if len(data) == 0 {
```
mcrakhman, line 151: How is it possible that data == 0? Because I guess we will always call func on first element

mcrakhman (approved):

PR #1468, opened 2024-08-12
Fix Not filter

deff7:


pkg/lib/database/filter.go
```diff
@@ -275,6 +275,12 @@ func negateFilter(filter query.Filter) query.Filter {
 		}
 		return negated
 	case query.Key:
+		if nested, ok := v.Filter.(query.Not); ok {
```
deff7, line 278: Not(Not(Filter)) -> Filter

PR #1472, opened 2024-08-13
GO-3907 bug we somehow broke an object in multiplayer at some point

deff7:


KirillSto (approved):

core/block/editor/state/state.go
```diff
@@ -329,18 +329,18 @@ func (s *State) PickParentOf(id string) (res simple.Block) {
...
-			for topParentId != rootId {
-				if nextId, ok := cache[topParentId]; ok {
-					topParentId = nextId
-				} else {
-					cacheFound = nil
-					break
-				}
-			}
-		}
+		// if cacheFound != nil {
```
KirillSto, line 332: If we just comment this code we will get an avalanche of **discrepancy in state parent search** log messages

PR #1483, opened 2024-08-14
GO-3923 Include images in links

KirillSto:
https://linear.app/anytype/issue/GO-3923/images-are-not-added-to-links

PR includes refactoring of DependentObjectIds func. Now this function is used to fill **links** detail

### DIFF
--- a/core/block/object/objectlink/dependent_objects.go
+++ b/core/block/object/objectlink/dependent_objects.go
@@ -94,6 +94,7 @@ func collectIdsFromBlocks(s *state.State, flags Flags) (ids []string) {
 			}
 		}
 
+		// if NoImages == false, then file block will be processed with FillSmartIds
 		if flags.NoImages {
 			if f := b.Model().GetFile(); f != nil {
 				if f.TargetObjectId != "" && f.Type != model.BlockContentFile_Image {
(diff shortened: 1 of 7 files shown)


AnastasiaShemyakinskaya (approved):

core/block/object/objectlink/dependent_objects.go
```diff
@@ -74,65 +68,133 @@ func DependentObjectIDs(s *state.State, converter KeyToIDConverter, blocks, deta
...
 			}
-			continue
 		}
 
-		if rel.Key == bundle.RelationKeyCreator.String() ||
-			rel.Key == bundle.RelationKeyLastModifiedBy.String() {
-			if creatorModifierWorkspace {
-				v := pbtypes.GetString(det, rel.Key)
-				ids = append(ids, v)
+		if flags.NoImages {
```
deff7, line 97: what if this flag is false?

KirillSto: If flag is false, then file block will be processed by FillSmartIds logic, because file block implements linkSource interface

deff7: I suggest you to leave a comment about this

KirillSto: Done!

deff7 (approved):

PR #1488, opened 2024-08-15
GO-3795 Add ObjectListModifyDetailValues RPC

KirillSto:
https://linear.app/anytype/issue/GO-3795/addremove-details-from-multiple-objects

New RPC request allows to set/add/remove detail values for multiple objects


util/slice/slice.go
```diff
@@ -237,3 +237,24 @@ func FilterCID(cids []string) []string {
...
 }
+
+// MergeUniqBy merges two slices with comparator. Resulting slice saves values' order and uniqueness
+// Input slices MUST contain only unique values
+func MergeUniqBy[T comparable](s1, s2 []T, equal func(v1, v2 T) bool) (result []T) {
+	result = make([]T, len(s1))
+	copy(result, s1)
+	for _, v2 := range s2 {
+		isUnique := true
+		for _, v1 := range s1 {
```
deff7, line 248: you could use Contains

util/pbtypes/pbtypes.go
```diff
@@ -582,3 +606,46 @@ func RelationIdToKey(id string) (string, error) {
 
 	return "", fmt.Errorf("incorrect id format")
 }
+
+// AddValue adds values to int lists and string lists
+func AddValue(s *types.Struct, key string, v *types.Value) {
```
deff7, line 611: I don't think these functions should be here, they are pretty specialized for only one task, so maybe move them close to place when they're called?

core/block/editor.go
```diff
@@ -639,6 +639,36 @@ func (s *Service) ModifyDetails(objectId string, modifier func(current *types.St
...
+	var anySucceed bool
+	for _, objectId := range req.ObjectIds {
+		err := s.ModifyDetails(objectId, func(current *types.Struct) (*types.Struct, error) {
+			for _, op := range req.Operations {
+				if op.Set != nil && pbtypes.ValueToInterface(op.Set) != nil {
+					// Set operation has higher priority than Add and Remove, because it modifies full value
+					current.Fields[op.RelationKey] = op.Set
+					continue
+				}
+				pbtypes.AddValue(current, op.RelationKey, op.Add)
```
deff7, line 652: maybe use switch / if between Set/Add/Remove operations? It will be clearer

util/pbtypes/pbtypes.go
KirillSto: I moved it to separate file core/block/details.go, so all block.Service operations on details are described here

core/block/editor.go
KirillSto: I did not get it :see_no_evil: 
Can you please suggest how we can do it via switch

deff7 (approved):

PR #1526, opened 2024-08-30
GO-4000: Fix options filter

deff7:


pkg/lib/database/filter.go
```diff
@@ -650,20 +659,48 @@ func (exIn FilterOptionsEqual) FilterObject(g *types.Struct) bool {
...
+
+	arr := v.GetArray(exIn.Key)
+	// Just fall back to precompiled filter
+	if len(arr) == 0 {
+		return exIn.valueFilter.Ok(v.Get(exIn.Key))
+	}
+
+	// Discard deleted options
+	optionList := exIn.arena.NewArray()
+	var i int
```
AnastasiaShemyakinskaya, line 673: It seems like it doesn't used 🧐

deff7: `i` is used to set array item

AnastasiaShemyakinskaya (approved):

PR #1531, opened 2024-09-02
GO-3946 | GO-3587 - Gallery cache

KirillSto:
https://linear.app/anytype/issue/GO-3946/add-cache-for-index-and-for-usecase
https://linear.app/anytype/issue/GO-3587/add-cache-on-gallerydownloadindex

- Introduce gallery.Service, where all logic with import of gallery usecases is accumulated
- Introduce IndexCache that saves latest index locally
- Introduce DashboardInjector that injects Homepage object during ooooold space migration
- Import built-in usecases using experience pipeline (it is possible due to widget and workspace being added to archives)
- Add **cache** parameter to all gallery-related RPCs
- Experience import uses clientCache in case built-in archive could not be downloaded from remote
- Slight changes on Usecase validator


core/gallery/cache.go
```diff
@@ -0,0 +1,231 @@
...
+		return nil, err
+	}
+
+	c.storage.save(index, newVersion)
+
+	return index, nil
+}
+
+func (c *cache) GetManifest(downloadLink string, timeoutInSeconds int) (info *model.ManifestInfo, err error) {
+	localIndex, err := c.storage.getIndex()
```
AnastasiaShemyakinskaya, line 99: May be extract some duplicate code from GetIndex and GetManifest? Looks like they have a lot of common code?

core/gallery/cache.go
```diff
@@ -0,0 +1,231 @@
...
+
+	index, newVersion, err := c.downloadGalleryIndex(timeoutInSeconds, version, true)
+	if err == nil {
+		c.storage.save(index, newVersion)
+
+		return getManifestByDownloadLink(index, downloadLink)
+	}
+
+	if errors.Is(err, ErrNotModified) {
+		manifest, err := getManifestByDownloadLink(localIndex, downloadLink)
```
AnastasiaShemyakinskaya, line 120: Can be just `return getManifestByDownloadLink(localIndex, downloadLink)`?

core/gallery/cache.go
```diff
@@ -0,0 +1,231 @@
...
+		if err != nil {
+			return nil, err
+		}
+		return manifest, nil
+	}
+
+	if localIndex != nil {
+		log.Warn("failed to download index from remote. Returning local index", zap.Error(err))
+		manifest, err := getManifestByDownloadLink(localIndex, downloadLink)
```
AnastasiaShemyakinskaya, line 129: The same

core/gallery/cache.go
```diff
@@ -0,0 +1,231 @@
...
+		return fmt.Errorf("manifest does not correspond scema: %w", err)
+	}
+
+	for _, urlToCheck := range append(info.Screenshots, info.DownloadLink) {
+		if !isInWhitelist(urlToCheck) {
+			return fmt.Errorf("URL '%s' provided in manifest is not in whitelist", urlToCheck)
+		}
+	}
+
+	info.Description = stripTags(info.Description)
```
AnastasiaShemyakinskaya, line 229: I think, it should be moved out of function, because it doesn't relate to validation, because according to name, this function is supposed to just validate manifest?🧐

core/gallery/cache.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+func (c *cache) downloadGalleryIndex(
+	timeoutInSeconds int, // timeout to wait for HTTP response
+	version string, // Etag of gallery index, that allows to fetch index faster
+	withManifestValidation bool, // a flag that indicates that every manifest should be validated
+) (response *pb.RpcGalleryDownloadIndexResponse, newVersion string, err error) {
+	raw, newVersion, err := getRawJson(c.indexURL, timeoutInSeconds, version)
+	if err != nil {
+		if errors.Is(err, ErrNotModified) {
```
AnastasiaShemyakinskaya, line 195: This check is not needed, because it's implemented in upper function afaiu? 🧐

util/dashboardinjector/injector.go
```diff
@@ -0,0 +1,88 @@
...
+		Origin: objectorigin.Usecase(),
+		IsSync: true,
+	}
+	res := b.importer.Import(context.Background(), importRequest)
+
+	if res.Err != nil {
+		return err
+	}
+
+	if rmErr := os.Remove(path); rmErr != nil {
```
AnastasiaShemyakinskaya, line 84: Should this logic also be executed in case of error?

core/gallery/service.go
```diff
@@ -0,0 +1,494 @@
...
+	start := time.Now()
+
+	info, found := ucCodeToInfo[useCase]
+	if !found {
+		return pb.RpcObjectImportUseCaseResponseError_BAD_INPUT,
+			fmt.Errorf("failed to import built-in usecase: invalid Use Case value: %v", useCase)
+	}
+
+	if cachePath == "" {
+		if ctx == nil {
```
AnastasiaShemyakinskaya, line 131: May be if `ImportBuiltInUseCase` should be called with context.Background() instead of nil to avoid such check?

core/gallery/service.go
```diff
@@ -0,0 +1,494 @@
...
+		return nil, fmt.Errorf("failed to add progress bar: %w", err)
+	}
+	progress.SetProgressMessage("downloading archive")
+	progress.SetTotal(100)
+	return progress, nil
+}
+
+func getArchiveReaderAndSize(url string) (reader io.ReadCloser, size int64, err error) {
+	client := http.Client{Timeout: 15 * time.Second}
+	// nolint: gosec
```
AnastasiaShemyakinskaya, line 459: Why do we need this comment? It seems like there is no security issues? 🧐

core/gallery/service.go
```diff
@@ -0,0 +1,494 @@
...
+	for _, manifest := range index.Experiences {
+		if manifest.DownloadLink == link {
+			return manifest, nil
+		}
+	}
+	return nil, fmt.Errorf("failed to find manifest for url: %s", link)
+}
+
+func generateMapOfDownloadLinksByNames(index *pb.RpcGalleryDownloadIndexResponse) map[string]string {
+	m := make(map[string]string)
```
AnastasiaShemyakinskaya, line 489: Preallocate

core/gallery/service.go
```diff
@@ -0,0 +1,494 @@
...
+	if cachePath == "" {
+		return nil, nil, fmt.Errorf("no cache path specified")
+	}
+	r, err := zip.OpenReader(cachePath)
+	if err != nil {
+		return nil, nil, err
+	}
+	defer r.Close()
+
+	archives = make(map[string][]byte)
```
AnastasiaShemyakinskaya, line 277: Allocate it near usage please

core/gallery/cache.go
KirillSto: It is actually needed, as in case requested resource is not actually modified we do not wrap it into ErrDownloadIndex, that is translated into RPC return code

core/gallery/service.go
KirillSto: Actally there is. Linter worried about referencing some mal resources :see_no_evil:

core/gallery/service.go
KirillSto: Done!

core/gallery/service.go
KirillSto: Done!

core/gallery/service.go
KirillSto: Yeah, I have added ctx to all calls, thank you!

core/gallery/cache.go
KirillSto: I have changed strip logic a lil bit. Now it accepts manifest and is not sticked to validation

core/gallery/cache.go
KirillSto: Yes, it hepled to remove 30 lines :smile:

core/block/import/pb/converter.go
```diff
@@ -163,7 +163,7 @@ func (p *Pb) handleImportPath(
 		needToImportWidgets = p.needToImportWidgets(profile.Address, pr.AccountId)
 		profileID = profile.ProfileId
 	}
-	return p.getSnapshotsFromProvidedFiles(pathCount, importSource, allErrors, path, profileID, needToImportWidgets, isMigration, importType)
+	return p.getSnapshotsFromProvidedFiles(pathCount, importSource, allErrors, path, profileID, needToImportWidgets, isMigration, isNewSpace, importType)
```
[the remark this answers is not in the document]
KirillSto: I have refactored whole converter.go a little bit. Actaully, we can save a lot of info inside Pb struct state, so there is no need to transfer it throghout method calls. Please check if I could miss smth

util/dashboardinjector/injector.go
AnastasiaShemyakinskaya: Then it should be in defer?

KirillSto: Yeah, thank you! I put it into defer and also fixed error return. On import failure I used to return wrong error :see_no_evil:

PR #1532, opened 2024-09-02
GO-3853 Fix creationDate for get_started

KirillSto:
https://linear.app/anytype/issue/GO-3853/newly-created-accounts-still-have-incorrect-dates-in-their-objects

- We should remove lastModifiedDate from objects of get_started usecase, as this timestamp is set as creationDate of all objects on new account creation
- Introduce complex filter _'valueFromRelation'_ that allows to filter out objects by comparing values of two different relations


pkg/lib/database/filter.go
```diff
@@ -845,3 +868,91 @@ func (i *FilterNestedNotIn) AnystoreFilter() query.Filter {
...
+type Anystore2ValuesComp struct {
+	Path1, Path2 []string
+	CompOp       query.CompOp
+	buf1, buf2   []byte
+}
+
+func (e *Anystore2ValuesComp) Ok(v *fastjson.Value) bool {
+	value1 := v.Get(e.Path1...)
+	value2 := v.Get(e.Path2...)
+	comp := bytes.Compare(encoding.AppendJSONValue(e.buf1[:0], value1), encoding.AppendJSONValue(e.buf2[:0], value2))
```
deff7, line 916: e.buf is always nil, you should set it to a new value

```
e.buf = encoding.AppendJSONValue(e.buf1[:0], value1)
```

util/pbtypes/pbtypes.go
```diff
@@ -276,6 +276,16 @@ func GetStringListValue(v *types.Value) []string {
 	return nil
 }
 
+func GetStructValue(v *types.Value) *types.Struct {
```
deff7, line 279: `v.GetStructValue()`

pkg/lib/database/filter.go
KirillSto: Thank you! Added init of each buffer before comparison

pkg/lib/database/filter_test.go
```diff
@@ -26,6 +26,15 @@ func assertFilter(t *testing.T, f Filter, obj *types.Struct, expected bool) {
 	assert.Equal(t, expected, anystoreFilter.Ok(val))
 }
 
+func assertFilterNoParseValidation(t *testing.T, f Filter, obj *types.Struct, expected bool) {
```
deff7, line 29: change original function please, this parsing validation is not necessary

deff7 (approved):

pkg/lib/database/filter_test.go
KirillSto: Yeah, removed parse validation from filter assertion

PR #1552, opened 2024-09-06
GO-3964 ListRelationsByValue RPC

KirillSto:
https://linear.app/anytype/issue/GO-3964/provide-the-list-of-relations-on-the-layout-with-date-as-a-value

- Introducing new RPC for listing relations that hold the value provided in request parameter. Response includes list of relation keys in alphabetical order and number of objects that have such value under particular relation
- New method **objectstore.QueryIterate** allows to query records from object store and process its details using custom functor
- Big bench of block.Service methods are drawn into new detailservice.Service
- Unit tests on new service are added


pkg/lib/localstore/objectstore/queries.go
```diff
@@ -504,3 +504,41 @@ func (s *dsObjectStore) QueryByIDAndSubscribeForChanges(ids []string, sub databa
 	s.addSubscriptionIfNotExists(sub)
 	return
 }
+
+func (s *dsObjectStore) QueryAndProcess(q database.Query, proc func(details *types.Struct)) error {
```
deff7, line 508: QueryIterate?

core/block/relation.go
```diff
@@ -56,3 +64,75 @@ func (s *Service) ObjectTypeRemoveRelations(ctx context.Context, objectTypeId st
 		return b.Apply(st)
 	})
 }
+
+func (s *Service) ListRelationsWithValue(spaceId string, value *types.Value) (keys []string, counters []int64, err error) {
```
deff7, line 68: is it time for relationService?

pkg/lib/localstore/objectstore/queries.go
KirillSto: Renamed!

core/block/relation.go
KirillSto: Yes :relieved:

pkg/lib/localstore/addr/addr.go
```diff
@@ -48,3 +48,10 @@ func ExtractVirtualSourceType(id string) (model.SmartBlockType, error) {
 func TimeToID(t time.Time) string {
 	return DatePrefix + t.Format("2006-01-02")
 }
+
+func DateIDToDayStart(id string) (time.Time, error) {
```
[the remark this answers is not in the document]
KirillSto: Done!

core/block/detailservice/set_details.go
```diff
@@ -0,0 +1,248 @@
...
+		if value {
+			return coll.AddObject(objectId)
+		} else {
+			return coll.RemoveObject(objectId)
+		}
+	})
+}
+
+func (s *service) partitionObjectIDsBySpaceID(objectIDs []string) (map[string][]string, error) {
+	res := map[string][]string{}
```
AnastasiaShemyakinskaya, line 193: May be it can be preallocated?

core/block/detailservice/set_details.go
```diff
@@ -0,0 +1,248 @@
...
+	return res, nil
+}
+
+func (s *service) setIsArchivedForObjects(spaceID string, objectIDs []string, isArchived bool) error {
+	spc, err := s.spaceService.Get(context.Background(), spaceID)
+	if err != nil {
+		return fmt.Errorf("get space: %w", err)
+	}
+	return cache.Do(s, spc.DerivedIDs().Archive, func(b smartblock.SmartBlock) error {
+		archive, ok := b.(collection.Collection)
```
AnastasiaShemyakinskaya, line 210: Maybe extract to separate function?

core/block/detailservice/set_details.go
```diff
@@ -0,0 +1,248 @@
...
+		if !ok {
+			return fmt.Errorf("unexpected archive block type: %T", b)
+		}
+
+		var multiErr multierror.Error
```
AnastasiaShemyakinskaya, line 215: Can be combined
var (
       multiErr multierror.Error.
       anySucceed bool
)

core/block/detailservice/set_details.go
```diff
@@ -0,0 +1,248 @@
...
+				multiErr.Errors = append(multiErr.Errors, err)
+				continue
+			}
+			anySucceed = true
+		}
+
+		if err := multiErr.ErrorOrNil(); err != nil {
+			log.Warn("failed to archive", zap.Error(err))
+		}
+		if anySucceed {
```
AnastasiaShemyakinskaya, line 243: It is seems like unnecessary check, because multiErr.ErrorOrNil() is nil for sure. So can we just return nil instead of lines 243-246?

core/block/detailservice/set_details.go
```diff
@@ -0,0 +1,248 @@
...
+			log.Error("failed to favorite object", zap.String("objectId", id), zap.Error(err))
+			resultError = errors.Join(resultError, err)
+		} else {
+			anySucceed = true
+		}
+	}
+	if resultError != nil {
+		log.Warn("failed to set objects as favorite", zap.Error(resultError))
+	}
+	if anySucceed {
```
AnastasiaShemyakinskaya, line 137: The same comment as for isArchive function

core/block/detailservice/relations.go
```diff
@@ -0,0 +1,146 @@
...
+func (s *service) ListRelationsWithValue(spaceId string, value *types.Value) (keys []string, counters []int64, err error) {
+	countersByKeys := make(map[string]int64)
+	detailHandlesValue := generateFilter(value)
+
+	err = s.store.QueryIterate(database.Query{Filters: []*model.BlockContentDataviewFilter{{
+		RelationKey: bundle.RelationKeySpaceId.String(),
+		Condition:   model.BlockContentDataviewFilter_Equal,
+		Value:       pbtypes.String(spaceId),
+	}}}, func(details *types.Struct) {
+		for key, v := range details.Fields {
```
AnastasiaShemyakinskaya, line 78: Can we rename v to value or something? Because it looks like not consistent naming 🧐

pkg/lib/localstore/objectstore/queries.go
```diff
@@ -504,3 +504,41 @@ func (s *dsObjectStore) QueryByIDAndSubscribeForChanges(ids []string, sub databa
...
+		if err != nil {
+			return errors.Join(fmt.Errorf("json to proto: %w", err), iter.Close())
+		}
+		proc(details)
+	}
+	err = iter.Err()
+	if err != nil {
+		return errors.Join(fmt.Errorf("iterate: %w", err), iter.Close())
+	}
+	err = iter.Close()
```
cheggaaa, line 539: Call "Close" in the defer, because it may cause a connection leak
```
iter, err := query.Iter(s.componentCtx)
	if err != nil {
		return fmt.Errorf("find: %w", err)
	}
	defer func() {
	    err = iter.Close()
	}()
```

cheggaaa (changes requested):

core/block/detailservice/set_details.go
KirillSto: Thank you!

core/block/detailservice/set_details.go
KirillSto: The point here is to return nil, if at least one object was processed without error.
If some objects were processed with error, than we should log it.
If all objects were processed with error, than we should return error.
So there is no issue here

core/block/detailservice/set_details.go
KirillSto: Same situation is here

core/block/detailservice/relations.go
KirillSto: Yeah, fixed

pkg/lib/localstore/objectstore/queries.go
KirillSto: Done!

core/block/detailservice/set_details.go
KirillSto: Done!

core/block/detailservice/set_details.go
KirillSto: I have splitted this function onto two separate ones

cheggaaa (approved):

PR #1561, opened 2024-09-09
GO-3924: Fix event message ordering for state change events

deff7:


core/block/editor/state/state.go
```diff
@@ -686,6 +686,8 @@ func (s *State) apply(fast, one, withLayouts bool) (msgs []simple.EventMessage,
 	}
 
 	msgs = s.processTrailingDuplicatedEvents(msgs)
+
+	sortEventMessages(msgs)
```
KirillSto, line 690: It seems we should insert sorting before trailing dups analysis. Or we don't need this analysis at all? :grinning:

fat-fellow (approved):

KirillSto (approved):

PR #1577, opened 2024-09-13
GO-4025 Modify default rel lists for dataview

KirillSto:
https://linear.app/anytype/issue/GO-4025/prepare-default-list-of-relations-for-new-set-views

- default lists of relations for new dataview views are changed a lil bit
- logic on relation list calculation is drawn to _editor.template_
- block.SetSource is removed, as its logic duplicates SetDataviewSource


core/block/editor/template/collection.go
```diff
@@ -9,68 +11,137 @@ import (
...
 					Relations: viewRelations,
 				},
 			},
 		},
 	}
-	return blockContent
 }
 
-var DefaultDataviewRelations = make([]domain.RelationKey, 0, len(bundle.RequiredInternalRelations))
+func generateRelationLists(
```
AnastasiaShemyakinskaya, line 89: Personally, I would rename this function, because it doesn't generate anything 🧐 Maybe `retrieveViewRelations`?

core/block/editor/template/collection.go
```diff
@@ -9,68 +11,137 @@ import (
...
+)
+
+func MakeDataviewContent(isCollection bool, ot *model.ObjectType, relLinks []*model.RelationLink) *model.BlockContentOfDataview {
+	var (
+		defaultRelations = defaultCollectionRelations
+		visibleRelations = defaultVisibleRelations
+		sorts            = DefaultLastModifiedDateSort()
+	)
+
+	if isCollection {
```
AnastasiaShemyakinskaya, line 56: L56-66 - may be extract to separate function? Seems like it make logic of `MakeDataviewContent` complicated

PR #1609, opened 2024-09-24
GO-4134 Fix linux build, musl: fix tantivy seg violation crash

zarkone:


zarkone:
maybe we should increase this to 2-4 mb, as in libc, should be safe..?

fat-fellow (changes requested):
Waiting for hashes updates

fat-fellow (approved):

PR #1619, opened 2024-09-27
GO-4147 divide any store indexes per space

deff7:


pkg/lib/localstore/objectstore/spaceindex/queries.go
```diff
@@ -585,3 +585,24 @@ func (s *dsObjectStore) QueryIterate(q database.Query, proc func(details *types.
...
 	return
 }
+
+func (s *dsObjectStore) ListIds() ([]string, error) {
+	var ids []string
+	iter, err := s.objects.Find(nil).Iter(s.componentCtx)
+	if err != nil {
+		return nil, fmt.Errorf("find all: %w", err)
+	}
+	for iter.Next() {
```
cheggaaa, line 595: call iter.Close in defer because in case of error, iter will not be closed and it will cause connection leak

mcrakhman (approved):

PR #1624, opened 2024-10-01
GO-4162 corrupted links in markdown import

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-4162/corrupted-links-in-markdown-import

The problem is that we don't handle cases, when block have more than one markdown besides links. For example link can be also bold or italic.

To fix it 
* I check if text block have more than one markdown 
* Then I go through all marks in text block and change the type of mark to page mention in case it is link to existing page or file.

### DIFF
--- a/core/block/import/markdown/blockconverter_test.go
+++ b/core/block/import/markdown/blockconverter_test.go
@@ -89,3 +89,155 @@ func Test_processFiles(t *testing.T) {
 		assert.Len(t, fileBlocks, 0)
 	})
 }
+
+func Test_processTextBlock(t *testing.T) {
+	t.Run("block with single markdown - file doesn't exist", func(t *testing.T) {
+		// given
+		mc := mdConverter{}
+		block := getTestTxtBlock("file.md")
+
+		// when
+		mc.processTextBlock(block, map[string]*FileInfo{})
+
+		// then
+		assert.NotNil(t, block.GetBookmark())
+	})
+	t.Run("block with single markdown - file exists", func(t *testing.T) {
+		mc := mdConverter{}
+		block := getTestTxtBlock("file.md")
+
+		// when
+		mc.processTextBlock(block, map[string]*FileInfo{"file.md": {PageID: "id"}})
+
+		// then
+		assert.NotNil(t, block.GetLink())
+		assert.Equal(t, "file.md", block.GetLink().GetTargetBlockId())
+	})
+	t.Run("block with single markdown - file exists, but not md or csv", func(t *testing.T) {
+		mc := mdConverter{}
+		block := getTestTxtBlock("file.txt")
+
+		// when
+		mc.processTextBlock(block, map[string]*FileInfo{"file.txt": {}})
+
+		// then
+		assert.NotNil(t, block.GetFile())
+		assert.Equal(t, "file.txt", block.GetFile().GetName())
+	})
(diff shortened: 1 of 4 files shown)


core/block/import/markdown/blockconverter_test.go
```diff
@@ -89,3 +89,155 @@ func Test_processFiles(t *testing.T) {
...
+
+		// then
+		// then
+		assert.NotNil(t, block.GetText())
+		assert.Len(t, block.GetText().GetMarks().GetMarks(), 3)
+		assert.Equal(t, "file.csv", block.GetText().GetMarks().GetMarks()[1].Param)
+		assert.Equal(t, model.BlockContentTextMark_Object, block.GetText().GetMarks().GetMarks()[1].Type)
+	})
+}
+
```
fat-fellow, line 186: сonsider use @deff7's  testutil.BuildStateFromAST

AnastasiaShemyakinskaya: I don't see advantages in using it, because we need just to manipulate text block

core/block/import/markdown/blockconverter_test.go
```diff
@@ -89,3 +89,155 @@ func Test_processFiles(t *testing.T) {
 		assert.Len(t, fileBlocks, 0)
 	})
 }
+
+func Test_processTextBlock(t *testing.T) {
```
deff7, line 93: I suggest using real markdown test data here

testing small internal functions leads to fragile tests and don't cover the whole process of import

AnastasiaShemyakinskaya: Fixed

core/block/import/markdown/import_test.go
```diff
@@ -75,6 +76,89 @@ func TestMarkdown_GetSnapshots(t *testing.T) {
...
+				found = true
+				assert.Len(t, snapshot.Snapshot.Data.Blocks, 14)
+				assertLinkBlocks(t, snapshot)
+			}
+		}
+		assert.True(t, found)
+	})
+}
+
+func assertLinkBlocks(t *testing.T, snapshot *common.Snapshot) {
```
deff7, line 111: please, assert whole blocks 

you can use `blockbuilder` to compare two state trees, it ignores blockIds and compares only content and "shape" of document

deff7 (changes requested):

core/block/import/markdown/import_test.go
AnastasiaShemyakinskaya: Added

core/block/import/markdown/import_test.go
```diff
@@ -75,6 +76,146 @@ func TestMarkdown_GetSnapshots(t *testing.T) {
...
+		assert.Nil(t, err)
+		assert.NotNil(t, sn)
+		assert.Len(t, sn.Snapshots, 4)
+
+		fileNameToObjectId := make(map[string]string, len(sn.Snapshots))
+		for _, snapshot := range sn.Snapshots {
+			fileNameToObjectId[snapshot.FileName] = snapshot.Id
+		}
+		var found bool
+		expectedPath := "testdata" + string(filepath.Separator) + "links.md"
```
deff7, line 105: use `filepath.Join` everywhere

AnastasiaShemyakinskaya: Fixed

deff7 (approved):

PR #1629, opened 2024-10-02
GO-4168 Rollback SetSource

KirillSto:
https://linear.app/anytype/issue/GO-4168/the-selected-query-is-not-deleted-in-the-set-when-clicking-on-the

- We should rollback previous behaviour of ObjectSetSource, because BlockDataviewSetSource does completely other task
- Introduce dataviewservice.Service with all gataview-related methods drawn from block.Service
- Remove BlockDataviewRelationListAvailable from protocol


core/block/detailservice/set_details.go
```diff
@@ -57,6 +59,33 @@ func (s *service) SetWorkspaceDashboardId(ctx session.Context, workspaceId strin
 	return id, err
 }
 
+func (s *service) SetSource(ctx session.Context, objectId string, source []string) error {
+	return cache.Do(s, objectId, func(sb smartblock.SmartBlock) error {
+		st := sb.NewStateCtx(ctx)
+
+		// nolint:errcheck
+		_ = st.Iterate(func(b simple.Block) (isContinue bool) {
```
deff7, line 67: encapsulate this in `simple/dataview` block

deff7:
```
if dv := b.Model().GetDataview(); dv != nil {
				for _, view := range dv.Views {
					view.DefaultTemplateId = ""
					view.DefaultObjectTypeId = ""
				}
				st.Set(b)
				return false
			}
```

KirillSto: I have implemented separate method in editor.Dataview for setting source in sets

core/block/editor/dataview/dataview.go
```diff
@@ -112,6 +112,31 @@ func (d *sdataview) SetSource(ctx session.Context, blockId string, source []stri
...
+	if err != nil {
+		return err
+	}
+
+	for _, view := range dv.ListViews() {
+		// TODO: GO-4189 Need to review relation lists modification in each view on source change
+		view.DefaultTemplateId = ""
+		view.DefaultObjectTypeId = ""
+		if err = dv.SetView(view.Id, *view); err != nil {
+			log.With("objectId", d.Id()).Errorf("failed to update view '%s' of set: %v", view.Id, err)
```
deff7, line 127: maybe return with error?

btw use log fields for anything except error

KirillSto: Done!

deff7 (approved):

PR #1637, opened 2024-10-04
GO-4192 app start improvements

requilence:
- do not wait for system relations/types load/install if missing
- try to load in background, if failed (with 30 sec timeout) then create(derive) the new one 
- if somebody will try to open this object directly - it will hang for the cache loader

Also, current bundled objects install approach work bad during reindex. 
- We have all the bundled object in spacestore, But we don't have them in objectstore
- We check objectstore and then generate payload and try to install of them for all spaces (>1k objects if you have 10 spaces)
- Then we receive tree already exists error for each of them
This PR  improves this by checking StoredIds along with it


requilence:
close in favor of #1672

PR #1651, opened 2024-10-07
GO-4202 make the ft reindex logic atomic

requilence:
Add ids to the queue in one tx

### DIFF
--- a/pkg/lib/localstore/objectstore/indexer_store.go
+++ b/pkg/lib/localstore/objectstore/indexer_store.go
@@ -8,17 +8,26 @@ import (
 	"github.com/anyproto/anytype-heart/pkg/lib/pb/model"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/lib/localstore/objectstore/indexer_store.go
```diff
@@ -8,17 +8,26 @@ import (
...
 		s.arenaPool.Put(arena)
 	}()
-	obj := arena.NewObject()
-	obj.Set("id", arena.NewString(id))
 
-	_, err := s.fulltextQueue.UpsertOne(ctx, obj)
-	return err
+	obj := arena.NewObject()
+	for _, id := range ids {
+		obj.Set("id", arena.NewString(id))
```
deff7, line 24: I'm not sure if it's ok to reuse object

deff7 (approved):

PR #1690, opened 2024-10-15
GO-4171: Fix potential race in subscription cache

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-4171/fix-potential-race-in-subscription-cache


core/subscription/cache.go
```diff
@@ -86,13 +88,18 @@ func (e *entry) Get(key string) *types.Value {
 
 type cache struct {
 	entries map[string]*entry
+	sync.Mutex
```
fat-fellow, line 91: should we use the rwmutex and use readlock for get?

deff7: you could just lock the whole spaceSubscription in `collectionObserver.fetchEntries`, the 99% of time the cache is accessed in one goroutine, and it's accessed a lot, so constant locking-unlocking will add performance overhead 

btw use `feature/chat` as target branch

deff7 (approved):

PR #1692, opened 2024-10-15
GO-4282 reindex invalid participant permissions

requilence:
make `FetchRelationByKeys` works the same as `FetchRelationByKey` - to return bundled objects directly from map.
This was called via `ModifyParticipantAclState` and in case of empty objectstore (new account, reindex), failed to add the relationLinks.

RelationLinks will be removed next release, because we will store relation list in the type.

### DIFF
--- a/core/block/editor/basic/details.go
+++ b/core/block/editor/basic/details.go
@@ -321,10 +321,13 @@ func (bs *basic) addRelationLink(st *state.State, relationKey string) error {
 	return nil
 }
 
+// addRelationLinks is deprecated and will be removed in release 7
 func (bs *basic) addRelationLinks(st *state.State, relationKeys ...string) error {
 	if len(relationKeys) == 0 {
 		return nil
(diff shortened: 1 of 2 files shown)


core/block/editor/basic/details.go
```diff
@@ -321,10 +321,13 @@ func (bs *basic) addRelationLink(st *state.State, relationKey string) error {
 	return nil
 }
 
+// addRelationLinks is deprecated and will be removed in release 7
 func (bs *basic) addRelationLinks(st *state.State, relationKeys ...string) error {
 	if len(relationKeys) == 0 {
 		return nil
 	}
+	// this code depends on the objectstore being indexed, but can be run on start with empty account
+	// todo: remove this code in release because we will no longer need relationLinks
```
fat-fellow, line 330: Create and link a task please

mcrakhman (approved):

PR #1703, opened 2024-10-17
Go 4256 read exported ipld tree with assets and render

zarkone:
implements initial support for publishing multiple files, implemented via ufs. For now, I'm just sending dummy assets (for testing purposes) and next step would be to send real web page assets, such as images. Same goes for renderer on gateway, -- it doesn't support any assets at the moment.

current structure of published ufs dir:
```
- keys.json <- encrypted with main key, has keys for all the other files
- index.json <- renderer input 
- asset1
- asset2....
```

Uploading implemented via same mechanism which we use for invites but for the future I'm thinking about how to implement it on top of  `UploadFile`. It will also allow us to get rid of custom encryption scheme (`keys.json`) and reuse standard anytype-heart encryption mechanics for file objects.  So we need a dedicated file type for publishing, so it requires some design according to our current requirements.

### DIFF
--- a/core/publish.go
+++ b/core/publish.go
@@ -11,7 +11,6 @@ func (mw *Middleware) ObjectPublish(ctx context.Context, req *pb.RpcObjectPublis
 	publishService := getService[publish.Service](mw)
 
 	res, err := publishService.Publish(ctx, req.SpaceId, req.ObjectId)
-
 	code := mapErrorCode(err,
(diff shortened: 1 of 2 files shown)


zarkone:
linter complains about too long function but I don't think it makes sense to refactor it now: future PRs will have different implementation of this, as I described in PR description

PR #1713, opened 2024-10-21
GO-4310 anystore objectstore intergrity check

requilence:


pkg/lib/localstore/objectstore/helper/helper.go
```diff
@@ -0,0 +1,71 @@
+package helper
```
deff7, line 1: don't use Helper package name, it's too general

deff7: I think you could use `datastore` package for this purpose

pkg/lib/localstore/objectstore/anystorehelper/helper.go
```diff
@@ -0,0 +1,85 @@
...
+	} else {
+		// todo: add pid check after a full badger deprecation
+		_, _ = lockFile.Write([]byte(fmt.Sprintf("%d", os.Getpid())))
+		_ = lockFile.Close()
+	}
+
+	store, err = anystore.Open(ctx, path, config)
+	if err != nil {
+		code := sqlite.ErrCode(err)
+		l := log.With("err", err).With("code", code)
```
deff7, line 44: field should be named "error"

deff7 (approved):

PR #1744, opened 2024-10-28
GO-4244 Restrict chat duplication

KirillSto:
https://linear.app/anytype/issue/GO-4244/restricted-duplication-of-objects-in-the-menu

Duplication should be restricted for Chat object

### DIFF
--- a/core/block/restriction/object.go
+++ b/core/block/restriction/object.go
@@ -24,7 +24,7 @@ var (
 		model.Restrictions_Template,
 		model.Restrictions_Duplicate,
 	}
-	objFileRestrictions = ObjectRestrictions{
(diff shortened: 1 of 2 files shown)


core/block/restriction/object.go
```diff
@@ -24,7 +24,7 @@ var (
 		model.Restrictions_Template,
 		model.Restrictions_Duplicate,
 	}
-	objFileRestrictions = ObjectRestrictions{
+	objRestrictEditDup = ObjectRestrictions{
```
deff7, line 27: maybe EditAndDuplicate?

deff7 (approved):

PR #1769, opened 2024-11-01
GO-4408 Unify Date object creation

KirillSto:
https://linear.app/anytype/issue/GO-4408/dates-created-via-objectcreate-are-normal-objects
https://linear.app/anytype/issue/GO-4219/support-the-date-format
https://linear.app/anytype/issue/GO-4447/when-opening-the-date-object-another-space-opens

- Make single place in code where date object is created
- On ObjectCreate with ot-date we should not create real object, but just return its details based on date's name in details
- Introduce timestamp relation, that handles Unix format timestamp of the day's midnight
- Exclude dataview block insertion on date object openning (maybe some other block would be presented)
- Fix a bug with incorrect spaceId insertion


util/date/util.go
```diff
@@ -0,0 +1,33 @@
+package date
```
deff7, line 1: maybe rename package to `dateutil`?

deff7: and add simple tests

KirillSto: I will add tests on https://linear.app/anytype/issue/GO-4451/support-seconds-precision-date-object-id as date Id format will be changed

deff7 (approved):

PR #1783, opened 2024-11-05
GO-4360 cannot filter by action done in 0430

AnastasiaShemyakinskaya:
https://linear.app/anytype/issue/GO-4360/cannot-filter-by-action-done-in-0430

### DIFF
--- a/space/internal/components/migration/readonlyfixer/relationsfixer.go
+++ b/space/internal/components/migration/readonlyfixer/relationsfixer.go
@@ -12,7 +12,6 @@ import (
 	"github.com/anyproto/anytype-heart/core/session"
 	"github.com/anyproto/anytype-heart/pkg/lib/bundle"
 	"github.com/anyproto/anytype-heart/pkg/lib/database"
-	"github.com/anyproto/anytype-heart/pkg/lib/localstore/objectstore"
 	"github.com/anyproto/anytype-heart/pkg/lib/pb/model"
 	"github.com/anyproto/anytype-heart/space/internal/components/dependencies"
 	"github.com/anyproto/anytype-heart/util/pbtypes"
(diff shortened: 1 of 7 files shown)


space/internal/components/migration/runner.go
```diff
@@ -26,7 +25,7 @@ const (
 var log = logger.NewNamed(CName)
 
 type Migration interface {
-	Run(context.Context, logger.CtxLogger, dependencies.QueryableStore, dependencies.QueryableStore, dependencies.SpaceWithCtx) (toMigrate, migrated int, err error)
+	Run(context.Context, logger.CtxLogger, objectstore.ObjectStore, dependencies.SpaceWithCtx) (toMigrate, migrated int, err error)
```
KirillSto, line 28: We need to introduce new interface in `space/internal/components/dependencies`.
It could be called SpaceIndexGetter and implement method
`SpaceIndex(spaceId string) QuaryableStore`

space/internal/components/migration/readonlyfixer/relationsfixer.go
```diff
@@ -31,10 +32,12 @@ func (Migration) Name() string {
...
 }
 
-func (Migration) Run(ctx context.Context, log logger.CtxLogger, store, _ dependencies.QueryableStore, space dependencies.SpaceWithCtx) (toMigrate, migrated int, err error) {
+func (Migration) Run(ctx context.Context, log logger.CtxLogger, store objectstore.ObjectStore, space dependencies.SpaceWithCtx) (toMigrate, migrated int, err error) {
 	spaceId := space.Id()
 
-	relations, err := listReadonlyTagAndStatusRelations(store, spaceId)
+	spaceStore := store.SpaceIndex(spaceId)
+
+	relations, err := listReadonlyTagAndStatusRelations(spaceStore, spaceId)
```
KirillSto, line 40: It seems `listReadonlyTagAndStatusRelations` does not need spaceId parameter anymore

space/internal/components/migration/runner.go
KirillSto: This dependency is needed to guarantee migration will not use data changing methods of ObjectStore

KirillSto (changes requested):

PR #1788, opened 2024-11-06
GO-4450 Details structs

deff7:


core/domain/details.go
```diff
@@ -0,0 +1,173 @@
...
+			return nil
+		} else {
+			return fmt.Errorf("unsupported array type %s", firstVal.Type())
+		}
+	}
+	d.Set(key, Null())
+	return nil
+}
+
+// StructDiff returns pb struct which contains:
```
fat-fellow, line 126: You can clarify by linking to the concept of [symmetric difference](https://en.wikipedia.org/wiki/Symmetric_difference) for better understanding.

fat-fellow (approved):

PR #1790, opened 2024-11-06
GO-3913 separate getting identities into multiple calls to identity

AnastasiaShemyakinskaya:
Description: https://linear.app/anytype/issue/GO-3913/separate-getting-identities-into-multiple-calls-to-identity-repo


core/identity/identity.go
```diff
@@ -280,6 +278,32 @@ func (s *service) getIdentitiesDataFromRepo(ctx context.Context, identities []st
...
+			repoIdentities      []*identityrepoproto.DataWithIdentity
+			err                 error
+			requestedIdentities []string
+		)
+		if j+identityBatch < len(identities) {
+			requestedIdentities = identities[j : j+identityBatch]
+		} else {
+			requestedIdentities = identities[j:]
+		}
+		repoIdentities, err = s.identityRepoClient.IdentityRepoGet(ctx, requestedIdentities, []string{identityRepoDataKind})
```
deff7, line 295: maybe send requests in parallel?

core/identity/identity.go
```diff
@@ -280,6 +278,32 @@ func (s *service) getIdentitiesDataFromRepo(ctx context.Context, identities []st
...
+		if err != nil {
+			log.Error("error getting identity repo data", zap.Error(err))
+			failedIdentities = append(failedIdentities, requestedIdentities...)
+		} else {
+			resultIdentities = append(resultIdentities, repoIdentities...)
+		}
+		j += identityBatch
```
deff7, line 302: maybe move it to the loop expression?

```
for j := 0; j < len(identities); j+= identityBatch {
```

core/identity/identity.go
```diff
@@ -280,6 +278,32 @@ func (s *service) getIdentitiesDataFromRepo(ctx context.Context, identities []st
...
+			requestedIdentities []string
+		)
+		if j+identityBatch < len(identities) {
+			requestedIdentities = identities[j : j+identityBatch]
+		} else {
+			requestedIdentities = identities[j:]
+		}
+		repoIdentities, err = s.identityRepoClient.IdentityRepoGet(ctx, requestedIdentities, []string{identityRepoDataKind})
+		if err != nil {
+			log.Error("error getting identity repo data", zap.Error(err))
```
deff7, line 297: I'm concerned about the number of log messages. Is it possible that we will log unnecessary errors?

core/identity/identity.go
```diff
@@ -280,6 +278,32 @@ func (s *service) getIdentitiesDataFromRepo(ctx context.Context, identities []st
 	return res, nil
 }
 
+func (s *service) batchIdentities(ctx context.Context, identities []string) ([]*identityrepoproto.DataWithIdentity, []string) {
+	resultIdentities := make([]*identityrepoproto.DataWithIdentity, 0, len(identities))
```
mcrakhman, line 282: Just summing all the comments, there should be one abstract function, which splits the []T to several batches of size <= N (take N as 200 pls), this function should be in package `slice` or whatever and tested

and then call in parallel some named func for every batch which returns results using waitgroup, gather everything and return.

This func should either get stuff from remote or if nothing in remote get stuff from local.

I would do it like this

core/identity/identity.go
AnastasiaShemyakinskaya: May be we can collect such errors and then log it in one log? 🧐

mcrakhman (approved):

mcrakhman (approved):

mcrakhman (approved):

PR #1803, opened 2024-11-08
GO-4451 Long Date object id

KirillSto:
https://linear.app/anytype/issue/GO-4451/support-seconds-precision-date-object-id

- Date object id now carries seconds precision date information
- RelationListWithValue searches for relations that handle values of timestamps and date objects in range of the queried day
- Filter HasPrefix is introduced
- If [In] filter goes along with date object id value, it is substituted with HasPrefix filter

### DIFF
--- a/pkg/lib/database/filter.go
+++ b/pkg/lib/database/filter.go
@@ -468,11 +468,14 @@ func (p FilterHasPrefix) FilterObject(s *types.Struct) bool {
 }
 
 func (p FilterHasPrefix) AnystoreFilter() query.Filter {
-	f, err := query.ParseCondition(`{"` + p.Key + `":{"$regex": "^` + p.Prefix + `"}}`)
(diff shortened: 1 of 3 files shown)


pkg/lib/database/filter.go
```diff
@@ -435,6 +444,37 @@ func (e FilterEq) filterObject(v *types.Value) bool {
...
+	for _, v := range list.Values {
+		if strings.HasPrefix(v.GetStringValue(), p.Prefix) {
+			return true
+		}
+	}
+	return false
+}
+
+func (p FilterHasPrefix) AnystoreFilter() query.Filter {
+	f, err := query.ParseCondition(`{"` + p.Key + `":{"$regex": "^` + p.Prefix + `"}}`)
```
deff7, line 471: Is it possible to use constructors from anystore like it is done in other places?

KirillSto: Thank you, fixed!

deff7 (approved):

PR #1827, opened 2024-11-18
Fix fileObjects handling in usecase validator

KirillSto:
- FileObjects were not handled, because of dumb file name prefix check
- Remove relation options that are sticked to absent relations
- Delete mentions detail from objects, as it is derived

### DIFF
--- a/cmd/usecasevalidator/main.go
+++ b/cmd/usecasevalidator/main.go
@@ -71,7 +71,12 @@ func (f cliFlags) isUpdateNeeded() bool {
 	return f.analytics || f.creator || f.removeRelations || f.exclude || f.rules != ""
 }
 
-const anytypeProfileFilename = addr.AnytypeProfileId + ".pb"
+const (
+	anytypeProfileFilename = addr.AnytypeProfileId + ".pb"
+
+	filesPrefix       = "files"
+	fileObjectsPrefix = "filesObjects"
+)
 
 var (
 	errIncorrectFileFound = fmt.Errorf("incorrect protobuf file was found")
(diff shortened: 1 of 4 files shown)


cmd/usecasevalidator/main.go
```diff
@@ -208,7 +208,7 @@ func collectUseCaseInfo(files []*zip.File, fileName string) (info *useCaseInfo,
 			continue
 		}
 
-		if strings.HasPrefix(f.Name, "files") || f.FileInfo().IsDir() {
+		if (strings.HasPrefix(f.Name, "files") && !strings.HasPrefix(f.Name, "filesObjects")) || f.FileInfo().IsDir() {
```
AnastasiaShemyakinskaya, line 211: Use constants please

KirillSto: Done!

cmd/usecasevalidator/main.go
```diff
@@ -71,7 +71,12 @@ func (f cliFlags) isUpdateNeeded() bool {
 	return f.analytics || f.creator || f.removeRelations || f.exclude || f.rules != ""
 }
 
-const anytypeProfileFilename = addr.AnytypeProfileId + ".pb"
+const (
+	anytypeProfileFilename = addr.AnytypeProfileId + ".pb"
+
+	filesPrefix       = "files"
```
AnastasiaShemyakinskaya, line 77: We have the same constants in export package, is is possible to reuse it somehow?

KirillSto: Oh yeah, did not think of them first...
Fixed!

PR #1833, opened 2024-11-18
GO-4406: creation date of object type is 0

AnastasiaShemyakinskaya:
Set original created date of derived objects in case it is provided in details in function `CreateObject`


core/block/object/objectcreator/object_type.go
```diff
@@ -43,6 +43,9 @@ func (s *service) createObjectType(ctx context.Context, space clientspace.Space,
 
 	createState := state.NewDocWithUniqueKey("", nil, uniqueKey).(*state.State)
 	createState.SetDetails(object)
```
fat-fellow, line 45: DRY

requilence (changes requested):
please add clarifying comment to the state.originalCreatedTimestamp. Right now it only mentions imported objects.

requilence (approved):

PR #1853, opened 2024-11-21
GO-4560 Mentions should be 1st in RelationsListWithValue response

KirillSto:
https://linear.app/anytype/issue/GO-4560/mentions-should-be-put-on-1st-place-in-relationlistwithvalue-output

- Mentions should appear as first relation in RelationListWithValue output (if presented)
- Rename Last Used date -> Last used date for consistency
- For dates with time included enrich date names with time hh:mm

### DIFF
--- a/core/block/detailservice/relations.go
+++ b/core/block/detailservice/relations.go
@@ -140,18 +140,19 @@ func generateFilter(value *types.Value) func(v *types.Value) bool {
 
 	// date object section
 
-	ts, _, err := dateutil.ParseDateId(stringValue)
+	dateObject, err := dateutil.BuildDateObjectFromId(stringValue)
 	if err != nil {
 		log.Error("failed to parse Date object id", zap.Error(err))
 		return equalOrHasFilter
 	}
 
-	start := timeutil.CutToDay(ts)
+	start := timeutil.CutToDay(dateObject.Time())
 	end := start.Add(24 * time.Hour)
 	startTimestamp := start.Unix()
 	endTimestamp := end.Unix()
 
-	shortId := dateutil.TimeToDateId(start, false)
+	startDateObject := dateutil.NewDateObject(start, false)
+	shortId := startDateObject.Id()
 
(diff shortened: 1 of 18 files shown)


core/block/detailservice/relations.go
```diff
@@ -97,6 +104,14 @@ func (s *service) ListRelationsWithValue(spaceId string, value *types.Value) ([]
 		}
 	}
 
+	if mentionsCounter > 0 {
```
deff7, line 107: Maybe write custom sorter with `sort.Slice`? Mentions first, then all other relations alphabetically

core/block/detailservice/relations.go
```diff
@@ -126,15 +141,19 @@ func generateFilter(value *types.Value) func(v *types.Value) bool {
...
 	if err != nil {
 		log.Error("failed to parse Date object id", zap.Error(err))
 		return equalOrHasFilter
 	}
 
-	shortId := dateutil.TimeToDateId(ts)
+	shortId := dateutil.TimeToDateId(ts, false)
 
-	start := ts.Truncate(24 * time.Hour)
+	start, _, err := dateutil.ParseDateId(shortId)
```
deff7, line 152: why serialize and then parse id? If you want to drop hours, we have

```
func CutToDay(t time.Time) time.Time {
	roundTime := time.Date(t.Year(), t.Month(), t.Day(), 0, 0, 0, 0, t.Location())
	return roundTime
}
```

core/block/detailservice/relations.go
KirillSto: Yes, I was thinking about it, but it will increase calculations time and sorting looks lil bit ugly :see_no_evil: 

```
sort.Slice(keys, func(i, j int) bool {
		if keys[i] == bundle.RelationKeyMentions.String() {
			return true
		}
		if keys[j] == bundle.RelationKeyMentions.String() {
			return false
		}
		return keys[i] < keys[j]
	})
```

deff7 (approved):

PR #1870, opened 2024-11-26
GO-4586 | GO-4590 - Do not generate long date ids + Change date format again

KirillSto:
https://linear.app/anytype/issue/GO-4586/do-not-generate-long-ids
https://linear.app/anytype/issue/GO-4590/support-the-date-format-once-again

- We should not allow to generate second-precision date object ids for now, as old clients could not show these objects correctly
- Change date format to Mon, Dec 25, 2124 11:34 PM

### DIFF
--- a/util/dateutil/util.go
+++ b/util/dateutil/util.go
@@ -10,8 +10,8 @@ import (
 const (
 	shortDateIdLayout   = "2006-01-02"
 	dateIdLayout        = "2006-01-02-15-04-05Z-0700"
-	shortDateNameLayout = "02 Jan 2006"
-	dateNameLayout      = "02 Jan 2006 15:04"
+	shortDateNameLayout = "Mon, Jan 02, 2006"
+	dateNameLayout      = "Mon, Jan 02, 2006 3:04 PM"
 )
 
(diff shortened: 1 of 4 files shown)


util/dateutil/util.go
```diff
@@ -10,8 +10,8 @@ import (
 const (
 	shortDateIdLayout   = "2006-01-02"
 	dateIdLayout        = "2006-01-02-15-04-05Z-0700"
-	shortDateNameLayout = "02 Jan 2006"
-	dateNameLayout      = "02 Jan 2006 15:04"
+	shortDateNameLayout = "Mon, Jan 02, 2006"
```
requilence, line 13: i guess you should not revert the date string format? As we agreed to change it

requilence (changes requested):

util/dateutil/util.go
KirillSto: That's brand new prod requirement!
https://linear.app/anytype/issue/GO-4590/support-the-date-format-once-again
Extra comma added

requilence (approved):

PR #1871, opened 2024-11-26
GO-4577 Add date objects to graph

KirillSto:
https://linear.app/anytype/issue/GO-4577/mix-in-dates-to-objectgraph

Add date objects to graph

### DIFF
--- a/core/block/object/objectgraph/graph.go
+++ b/core/block/object/objectgraph/graph.go
@@ -128,7 +128,7 @@ func (gr *Builder) buildGraph(
 
 		outgoingRelationLink := make(map[string]struct{}, 10)
 		edges = gr.appendRelations(rec, relations, edges, existedNodes, sourceId, outgoingRelationLink)
-		nodesToAdd := make([]*types.Struct, 0)
+		var nodesToAdd []*types.Struct
 		nodesToAdd, edges = gr.appendLinks(req.SpaceId, rec, outgoingRelationLink, existedNodes, edges, sourceId)
 
 		if len(nodesToAdd) != 0 {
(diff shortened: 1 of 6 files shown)


core/block/object/objectgraph/graph.go
```diff
@@ -128,7 +128,12 @@ func (gr *Builder) buildGraph(
 
 		outgoingRelationLink := make(map[string]struct{}, 10)
 		edges = gr.appendRelations(rec, relations, edges, existedNodes, sourceId, outgoingRelationLink)
-		edges = gr.appendLinks(req.SpaceId, rec, outgoingRelationLink, existedNodes, edges, sourceId)
+		var nodesToAdd []*types.Struct
+		nodesToAdd, edges = gr.appendLinks(req.SpaceId, rec, outgoingRelationLink, existedNodes, edges, sourceId)
+
+		if len(nodesToAdd) != 0 {
```
requilence, line 134: no need to do an empty check. `append` will be noop in case `nodesToAdd` is empty

core/block/object/objectgraph/graph.go
```diff
@@ -191,14 +196,31 @@ func (gr *Builder) appendLinks(
...
+			}
+			if err != nil {
+				log.Error("get details of Date object", zap.String("objectId", link), zap.Error(err))
+				continue
+			}
+			existedNodes[link] = struct{}{}
+			nodes = append(nodes, details[0].Details)
+		}
+
 		if sbType != smartblock.SmartBlockTypeFileObject {
```
requilence, line 224: move to switch default case?

requilence (changes requested):

core/block/object/objectgraph/graph.go
KirillSto: Thank you! 
Also we should map relations in details as requested in keys field of request

PR #1872, opened 2024-11-26
GO-4587 ObjectSearch should return dates by ids

KirillSto:
https://linear.app/anytype/issue/GO-4587/objectsearch-by-ids-should-return-date-objects

If ObjectSearch request includes filters by id and values are date object ids, than response should be enriched with date objects' details

### DIFF
--- a/core/date/suggest.go
+++ b/core/date/suggest.go
@@ -18,38 +18,18 @@ import (
 	"github.com/anyproto/anytype-heart/pkg/lib/localstore/objectstore"
 	"github.com/anyproto/anytype-heart/space"
 	"github.com/anyproto/anytype-heart/util/dateutil"
+	"github.com/anyproto/anytype-heart/util/pbtypes"
 )
 
-func EnrichRecordsWithDateSuggestion(
+func EnrichRecordsWithDateSuggestions(
 	ctx context.Context,
 	records []database.Record,
 	req *pb.RpcObjectSearchRequest,
 	store objectstore.ObjectStore,
 	spaceService space.Service,
 ) ([]database.Record, error) {
-	dt := suggestDateForSearch(time.Now(), req.FullText)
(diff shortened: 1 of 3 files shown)


core/date/suggest.go
```diff
@@ -18,55 +18,69 @@ import (
...
+	if err != nil {
+		return nil, fmt.Errorf("get space: %w", err)
+	}
 
-	// Don't duplicate search suggestions
-	var found bool
-	for _, r := range records {
-		if r.Details == nil || r.Details.Fields == nil {
+	for _, id := range ids {
+		if isRecordDuplicated(records, id) {
```
requilence, line 42: isRecordDuplicated -> recordsHasId

core/date/suggest.go
```diff
@@ -18,55 +18,69 @@ import (
...
 	ctx context.Context,
 	records []database.Record,
 	req *pb.RpcObjectSearchRequest,
 	store objectstore.ObjectStore,
 	spaceService space.Service,
 ) ([]database.Record, error) {
-	dt := suggestDateForSearch(time.Now(), req.FullText)
-	if dt.IsZero() {
+	ids := suggestDateObjectIds(req)
+	if ids == nil {
```
requilence, line 32: It is not a good practice to compare slice with nil, cause it can be empty but not nil. In your case if client sends id filter but it has no date ids suggestDateObjectIds will return empty slice instead of nil.

requilence (changes requested):

core/date/suggest.go
KirillSto: Fixed!

core/date/suggest.go
KirillSto: Yeah, thank you for mentioning that. Fixed!

requilence (approved):

PR #1897, opened 2024-12-03
GO-4277 Revise system objects using bundled info

KirillSto:
https://linear.app/anytype/issue/GO-4277/make-the-list-of-relations-hidden-for-all-objects

We should use bundled relations and types as origin bundled objects information, as Marketplace store need to be indexed first


requilence:
1. if user want to make a featured relation not-featured client need to send 2 requests? It will be also 2 changes and a chance that only one of them got synced
2. what if somehow we will have the same relation in both featured list and recommended? should we take a responsibility to prevent such cases on our side?

PR #1919, opened 2024-12-06
GO-4079 Fix bool unset

KirillSto:
https://linear.app/anytype/issue/GO-4079/objectlistsetdetails-fix-bool-false

Fix ObjectListModifyDetailValues for setting bool details to false


fat-fellow (changes requested):
discussed in dm

PR #1945, opened 2024-12-12
GO-4711 Fix creator in usecases

KirillSto:
https://linear.app/anytype/issue/GO-4711/create-new-account-on-04332-alpha-types-and-relation-created-by-user

Creator=anytype did not propagate to objects buing created on usecase import


AnastasiaShemyakinskaya (dismissed):

AnastasiaShemyakinskaya (commented):
Also should we add test for import?

PR #1989, opened 2025-01-06
GO-4719: After import I can no longer add an object to a collection

AnastasiaShemyakinskaya:
Add mapping of default object type id to the new one in import, so we get new id of default type in result collection object

### DIFF
--- a/core/block/export/export.go
+++ b/core/block/export/export.go
@@ -345,13 +345,13 @@ func (e *exportContext) queryAndFilterObjectsByRelation(spaceId string, reqIds [
 	const singleBatchCount = 50
 	for j := 0; j < len(reqIds); {
 		if j+singleBatchCount < len(reqIds) {
-			records, err := e.queryObjectsByIds(spaceId, reqIds[j:j+singleBatchCount], relationKey)
+			records, err := e.queryObjectsByRelation(spaceId, reqIds[j:j+singleBatchCount], relationKey)
 			if err != nil {
 				return nil, err
 			}
 			allObjects = append(allObjects, records...)
 		} else {
-			records, err := e.queryObjectsByIds(spaceId, reqIds[j:], relationKey)
+			records, err := e.queryObjectsByRelation(spaceId, reqIds[j:], relationKey)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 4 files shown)


core/block/export/export.go
```diff
@@ -413,6 +420,55 @@ func (e *exportContext) processProtobuf() error {
...
 }
 
+func (e *exportContext) addDependentObjectsFromDataview() error {
+	var (
+		viewDependantObjectsIds []string
+		err                     error
+	)
+	for id, details := range e.docs {
+		layout := details.GetInt64(bundle.RelationKeyLayout)
+		if layout == int64(model.ObjectType_collection) {
```
KirillSto, line 430: These settings could be set for views of sets by relations as well, so check on **set** layout should be added

core/block/export/export.go
```diff
@@ -413,6 +420,55 @@ func (e *exportContext) processProtobuf() error {
...
+		e.docs[id] = object.Details
+	}
+	return nil
+}
+
+func (e *exportContext) getViewDependentObjects(id string, viewDependantObjectsIds []string) ([]string, error) {
+	err := cache.Do(e.picker, id, func(sb sb.SmartBlock) error {
+		st := sb.NewState()
+		return st.Iterate(func(b simple.Block) (isContinue bool) {
+			if dv := b.Model().GetDataview(); dv != nil {
```
KirillSto, line 452: This code could be simplified:

```
dv := b.Model().GetDataview()
if dv == nil {
  return true
}
for ...
return false
```

core/block/export/export.go
KirillSto: And unit test on it

core/block/export/export.go
```diff
@@ -413,6 +420,55 @@ func (e *exportContext) processProtobuf() error {
 	return nil
 }
 
+func (e *exportContext) addDependentObjectsFromDataview() error {
+	var (
+		viewDependantObjectsIds []string
```
KirillSto, line 425: viewDepend**e**ntObjectsIds :upside_down_face:

core/block/export/export.go
```diff
@@ -413,6 +420,55 @@ func (e *exportContext) processProtobuf() error {
...
+	for id, details := range e.docs {
+		layout := details.GetInt64(bundle.RelationKeyLayout)
+		if layout == int64(model.ObjectType_collection) {
+			viewDependantObjectsIds, err = e.getViewDependentObjects(id, viewDependantObjectsIds)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	viewDependantObjects, err := e.queryAndFilterObjectsByRelation(e.spaceId, viewDependantObjectsIds, bundle.RelationKeyId)
```
KirillSto, line 437: viewDepend**e**ntObjects :upside_down_face:

core/block/export/export.go
AnastasiaShemyakinskaya: Default object type is only applicable for collection, doesn't it? 🧐

core/block/export/export.go
AnastasiaShemyakinskaya: Thanks, I always make error in this word 😅

core/block/export/export.go
KirillSto: If set is aggregated by relation, then we can choose default object type for views

KirillSto (approved):

PR #1997, opened 2025-01-09
GO-4677 RPC to list conficting relations in type

KirillSto:
https://linear.app/anytype/issue/GO-4677/rpc-to-return-conflict-relations-among-objects-of-one-type

RPC that returns list of relation ids that present in details of objects of particular type, but does not included in recommended relations of this type


core/block/detailservice/relations.go
```diff
@@ -91,6 +93,70 @@ func (s *service) objectTypeSetRelations(
...
+				Condition:   model.BlockContentDataviewFilter_Equal,
+				Value:       domain.Int64(int64(model.ObjectType_relation)),
+			},
+		},
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch relations by keys: %w", err)
+	}
+	if len(records) != len(allRelationKeys) {
+		return nil, fmt.Errorf("failed to query relations from store, expected %d record, got %d", len(allRelationKeys), len(records))
```
deff7, line 146: Maybe remove %d from the error string?

KirillSto: Done!
Also removed %d for error on first store query

deff7 (approved):

PR #2001, opened 2025-01-10
GO-4418: imported objects are not added to a new collection

AnastasiaShemyakinskaya:
1. The problem occurred when imported archive or directory have directories without files but only with another directories inside them. 
Example cases:
#### Case with archive:
```
Archive.zip
|____directory1
         |_____file.md
|____directory2
         |_____collection.csv 
         |_____file.md
```

#### Case with directory:
```
directory
|____directory1
         |____directory2
                  |_____collection.csv 
                  |_____file.md
```

Imported source doesn't contains any files in root directory In such cases, so the logic with determination of root files is broken for markdown import. During markdown import we create root collection (Markdown import) and add root files. But in such cases there are no files in root, only directories

2. So to fix, I try to determine actual root directories in archive/directories. 
* In the first case, we have 2 root directories directory1 and directory2, because they contains files.
* In the second case, we have 1 root directory `directory/directory1/directory2`, because directory2 is the first directory with actual files. Result Markdown collection will have collection.csv and file.md objects, because they are now considered as root files.


core/block/import/common/source/directory.go
```diff
@@ -80,7 +88,41 @@ func (d *Directory) CountFilesWithGivenExtensions(extension []string) int {
...
+func findNonEmptyDirs(files map[string]struct{}) []string {
+	dirs := make([]string, 0, len(files))
+	for file := range files {
+		dir := filepath.Dir(file)
+		if dir == "." {
+			return []string{dir}
+		}
+		dirs = append(dirs, dir)
+	}
+	sort.Strings(dirs)
```
KirillSto, line 106: Why do we need sorting here?
Maybe it is better to do sorting for result slice at the end of the func?

core/block/import/common/source/directory.go
```diff
@@ -80,7 +88,41 @@ func (d *Directory) CountFilesWithGivenExtensions(extension []string) int {
...
+	var result []string
+	visited := make(map[string]bool)
```
KirillSto, line 108: It seems visited map is not used

core/block/import/common/source/directory.go
```diff
@@ -80,7 +88,41 @@ func (d *Directory) CountFilesWithGivenExtensions(extension []string) int {
...
+
+	for _, dir := range dirs {
+		if isSubdirectoryOfAny(dir, result) {
+			continue
+		}
+		result = lo.Union(result, []string{dir})
```
KirillSto, line 114: Union looks really complicated inside. Two-layer cycles and maps allocations.
Maybe it is better to append directories to single slice and call lo.Uniq afterwards :thinking:

KirillSto: Or use our own logic with visited map, as you started to implement first

core/block/import/common/source/directory.go
AnastasiaShemyakinskaya: We need this logic, because it is import to start iteration from the dir with smallest path, as we are checking if current directory is sub dir of existing directories.

core/block/import/common/source/directory.go
AnastasiaShemyakinskaya: Add checking of visited map in cycle

core/block/import/common/source/directory.go
AnastasiaShemyakinskaya: I rewrite this logic and decided to make result slice as map. Because during import we check if directory exists is result slice. But maps are faster than slice for searching. So now result is a map

KirillSto (approved):

PR #2018, opened 2025-01-16
GO-4234 Layout conversion restrictions

KirillSto:
https://linear.app/anytype/issue/GO-4234/divide-layouts-page-basicprofilenotetask-list-setcollection-tag

- on layout change we should check if we do not swich between different layout groups Page and List
- same check should be done on recommendedLayout switch


core/block/editor/basic/details.go
```diff
@@ -106,19 +111,18 @@ func (bs *basic) updateDetails(update func(current *domain.Details) (*domain.Det
...
 	keys := make([]domain.RelationKey, 0, len(details))
 	for _, detail := range details {
 		update, err := bs.createDetailUpdate(s, detail)
-		if err == nil {
-			updates = append(updates, update)
-			keys = append(keys, update.Key)
-		} else {
-			log.Errorf("can't set detail %s: %s", detail.Key, err)
+		if err != nil {
+			return nil, nil, err
```
deff7, line 120: Do you think we should return with an error? We ignored it intentionally

core/block/editor/basic/details.go
```diff
@@ -293,6 +297,15 @@ func (bs *basic) setDetailSpecialCases(st *state.State, detail domain.Detail) er
 		// special case when client sets the layout detail directly instead of using SetLayoutInState command
 		return bs.SetLayoutInState(st, model.ObjectTypeLayout(detail.Value.Int64()), false)
 	}
+	if detail.Key == bundle.RelationKeyRecommendedLayout {
+		fromLayout := st.Details().GetInt64(bundle.RelationKeyRecommendedLayout)
+		toLayout := detail.Value.Int64()
+		// nolint:gosec
+		if !isLayoutConversionAllowed(model.ObjectTypeLayout(fromLayout), model.ObjectTypeLayout(toLayout)) {
```
deff7, line 304: why is it here? Layout conversion checks are in layout converter

core/block/editor/basic/details.go
KirillSto: No, I agree that logging an error and simple detail change ignore are enough. Fixed

core/block/editor/basic/details.go
KirillSto: Fixed

deff7 (approved):

PR #2023, opened 2025-01-20
GO-3189: set checkbox relation as false in ObjectRelationAdd command

AnastasiaShemyakinskaya:
Decided to rewrite this task as we are removing relation links. So I just set details as false for checkoboxes relations, if we add such relation to object

### DIFF
--- a/core/details.go
+++ b/core/details.go
@@ -188,8 +188,7 @@ func (mw *Middleware) ObjectRelationAdd(cctx context.Context, req *pb.RpcObjectR
 			}
 			format, err := mw.extractRelationFormat(current, objectStore, key)
 			if err != nil {
-				current.Set(domain.RelationKey(key), domain.Null())
-				continue
+				log.Errorf("failed to fetch relation from store to get format %s, falling back to basic", err)
 			}
 			switch format {
 			case model.RelationFormat_checkbox:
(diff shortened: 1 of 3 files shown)


deff7 (approved):

core/details.go
```diff
@@ -177,12 +180,21 @@ func (mw *Middleware) ObjectRelationAdd(cctx context.Context, req *pb.RpcObjectR
...
+	objectStore := mustService[objectstore.ObjectStore](mw)
 	err := detailsService.ModifyDetails(ctx, req.ContextId, func(current *domain.Details) (*domain.Details, error) {
 		for _, key := range req.RelationKeys {
 			if current.Has(domain.RelationKey(key)) {
 				continue
 			}
-			current.Set(domain.RelationKey(key), domain.Null())
+			format, err := mw.extractRelationFormat(current, objectStore, key)
+			if err != nil {
+				continue
```
KirillSto, line 191: It seems in case of failure during relation fetch from objectstore we should fall back to setting Null value to the object

core/details.go
```diff
@@ -177,12 +180,21 @@ func (mw *Middleware) ObjectRelationAdd(cctx context.Context, req *pb.RpcObjectR
...
+			}
+			if format == model.RelationFormat_checkbox {
```
KirillSto, line 193: Maybe we could use switch with default here?

KirillSto (changes requested):

KirillSto (approved):

PR #2036, opened 2025-01-22
GO-4889: add state filters to export

AnastasiaShemyakinskaya:
1. Add new function `Filter` in state. Based on provided filters it remove blocks or filter relations by layout
2. Add parameter in `Export` function - `LinkStateFilters` and use it in Web publishing. We provide filters: remove all blocks and relations white list by layout. So after export we receive state without blocks and only with provided relations white list. So for publishing we can remove not needed information from state for security reason and also to reduce size of archive.
3. Use `DendenObjectsIDs` to retrieve links for objects in export.
4. Change docs map in export. So now it contains information if provided doc is a link. So we can use filter for its state.


zarkone:
https://linear.app/anytype/issue/JS-6256/the-limit-exceeded-message-does-not-match-the-design

PR #2065, opened 2025-01-30
GO-4945: Bug: not all published pages are shown in "My Sites" settings

AnastasiaShemyakinskaya:
1. Add new parameter in `PublishState` model - details
2. Save details from objects store for each published objects in this field during `PublishingList` call. So clients can retrieve objects information from these details

https://linear.app/anytype/issue/GO-4945/bug-not-all-published-pages-are-shown-in-my-sites-settings

### DIFF
--- a/core/publish/service.go
+++ b/core/publish/service.go
@@ -17,6 +17,7 @@ import (
 	"github.com/anyproto/anytype-publish-server/publishclient/publishapi"
 	"github.com/gogo/protobuf/jsonpb"
 	"github.com/gogo/protobuf/proto"
+	"github.com/gogo/protobuf/types"
 	"go.uber.org/zap"
 	"golang.org/x/exp/slices"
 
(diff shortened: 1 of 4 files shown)


core/publish/service.go
```diff
@@ -440,11 +445,26 @@ func (s *service) PublishList(ctx context.Context, spaceId string) ([]*pb.RpcPub
 			Timestamp: publish.Timestamp,
 			Size_:     publish.Size_,
 			JoinSpace: version.JoinSpace,
+			Details:   details,
 		})
 	}
 	return pbPublishes, nil
 }
 
+func (s *service) retrieveObjectDetails(publish *publishapi.Publish) *types.Struct {
```
KirillSto, line 454: Maybe it is better to do optimisation and Query all objects in single call?
Retrieve function could recieve []*Publish and return map[string]*types.Struct for example

KirillSto: However in this case you will need to iterate over publishes/ids three times :thinking:

zarkone (approved):

PR #2079, opened 2025-02-04
GO-4780 change orderid to afterorderid in subscriptiongo

deff7:


core/block/source/store_apply.go
```diff
@@ -16,31 +16,36 @@ type storeApply struct {
...
-	prevChange *objecttree.Change
-
-	nextCachedOrder string
-	nextCacheChange map[string]struct{}
+	needFetchPrevOrderId bool
 }
 
-func (a *storeApply) Apply() (err error) {
-	iterErr := a.ot.IterateRoot(UnmarshalStoreChange, func(change *objecttree.Change) bool {
+func (a *storeApply) Apply() error {
```
KirillSto, line 22: Why Apply does not return errors on tree iterations?

KirillSto (approved):

PR #2096, opened 2025-02-06
GO-5047: Space Member is not displayed in Created by/last modified by

AnastasiaShemyakinskaya:
Use `CreatorModifierWorkspace` for dependent objects, so we can extract participants from according details


zarkone (approved):
:raised_hands:

PR #2145, opened 2025-02-20
GO-4284 Remove Relation links

KirillSto:
https://linear.app/anytype/issue/GO-4284/get-rid-of-relation-links

// TODO


core/block/import/common/common.go
```diff
@@ -296,7 +293,12 @@ func handleObjectRelation(st *state.State, oldIDtoNew map[string]string, v domai
 		}
 		return
 	}
-	objectsIDs := v.StringList()
+
+	objectsIDs, ok := v.TryStringList()
```
AnastasiaShemyakinskaya, line 297: Only objects relations have list in details?

KirillSto: No, in general SringList could carry relation of every format. 
However, our clients do not allow to make relations of `format != object` and with StringList type

KirillSto: That's really good, you paid attention to this code, because I don't really know, how to check format here. We could propagate store index and it would be useful only for bundled relations and relations that already present in the store. Other relations are not indexed, because they are not imported yet

KirillSto: All we rely on here right now is cid.Parse check, but of course that is not the best check :slightly_smiling_face:

core/block/editor/state/details.go
```diff
@@ -0,0 +1,210 @@
...
+
+	return s.localDetails
+}
+
+func (s *State) CombinedDetails() *domain.Details {
+	// TODO Implement combined details struct with two underlying details
+	return s.Details().Merge(s.LocalDetails())
+}
+
+func (s *State) AllRelationKeys() []domain.RelationKey {
```
deff7, line 34: Every call it allocate 3 slices, I suggest you to use Iterators to make this code to have zero allocations. AllRelationKeys iterator will call Details.Iterate and LocalDetails.Iterate.

https://pkg.go.dev/slices#Collect is useful to collect result of iteration to slice

Also you can add new function ContainsInSeq to our library. This function will search for element in iter.Seq and have no allocations. So there will no need to convert iterator to slice and then use slices.Contains.

core/block/import/common/common.go
AnastasiaShemyakinskaya: I think, we can get bundled relations format without store the same way we do it in system objects revisor. As for imported relations, we should extract their format from according snapshot and provide this information to this function. But I think, it is a separate task

KirillSto: I have filed a separate task on it: https://linear.app/anytype/issue/GO-5214/resolve-relation-format-from-snapshots-on-import

core/block/editor/state/details.go
KirillSto: I have also introduced `State.iteratreKeys() iter.Seq[domain.Key]` method. It can be made global, but I am not sure about security

AnastasiaShemyakinskaya (approved):

PR #2155, opened 2025-02-25
GO-5165 Migrate types with recommendedFeaturedRelations

KirillSto:
https://linear.app/anytype/issue/GO-5165/add-to-all-existing-and-newly-created-types-feature-relations-object

Type, Tag and Backlinks realtions should be added to recommendedFeaturedRelations of all type objects

### DIFF
--- a/core/block/editor/page.go
+++ b/core/block/editor/page.go
@@ -289,6 +289,10 @@ func (p *Page) CreationStateMigration(ctx *smartblock.InitContext) migration.Mig
 func (p *Page) StateMigrations() migration.Migrations {
 	return migration.MakeMigrations([]migration.Migration{
 		{
+			Version: 2,
+			Proc:    func(s *state.State) {},
+		},
+		{
 			Version: 3,
 			Proc:    p.featuredRelationsMigration,
 		},
--- a/core/block/editor/page.go
+++ b/core/block/editor/page.go
@@ -304,7 +304,7 @@ func (p *Page) featuredRelationsMigration(s *state.State) {
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


core/relationutils/recommended.go
```diff
@@ -19,7 +19,7 @@ type ObjectIDDeriver interface {
 }
 
 var (
-	defaultFeaturedRelationKeys = []domain.RelationKey{
+	DefaultFeaturedRelationKeys = []domain.RelationKey{
```
AnastasiaShemyakinskaya, line 22: May be it's better to return defaultFeaturedRelationKeys in function. Because now everyone can modify this slice in application

deff7 (approved):

core/block/editor/page.go
```diff
@@ -286,8 +289,46 @@ func (p *Page) CreationStateMigration(ctx *smartblock.InitContext) migration.Mig
...
+func (p *Page) featuredRelationsMigration(s *state.State) {
+	if p.Type() != coresb.SmartBlockTypeObjectType {
+		return
+	}
+
+	if s.HasRelation(bundle.RelationKeyRecommendedFeaturedRelations.String()) {
+		return
+	}
+
+	featuredRelationIds := make([]string, len(relationutils.DefaultFeaturedRelationKeys))
```
deff7, line 307: `make([]string, 0, len(relationutils.DefaultFeaturedRelationKeys))`

core/relationutils/recommended.go
KirillSto: Yeah, good point, let me fix that!

PR #2157, opened 2025-02-25
GO-5167 Exclude featuredRelations from template

KirillSto:
https://linear.app/anytype/issue/GO-5167/do-not-take-featuredrelations-from-template

- we should not take `featuredRelations` from template state when creation new obejct
- we should not pre-fill `featuredRelations` when creating new object (`description` is an exception)


core/block/editor/page.go
```diff
@@ -284,10 +278,5 @@ func (p *Page) CreationStateMigration(ctx *smartblock.InitContext) migration.Mig
 }
 
 func (p *Page) StateMigrations() migration.Migrations {
-	return migration.MakeMigrations([]migration.Migration{
-		{
-			Version: 2,
-			Proc:    template.WithAddedFeaturedRelation(bundle.RelationKeyBacklinks),
-		},
-	})
+	return migration.MakeMigrations([]migration.Migration{})
```
AnastasiaShemyakinskaya, line 281: Are you sure it is safe to remove migration? We probably have objects with version 2. When we are going to add some new migration, we will make migration with version 1 (because we don't have any migration here). And this migration won't work for old objects with version 2

core/block/template/templateimpl/impl.go
```diff
@@ -181,7 +183,7 @@ func extractTargetDetails(originDetails *domain.Details, templateDetails *domain
 		templateVal := templateDetails.Get(key)
 		if templateVal.Ok() {
 			inTemplateEmpty := templateVal.IsEmpty()
-			if key == bundle.RelationKeyResolvedLayout {
+			if key == bundle.RelationKeyLayout {
```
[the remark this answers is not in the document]
deff7: and comment why it's necessary

core/block/editor/page.go
KirillSto: Yeah, I have removed it because of https://github.com/anyproto/anytype-heart/pull/2155, where we decided to add migration with version 3. Anyway this migration should be left to set migration version with empty function

core/block/template/templateimpl/impl.go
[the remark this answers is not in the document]
KirillSto: Done!

core/block/editor/page.go
KirillSto: Fixed!

deff7 (approved):

PR #2176, opened 2025-02-28
GO-5187 Move layout resolution to hook

KirillSto:
https://linear.app/anytype/issue/GO-5187/change-of-type-settings-should-leave-to-deletion-of-corresponding

On object type details change we should check all objects and templates corresponding to this type.

If **recommendedLayout** of type is changed, we should update **resolvedLayout** of these objects. Futhermore there could be old objects remaining **layout** detail, so this detail should be removed from state if its value equals to old or new value of **recommendedLayout**.

If **layoutAlign** is changed in type, we also should check that objects and templates corresponding to type do not have **layoutAlign** detail. We delete it in case its value corresponds to old or new value of **layoutAlign** detail of type.

If **recommendedFeaturedRelations** is changed, we do the same check as for **layoutAlign**, but instead of deleting detail from state, we clear it. **FeaturedRelations** detail could carry **description** value, and that is the only role of this relation for now


core/block/editor/objecttype.go
```diff
@@ -0,0 +1,438 @@
...
+		},
+	}})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get templates with this target type: %w", err)
+	}
+
+	return append(records, templates...), nil
+}
+
+func collectRelationsChanges(
```
deff7, line 388: maybe return structure? Too many values

KirillSto: Done!

deff7 (approved):

PR #2190, opened 2025-03-04
GO-5143 Sync relLinks on view relations update

KirillSto:
https://linear.app/anytype/issue/GO-5143/changing-visibility-does-not-work-for-types-as-widgets

On view.Relations update we should sync them with dataview.RelationLinks


core/block/dataviewservice/service.go
```diff
@@ -242,16 +263,71 @@ func (s *service) ReorderDataviewViewRelations(
...
 }
 
+func (s *service) syncViewRelationsAndRelationLinks(objectId, viewId string, dv dvblock.Block) {
+	view, err := dv.GetView(viewId)
+	if err != nil {
+		log.Error("failed to get view", zap.String("viewId", viewId), zap.Error(err))
+		return
+	}
+
+	relationLinksKeys := map[string]struct{}{}
```
AnastasiaShemyakinskaya, line 286: This map could be preallocated

core/block/dataviewservice/service.go
```diff
@@ -242,16 +263,71 @@ func (s *service) ReorderDataviewViewRelations(
...
+			spaceId, err := s.idResolver.ResolveSpaceID(objectId)
+			if err != nil {
+				return nil, fmt.Errorf("failed to resolve space id: %w", err)
+			}
+			spaceIndex = s.objectStore.SpaceIndex(spaceId)
+		}
+		return spaceIndex.GetRelationLink(key)
+	}
+
+	currentViewKeys := map[string]struct{}{}
```
AnastasiaShemyakinskaya, line 303: The same

core/block/dataviewservice/service.go
KirillSto: Done!

AnastasiaShemyakinskaya (approved):

PR #2211, opened 2025-03-07
GO-5099: add space name to publishing archive and page

AnastasiaShemyakinskaya:
1. Added new parameter in export `IncludeSpace`, so export also fetch space object with icon
2. For publishing we set this parameter as true


ra3orblade (approved):

zarkone (approved):

cmd/grpcserver/performance.go
```diff
@@ -307,7 +307,8 @@ func getTableSizes(mw *core.Middleware) (tables map[string]uint64) {
 	cfg := mw.GetApp().MustComponent(config.CName).(*config.Config)
 	sqliteDBs, err := findSQLiteFiles(cfg.GetNewSpaceStorePath())
 	if err != nil {
-		log.Fatalf("sqlite find error: %v", err)
+		fmt.Printf("sqlite find error: %v\n", err)
```
KirillSto, line 310: Dubug message remains :smiley:

core/block/export/export.go
```diff
@@ -164,6 +164,7 @@ type exportContext struct {
 	linkStateFilters *state.Filters
 	isLinkProcess    bool
 	includeBackLinks bool
+	includeSpace     bool
```
KirillSto, line 167: Maybe you can add pb.RpcObjectListExportRequest as separate field of exportContext?
In this case you can differentiate export settings required by client call and fields in exportContext that you fill during the very context. That's just a proposal :slightly_smiling_face:

cmd/grpcserver/performance.go
AnastasiaShemyakinskaya: It is fine, because It is for performance metrics. This command uses `fmt` to print. It is fix for unit tests

core/block/export/export.go
AnastasiaShemyakinskaya: I can create separate task for this, as it requires a lot of changes in export

PR #2221, opened 2025-03-11
Refactoring of Pb converter proposal

KirillSto:
Use the very Pb structure to accumulate necessary info during import


core/block/import/pb/converter.go
```diff
@@ -39,11 +39,22 @@ const (
 
 var ErrNotAnyBlockExtension = errors.New("not JSON or PB extension")
 
+type snapshotSet struct {
```
AnastasiaShemyakinskaya, line 42: I think this name is not suitable.🧐 Usually it is used for maps that are used as sets

KirillSto: Please take a look!
I have introduced **SnapshotList** structure in **common** package

AnastasiaShemyakinskaya (approved):

core/block/import/pb/converter.go
```diff
@@ -448,13 +439,13 @@ func (p *Pb) setSourceFilePath(snapshot *common.SnapshotModel) {
 	snapshot.Data.Details.SetString(bundle.RelationKeySourceFilePath, sourceFilePath)
 }
 
-func (p *Pb) shouldImportSnapshot(snapshot *common.Snapshot, needToImportWidgets bool, importType pb.RpcObjectImportRequestPbParamsType) bool {
-	return (snapshot.Snapshot.SbType == smartblock.SmartBlockTypeWorkspace && importType == pb.RpcObjectImportRequestPbParams_SPACE) ||
+func (p *Pb) shouldImportSnapshot(snapshot *common.Snapshot) bool {
+	return (snapshot.Snapshot.SbType == smartblock.SmartBlockTypeWorkspace && p.isNewSpace) ||
```
AnastasiaShemyakinskaya, line 443: Are you sure that `p.isNewSpace` should be used

AnastasiaShemyakinskaya (changes requested):

AnastasiaShemyakinskaya (approved):

PR #2250, opened 2025-03-21
GO-5307 auto-widgets improvements

requilence:
Bin auto-widget
- For existing spaces: if there are some archived objects - install widget
- For new spaces: when archiving the first object widget is installed
- If widget got removed by user it will be not auto-installed again

Favorite auto-widget
- we do not do auto creation of favorite widget on start because it was not a default behavior
- if user add an object to favorite, the favorite widget will be automatically added. But only in case there is no existing favorite widget added by user (including in the previous release)

Type widget improvements
- only add widget in case you are creating the first object of this type(without taking use-case and hidden objects into account)
- do not create widget for specific list of types(like templates, types, relations)
-


core/block/editor/dashboard.go
```diff
@@ -75,6 +78,40 @@ func (p *Dashboard) StateMigrations() migration.Migrations {
 	}})
 }
 
+// autoInstallBinWidget installs bin widget for the existing users if it was not added/removed before and the user has archived objects
```
fat-fellow, line 81: misleading description

core/block/editor/dashboard.go
```diff
@@ -75,6 +78,40 @@ func (p *Dashboard) StateMigrations() migration.Migrations {
...
+func (p *Dashboard) autoInstallFavoriteWidget() error {
```
fat-fellow, line 82: Looks like these two auto-install functions can reuse the same code and be called with a widgetObjectId argument

requilence: actually it was wrong because we don't need to create it with the same logic as archive. We only need to create explicitly when calling objectSetIsFavorite

core/block/editor/widget/widget.go
```diff
@@ -15,11 +18,16 @@ const (
...
 	DefaultWidgetCollection = "collection"
+	DefaultWidgetBin        = "bin"
 	DefaultWidgetRecentOpen = "recentOpen"
+	wrapperBlockIdSuffix    = "-wrapper" // in case blockId is specifically provided to avoid bad tree merges
 )
 
 type Widget interface {
 	CreateBlock(s *state.State, req *pb.RpcBlockCreateWidgetRequest) (string, error)
+	// AddAutoWidget adds a widget block. If widget with the same targetId was installed/removed before, it will not be added again.
+	// blockId is optional and used to protect from
```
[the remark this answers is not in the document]
requilence: fixed

core/block/editor/dashboard.go
requilence: removed, cause method needs not be done on another level

KirillSto (approved):

PR #2280, opened 2025-03-31
GO-5376 make a migration to install the all objects widget above all

requilence:


core/block/editor/archive.go
```diff
@@ -55,7 +55,7 @@ func (p *Archive) Init(ctx *smartblock.InitContext) (err error) {
 
 func (p *Archive) CreationStateMigration(ctx *smartblock.InitContext) migration.Migration {
 	return migration.Migration{
-		Version: 2,
+		Version: 3,
```
KirillSto, line 58: Why we increment migration version of Archive, not Widget object?

fat-fellow (approved):

PR #2330, opened 2025-04-10
GO-5449 Fallback to layout from bundle info

KirillSto:
https://linear.app/anytype/issue/GO-5449/layout-fallback-on-cold-sync

Fix layout resolution flow:
1. Value from layout detail has highest priority
2. First we should try to resolve layout, and then - inject local details from objectstore
3. If we fail to get details of type object from object store by tyoe id, we need to check if this type is bundled and fallback to layout from bundle info. Key of type must be stored in state
4. Same for templates - we must store 2 keys in store for them: "template" and key for target type

### DIFF
--- a/core/block/editor/smartblock/detailsinject.go
+++ b/core/block/editor/smartblock/detailsinject.go
@@ -219,44 +219,76 @@ func (sb *smartBlock) deriveChatId(s *state.State) error {
 	return nil
 }
 
+// pickResolvedLayout determines the final resolved layout based on following priority:
+// 1. explicitLayoutValue:     explicit layout from object details (if any)
+// 2. typeRecommendedLayout:   layout recommended by the object's type (if any)
+// 3. currentResolvedValue:    the current already-resolved layout (if any)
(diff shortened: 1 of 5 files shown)


core/block/editor/smartblock/detailsinject.go
```diff
@@ -293,26 +277,52 @@ func convertLayoutFromNote(st *state.State, oldLayout, newLayout domain.Value) {
...
 	records, err := sb.objectStore.SpaceIndex(sb.SpaceID()).QueryByIds([]string{typeObjectId})
-	if err != nil || len(records) != 1 {
-		return nil, fmt.Errorf("failed to query object %s: %w", typeObjectId, err)
+	if err == nil && len(records) == 1 {
+		return records[0].Details.Get(bundle.RelationKeyRecommendedLayout), nil
+	}
+
+	log.Debugf("failed to query object type: %v. Trying to get bundled version", err)
+	ot, e := bundle.GetType(s.ObjectTypeKey())
+	if e == nil {
```
deff7, line 323: `err`

PR #2357, opened 2025-04-17
GO-3938 Add includeTime relation - details level

KirillSto:
https://linear.app/anytype/issue/GO-3938/improve-date-and-datetime-filters

- Add includeTime relation and set it to true to system relations with SystemObjectReviser


requilence:
please rename includeTime relation to relationFormatIncludeTime to match the other ones

PR #2423, opened 2025-05-19
GO-5643 Fix panic on space load error

KirillSto:
https://linear.app/anytype/issue/GO-5643/crash-after-migration

We should set loadErr on mandatory objects upload failure


KirillSto:
Closing prior to https://github.com/anyproto/anytype-heart/pull/2424

PR #2425, opened 2025-05-19
GO-4678 Propagate spaceId to backlinks watcher

KirillSto:
https://linear.app/anytype/issue/GO-4678/backlinks-watcher-is-failing-to-resolve-space-id

- Propagate spaceId directly from store
- In case object has link on object from other space, we should parse spaceId from id

### DIFF
--- a/core/domain/id.go
+++ b/core/domain/id.go
@@ -1,6 +1,7 @@
 package domain
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 )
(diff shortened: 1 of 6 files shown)


core/domain/id.go
```diff
@@ -33,3 +36,11 @@ func ParseParticipantId(participantId string) (spaceId string, identity string,
...
 	return fmt.Sprintf("%s.%s", parts[2], parts[3]), parts[4], nil
 }
+
+func ParseLongId(id string) (FullID, error) {
+	if id == "" {
+		return FullID{}, ErrParseLongId
+	}
+	// TODO: support spaceId in long ids
+	return FullID{ObjectID: id}, nil
+}
```
requilence, line 46: what is this func for? I don't see you calling it. But you have impl inside the test file

KirillSto: This function is supposed to support trans-space links. We have not agree on format yet, e.g. <objectId>?spaceId=<spaceId>

requilence (approved):

PR #2448, opened 2025-05-25
GO-5645: REST API Stage 4

jmetrikat:


core/api/filter/parser.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/anyproto/anytype-heart/pkg/lib/bundle"
+	"github.com/anyproto/anytype-heart/pkg/lib/pb/model"
+)
+
+// topLevelAttributes maps JSON field names to internal relation keys
+// These attributes default to "contains" search and don't require a spaceId
+var topLevelAttributes = map[string]string{
+	"name":        bundle.RelationKeyName.String(),
+	"global_name": bundle.RelationKeyGlobalName.String(),
+	"snippet":     bundle.RelationKeySnippet.String(),
```
requilence: Can we use consts for query params

core/api/filter/parser.go
```diff
@@ -0,0 +1,167 @@
...
+// ParseQueryParams parses query parameters into filters
+func (p *Parser) ParseQueryParams(c *gin.Context, spaceId string) (*ParsedFilters, error) {
+	queryParams := c.Request.URL.Query()
+	filters := make([]Filter, 0)
+
+	skipParams := map[string]bool{
+		"offset": true,
+		"limit":  true,
+		"sort":   true,
+		"order":  true,
```
requilence: Here as well

core/api/filter/parser.go
```diff
@@ -0,0 +1,167 @@
...
+// parseFilterValue parses the filter value based on the condition
+func (p *Parser) parseFilterValue(rawValue string, condition model.BlockContentDataviewFilterCondition) (interface{}, error) {
+	decodedValue, err := url.QueryUnescape(rawValue)
+	if err != nil {
+		return nil, util.ErrBadInput(fmt.Sprintf("failed to decode value: %s", err.Error()))
+	}
+
+	switch condition {
+	case model.BlockContentDataviewFilter_Empty, model.BlockContentDataviewFilter_NotEmpty:
+		if decodedValue == "" || decodedValue == "true" || decodedValue == "1" {
```
requilence: Better use strconv.ParseBool as it parses more cases

core/api/model/filter.go
```diff
@@ -0,0 +1,414 @@
...
+}
+
+func (TextFilterItem) isFilterItem()                   {}
+func (f TextFilterItem) GetPropertyKey() string        { return f.PropertyKey }
+func (f TextFilterItem) GetCondition() FilterCondition { return f.Condition }
+func (f TextFilterItem) GetValue() interface{} {
+	if f.Text != nil {
+		return *f.Text
+	}
+	return nil
```
requilence, line 232: Lets use pointer value receivers, so we dont make the struct copy every call

core/api/model/icon.go
```diff
@@ -979,9 +975,9 @@ func (FileIcon) isIcon() {}
...
...ove,person,phone-landscape,phone-portrait,pie-chart,pin,pint,pizza,planet,play-back-circle,play-back,play-circle,play-forward-circle,play-forward,play-skip-back-circle,play-skip-back,play-skip-forward-circle,play-skip-forward,play,podium,power,pricetag,pricetags,print,prism,pulse,push,qr-code,radio-button-off,radio-button-on,radio,rainy,reader,receipt,recording,refresh-circle,refresh,reload-circle,reload,remove-circle,repeat,resize,restaurant,ribbon,rocket,rose,sad,save,scale,scan-circle,scan,school,search-circle,search,send,server,settings,shapes,share-social,share,shield-checkmark,shield-half,shield,shirt,shuffle,skull,snow,sparkles,speedometer,square,star-half,star,stats-chart,stop-circle,stop,stopwatch,storefront,subway,sunny,swap-horizontal,swap-vertical,sync-circle,sync,t.txt,tablet-landscape,tablet-portrait,telescope,tennisball,terminal,text,thermometer,thumbs-down,thumbs-up,thunderstorm,ticket,time,timer,today,toggle,trail-sign,train,transgender,trash-bin,trash,trending-down,trending-up,triangle,trophy,tv,umbrella,unlink,videocam-off,videocam,volume-high,volume-low,volume-medium,volume-mute,volume-off,walk,wallet,warning,watch,water,wifi,wine,woman"` // The name of the icon
```
requilence, line 979: Not sure about storing all possible values as enums. Do we need this for openapi specs generation?

requilence (changes requested):
I will continue review tomorrow. A lot of code but overall very nice and I don't have any critical comments for now

core/api/filter/parser.go
jmetrikat: yes, good point!

core/api/model/icon.go
jmetrikat: that's for openapi spec generation, yep

core/api/model/filter.go
jmetrikat: i'll check

core/api/filter/parser.go
jmetrikat: done!

core/api/model/filter.go
jmetrikat: @requilence unsure about this one, as the structs are small with read-only getters. using value receivers shouldn't be critically expensive - isn't copying even cheaper than dereferencing pointers at such a small size?

PR #2478, opened 2025-06-10
GO-5593 auto widget for space level chat

requilence:
- Automatically initialize space-level chat on SpaceMakeShareable.
- Add an auto-widget for Chat if it was not added before (on SpaceMakeShareable and WorkspaceOpen(withChat = true)).
- After the public release, withChat = true will be removed and will become the default behavior.


core/block/service.go
```diff
@@ -338,6 +343,34 @@ func (s *Service) SpaceInstallBundledObjects(
 	return s.objectCreator.InstallBundledObjects(ctx, spc, sourceObjectIds, false)
 }
 
+// autoInstallSpaceChatWidget automatically installs the chat widget in the space if it is not already installed.
+func (s *Service) autoInstallSpaceChatWidget(ctx context.Context, spc clientspace.Space) error {
```
KirillSto: Maybe we should move this method to core/block/widget.go ?

requilence: done

KirillSto (approved):

PR #2484, opened 2025-06-11
GO-5767 spaceuxtype and chat doesnt come with workspaceopen

requilence:
- make sure we fill the name and spaceUxType to the spaceView at the moment of creation
- cleanup workspace init
- return spacechatId alongside with other system object ids

### DIFF
--- a/core/block/editor/smartblock/smartblock.go
+++ b/core/block/editor/smartblock/smartblock.go
@@ -629,11 +629,6 @@ func (sb *smartBlock) EnabledRelationAsDependentObjects() {
 }
 
 func (sb *smartBlock) Apply(s *state.State, flags ...ApplyFlag) (err error) {
-	if sb.Type() == smartblock.SmartBlockTypeWorkspace {
-		if s.Details().Has(bundle.RelationKeyName) && s.Details().GetString(bundle.RelationKeyName) == "" {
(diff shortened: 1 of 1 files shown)


core/block/editor/smartblock/detailsinject.go
```diff
@@ -223,10 +223,17 @@ func (sb *smartBlock) deriveChatId(s *state.State) error {
...
-	if s.Details() == nil && s.LocalDetails() == nil {
+	if sb.Type() != smartblock.SmartBlockTypePage {
+		layout := s.Details().Get(bundle.RelationKeyLayout)
+		if layout.Ok() {
+			s.SetDetailAndBundledRelation(bundle.RelationKeyResolvedLayout, layout)
+		}
 		return
 	}
 
+	if s.Details() == nil && s.LocalDetails() == nil {
```
mcrakhman, line 234: do we need both checks here?

mcrakhman (approved):

core/block/editor/smartblock/detailsinject.go
requilence: This code already was here, I've just added additional check. TO be honest I'm not sure this check is the best option. I think it never triggers.

PR #2491, opened 2025-06-12
GO-5804 Add includeTypeEdges parameter to ObjectGraph RPC

requilence:
## Description

This PR adds an `includeTypeEdges` parameter to the ObjectGraph RPC endpoint, allowing clients to control whether type relations are included in the graph edges.

## Problem

Currently, the ObjectGraph endpoint always includes type relations (edges where the relation key is "type") in the returned graph. However, in many visualization and analysis scenarios, these type edges can clutter the graph and make it harder to understand the actual relationships between objects. Type relations are already excluded from the `relationsSkipList`, but there are use cases where clients need fine-grained control over their inclusion.

## Solution

Added a boolean `includeTypeEdges` parameter to the ObjectGraph RPC request that allows clients to explicitly control whether type edges should be included in the response.

### 1. Protocol Buffer Changes
- Added `bool includeTypeEdges = 8;` field to `Rpc.Object.Graph.Request` in `pb/protos/commands.proto`

### 2. Core Logic Changes
- Updated `ObjectGraphRequest` struct in `core/block/object/objectgraph/graph.go` to include `IncludeTypeEdges bool` field
- Modified `isRelationShouldBeIncludedAsEdge` function to accept an `includeTypeEdges` parameter
- Added logic to skip type edges when `includeTypeEdges` is false
- Updated all function calls throughout the chain to properly pass the parameter:
  - `ObjectGraph` method
  - `buildGraph` method
  - `appendRelations` method

### 3. Middleware Integration
- Updated the RPC handler in `core/object.go` to pass the `IncludeTypeEdges` field from the protobuf request to the service layer

### 4. Generated Code
- Regenerated all protobuf Go files using `make protos-go` and `make protos-server`

## Technical Details

The change is backward compatible - when `includeTypeEdges` is not specified (default `false`), type edges will not be included. Clients that want type edges must explicitly set `includeTypeEdges: true`.

The filtering logic is implemented in the `isRelationShouldBeIncludedAsEdge` function:
```go
// Skip type edges if includeTypeEdges is false
if \!includeTypeEdges && rel.Key == bundle.RelationKeyType.String() {
    return false
}
```

## Testing Recommendations

1. Test with `includeTypeEdges: false` (default) - verify type edges are not included
2. Test with `includeTypeEdges: true` - verify type edges are included
3. Verify other relation types are not affected by this parameter
4. Test with various graph queries to ensure performance is not impacted


core/block/object/objectgraph/graph.go
```diff
@@ -119,8 +120,22 @@ func (gr *Builder) ObjectGraph(req ObjectGraphRequest) ([]*domain.Details, []*pb
...
+	}
+
+	// Check if relation is in skip list
+	isInSkipList := lo.Contains(relationsSkipList, domain.RelationKey(rel.Key))
+
+	// Special handling for type relation based on includeTypeEdges parameter
+	if rel.Key == bundle.RelationKeyType.String() {
+		// If includeTypeEdges is true, we want to include it regardless of skip list
+		return includeTypeEdges
+	}
```
KirillSto: Let's exclude Type from **relationsSkipList** then. 
If from line 132 could be moved upper

requilence: @KirillSto done

KirillSto (approved):

PR #2493, opened 2025-06-13
GO-5677 use in filter instead of many OR

zarkone:


deff7 (approved):

zarkone:
depends on https://github.com/anyproto/any-store/pull/54/files to be merged

PR #2583, opened 2025-07-25
Go 5214 resolve rellinks from snapshots

KirillSto:
https://linear.app/anytype/issue/GO-5214/resolve-relation-format-from-snapshots-on-import

- introduce importCtx structure
- fetch format of custom relations during import and propagate it to DataObject

### DIFF
--- a/core/block/import/types.go
+++ b/core/block/import/types.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/anyproto/any-sync/app"
+	"github.com/anyproto/any-sync/commonspace/object/tree/treestorage"
 
 	"github.com/anyproto/anytype-heart/core/block/import/common"
 	_ "github.com/anyproto/anytype-heart/core/block/import/markdown"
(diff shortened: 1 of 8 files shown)


core/block/import/types.go
```diff
@@ -31,6 +33,36 @@ type ImportResponse struct {
...
+	convResponse *common.Response
+	error        *common.ConvertError
+
+	oldIDToNew           map[string]string
+	createPayloads       map[string]treestorage.TreeStorageCreatePayload
+	relationKeysToFormat map[domain.RelationKey]int32
+}
+
+func newImportContext(ctx context.Context, req *ImportRequest, resp *common.Response, origin objectorigin.ObjectOrigin, error *common.ConvertError) *importContext {
+	if req == nil || resp == nil {
```
deff7: why that?

KirillSto: I have modified constructor, so it returns error if input pars are not valid

PR #2584, opened 2025-07-27
Refactor usecase validator

KirillSto:
// TODO


util/slice/slice.go
```diff
@@ -320,3 +320,16 @@ func ContainsBySeq[T comparable](seq iter.Seq[T], v1 T) (found bool) {
 	})
 	return
 }
+
+// DeleteOrApplyFunc deletes elements that satisfy del predicate, otherwise apply func
+func DeleteOrApplyFunc[S ~[]E, E any](s S, del func(E) bool, apply func(E) E) S {
```
deff7: maybe call it FilterMapMut?

deff7: and pass a function `func(E) (bool, E)` to it

KirillSto: Basically, that is lo.FilterMap. So I just used that

PR #2613, opened 2025-08-05
GO-2679 Fix file extensions of bookmark icons and covers

KirillSto:
https://linear.app/anytype/issue/GO-2679/maybe-a-critical-bug-string-parsing-of-a-websites-title-can-go-wrong

We used to take extensions from the very URL of bookmark, not the file of icon or cover

### DIFF
--- a/core/block/editor/chatobject/mock_chatobject/mock_StoreObject.go
+++ b/core/block/editor/chatobject/mock_chatobject/mock_StoreObject.go
@@ -1655,21 +1655,31 @@ func (_c *MockStoreObject_MarkMessagesAsUnread_Call) RunAndReturn(run func(conte
 }
 
 // MarkReadMessages provides a mock function with given fields: ctx, req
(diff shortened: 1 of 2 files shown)


core/block/bookmark/bookmark_service.go
```diff
@@ -373,21 +373,21 @@ func (s *service) fetcher(spaceID string, blockID string, params bookmark.FetchP
...
 	}
-	var urlFileExt string
-	lastPath := strings.Split(u.Path, "/")
-	if len(lastPath) > 0 {
-		urlFileExt = filepath.Ext(lastPath[len(lastPath)-1])
+	fu, err := uri.ParseURI(fileUrl)
+	if err != nil {
+		return ""
 	}
+	urlFileExt := filepath.Ext(fu.Path)
```
deff7: path.Ext

KirillSto: Done!

PR #2619, opened 2025-08-06
Fix setup progress in import

KirillSto:
Fix progress setup in import:
- in case it was nil in request, we need to init it and enregister in process service
- otherwise do nothing, as progress is already set up

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@ module github.com/anyproto/anytype-heart
 
 go 1.23.10
 
-toolchain go1.24.0
+toolchain go1.23.11
 
 require (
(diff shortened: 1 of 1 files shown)


core/block/import/processor.go
```diff
@@ -106,6 +104,11 @@ func (p *importProcessor) setupProgressBar() {
 	}
 
 	p.request.Progress = progress
+
+	if p.deps.blockService != nil && !p.request.GetNoProgress() {
```
deff7: how can block service be nil?

KirillSto: Cannot be, some sort of relict. Removed it

PR #2620, opened 2025-08-07
GO-6044 Fix relation fetch on dataview operations

KirillSto:
https://linear.app/anytype/issue/GO-6044/blockdataviewrelationset-code-object-not-found-in-space-index

We should return Relation data from store even if relation object is archived or deleted

### DIFF
--- a/core/block/editor/dataview/dataview.go
+++ b/core/block/editor/dataview/dataview.go
@@ -155,9 +155,9 @@ func (d *sdataview) SetRelations(ctx session.Context, blockId string, relationKe
 	}
 	var links []*model.RelationLink
 	for _, key := range relationKeys {
-		relation, err2 := d.objectStore.FetchRelationByKey(key)
-		if err2 != nil {
-			return err2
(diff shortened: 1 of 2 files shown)


core/block/editor/dataview/dataview.go
```diff
@@ -155,9 +155,9 @@ func (d *sdataview) SetRelations(ctx session.Context, blockId string, relationKe
 	}
 	var links []*model.RelationLink
 	for _, key := range relationKeys {
-		relation, err2 := d.objectStore.FetchRelationByKey(key)
-		if err2 != nil {
-			return err2
+		relation, err := d.objectStore.FetchRelationByKey(key)
+		if err != nil {
+			return fmt.Errorf("failed to find relation '%s': %w", key, err)
```
deff7, line 160: Will this error be propagated to client or to log?

KirillSto: To client

deff7 (approved):

PR #2647, opened 2025-08-18
GO-6131 Add accountKey authentication support

requilence:
## Summary
Support authentication using accountKey instead of mnemonic. It will be used for headless mw

## Changes
- Remove mnemonic field from Service struct, use only derivedKeys
- Add accountKey parameter to wallet.recover for bot authentication
- Move key derivation to wallet.create/recover
- Account.select/create now use pre-derived keys

Backward compatible - existing mnemonic authentication unchanged.

Descendent PRs:
https://github.com/anyproto/any-sync/pull/510
https://github.com/anyproto/go-slip10/pull/3

### DIFF
--- a/core/application/account_create.go
+++ b/core/application/account_create.go
@@ -34,18 +34,18 @@ func (s *Service) AccountCreate(ctx context.Context, req *pb.RpcAccountCreateReq
 
 	s.requireClientWithVersion()
 
-	derivationResultPtr, err := s.getDerivationResult()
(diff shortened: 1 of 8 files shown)


jmetrikat (changes requested):
please address comments: base58 (prev approach) vs. base64 (now); and update PR description to avoid confusion

jmetrikat (approved):

PR #2648, opened 2025-08-19
GO-4284 Introduce relation format fetcher

KirillSto:
https://linear.app/anytype/issue/GO-4284/get-rid-of-relation-links

We are planning to completely remove relationLinks from state. To accomplish this the process should be divided into two stages:
1. Implement middleware where all services do not depend on relationLinks, but still write them in the State, so changes are generated
2. Get rid of relationLinks completely.

If we mark current version as v0, version with changes of stage 1 as v1 and stage 2 version as v2, we can state that:
- objects created/changed on version v2 will be shown as broken in v0
- v1 will be compatible both with v0 and v2 in both directions

This PR includes changes for stage 1:
- RelationFormatFetcher is introduced. It handles lazy space-based relation subscriptions, so relation format could be fetched from this service instead of RelationLinks in state. Fetcher is included in all services that need to get dependencies of object
- Details-related methods of state is moved to separate file
- remove ReplaceLink method from basic
- remove HasRelation and Relations methods from Smartblock
- replace GetRelationLinks method of Doc with AllRelationKeys
- refactor export tests, as they deeply depend on details of exported objects
- fix bug on collection change build
- remove RelationLinks from exported Snapshot
- exclude relationLinks from ObjectShow/Open and HistoryX commands


core/subscription/objectsubscription/objsubscription.go
```diff
@@ -143,6 +153,17 @@ func (o *ObjectSubscription[T]) Get(id string) (T, bool) {
 	return entry, ok
 }
 
+func (o *ObjectSubscription[T]) GetByKey(key string) (T, bool) {
+	o.mx.Lock()
+	defer o.mx.Unlock()
+	id, ok := o.keyToId[key]
+	if !ok {
+		return *new(T), false
```
deff7: It's better to use default value:

```
var defValue T
return defValue, false
```

core/block/editor/state/details.go
```diff
@@ -0,0 +1,237 @@
...
+				keys = append(keys, key)
+			}
+		}
+	}
+	return keys
+}
+
+func (s *State) iterateKeys() iter.Seq[domain.RelationKey] {
+	return func(yield func(domain.RelationKey) bool) {
+		for _, seq := range []iter.Seq[domain.RelationKey]{
```
deff7: I propose you to just run two loops instead of creating the slice of iterators to avoid an unnecessary memory allocation. Golang probably might optimize such a construction, but I'd better not to rely on that.

core/block/editor/smartblock/smartblock.go
```diff
@@ -1088,12 +1078,12 @@ func hasDepIds(relations pbtypes.RelationLinks, act *undo.Action) bool {
 		}
 
 		for k, after := range act.Details.After.Iterate() {
-			rel := relations.Get(string(k))
-			if rel != nil && (rel.Format == model.RelationFormat_status ||
-				rel.Format == model.RelationFormat_tag ||
-				rel.Format == model.RelationFormat_object ||
-				rel.Format == model.RelationFormat_file ||
-				isCoverId(rel)) {
+			format, err := sb.formatFetcher.GetRelationFormatByKey(sb.SpaceID(), k)
```
deff7, line 1081: Maybe add

```
if err != nil {
  continue
  }
```
?

core/subscription/objectsubscription/objsubscription.go
KirillSto: Done

core/block/editor/state/details.go
KirillSto: Done

core/block/editor/smartblock/smartblock.go
KirillSto: Done

PR #2666, opened 2025-08-29
GO-5800 Support sort by object relations

KirillSto:
https://linear.app/anytype/issue/GO-5800/support-sorting-by-object-relations

Right now any sort upon relations with format tag/status/object/file use object ids to calculate order of objects in subscription.
To support ordering by object names and orderIds of tags following changes were made:
- dependency service collects infomation on object sorts for subscriptions and triggers subscription reorder in case objects of dependent subscription change its order (orderId/name) and these objects influence sort
- **orderMap** is introduced. This struct handles orders of dependent objects and helps to compare objects between each other
- **UpdateOrderMap** method is added to **database.Order** interface. It is triggered on dependant objects change and keeps orders relevant all the time

### DIFF
--- a/pkg/lib/database/anystore.go
+++ b/pkg/lib/database/anystore.go
@@ -210,9 +210,10 @@ func (s objectSort) AppendKey(tuple anyenc.Tuple, v *anyenc.Value) anyenc.Tuple 
 
 	for _, key := range s.sortKeysBuffer {
 		if len(key) == 0 {
-			if s.nulls == model.BlockContentDataviewSort_Start {
+			switch s.nulls {
+			case model.BlockContentDataviewSort_Start:
 				tuple = tuple.Append(s.arena.NewNull())
-			} else if s.nulls == model.BlockContentDataviewSort_End {
+			case model.BlockContentDataviewSort_End:
 				tuple = tuple.AppendInverted(s.arena.NewNull())
 			}
 			continue


core/subscription/sorted.go
```diff
@@ -262,6 +260,40 @@ func (s *sortedSub) onEntryChange(ctx *opCtx, e *entry) (noChange bool) {
...
+		return
+	}
+
+	entries := s.getActiveEntries()
+	s.skl.Init()
+	for _, e := range entries {
+		s.skl.Set(e, nil)
+	}
+
+	defer s.diff.reset()
```
deff7, line 229: It's a large chunk of complex duplicated code from onChange, let's extract it to function

core/subscription/sorted.go
```diff
@@ -262,6 +260,40 @@ func (s *sortedSub) onEntryChange(ctx *opCtx, e *entry) (noChange bool) {
 	panic("subscription: check algo")
 }
 
+func (s *sortedSub) reorder(ctx *opCtx, depDetails []*domain.Details) {
+	if !s.order.UpdateOrderMap(depDetails) {
+		return
+	}
+
+	entries := s.getActiveEntries()
```
deff7: you have to use all entries, not only the active

core/subscription/dep.go
```diff
@@ -90,13 +111,56 @@ func (ds *dependencyService) depEntriesByEntries(ctx *opCtx, depIds []string) (d
 	return
 }
 
+func (ds *dependencyService) enregisterObjectSorts(subId string, sorts []database.SortRequest) {
```
deff7: maybe use verb `register`? It's much more common

core/subscription/dep.go
```diff
@@ -120,11 +184,53 @@ func (ds *dependencyService) isRelationObject(spaceId string, key domain.Relatio
...
+
+// depSubKeys returns keys that will be analyzed in objects filtered by dependent subscription
+// TODO: maybe we need to exclude some keys from initial list (lastModifiedDate, syncDate)
+func (ds *dependencyService) depSubKeys(subId string, keys []domain.RelationKey) []domain.RelationKey {
+	sorts, found := ds.sorts[subId]
+	if !found {
+		return keys
+	}
+	for _, sort := range sorts {
+		keys = append(keys, sort.orderKey())
```
deff7: You can also use `if !slices.Contains` here, lo.Uniq is a much heavier operation

core/subscription/dep.go
```diff
@@ -90,13 +111,56 @@ func (ds *dependencyService) depEntriesByEntries(ctx *opCtx, depIds []string) (d
...
+		if !ds.isRelationObject(sort.RelationKey) {
+			continue
+		}
+		sortRelations = append(sortRelations, sortKey{
+			key:   sort.RelationKey,
+			isTag: sort.Format == model.RelationFormat_tag || sort.Format == model.RelationFormat_status,
+		})
+	}
+
+	if len(sortRelations) != 0 {
```
deff7: len > 0 by code style

core/subscription/dep.go
```diff
@@ -13,39 +16,57 @@ func newDependencyService(s *spaceSubscriptions) *dependencyService {
...
+			_, isSortKey := ds.sorts.getSortKey(subId, k)
 			for _, depId := range e.data.WrapToStringList(k) {
-				if depId != "" && slice.FindPos(depIds, depId) == -1 && depId != e.id {
-					depIds = append(depIds, depId)
+				if depId != "" {
+					if slice.FindPos(depIds, depId) == -1 && depId != e.id {
+						depIds = append(depIds, depId)
+					}
+					if isSortKey {
+						if ds.depOrderObjects[depId] == nil {
```
deff7: maybe extract it to a function?

core/subscription/dep.go
```diff
@@ -120,11 +184,53 @@ func (ds *dependencyService) isRelationObject(spaceId string, key domain.Relatio
...
+func (k sortKey) orderKey() domain.RelationKey {
+	if k.isTag {
+		return bundle.RelationKeyOrderId
+	}
+	return bundle.RelationKeyName
+}
+
+type sortsMap map[string][]sortKey // subId -> sortRelationKeys
+
+func (m sortsMap) getSortKey(subId string, key domain.RelationKey) (sortKey, bool) {
```
deff7: maybe add a symmetrical function `setSortKeys`?

core/subscription/dep.go
```diff
@@ -120,11 +184,53 @@ func (ds *dependencyService) isRelationObject(spaceId string, key domain.Relatio
...
+}
+
+func (k sortKey) orderKey() domain.RelationKey {
+	if k.isTag {
+		return bundle.RelationKeyOrderId
+	}
+	return bundle.RelationKeyName
+}
+
+type sortsMap map[string][]sortKey // subId -> sortRelationKeys
```
deff7: Can we benefit from using a map for keys? It seems so

core/subscription/sorted.go
KirillSto: Moved to separate method onSklChange

core/subscription/sorted.go
KirillSto: Thank you! Now first we retrieve all entries by iteration over SkipList elements from Front, only then reset Skiplist

core/subscription/dep.go
KirillSto: Done!

core/subscription/dep.go
KirillSto: Done!

core/subscription/dep.go
KirillSto: SortsMap was refactored

core/subscription/dep.go
KirillSto: Done!

core/subscription/dep.go
KirillSto: SortsMap was refactored. Please take a look, maybe it is not necessary yet

core/subscription/dep.go
KirillSto: Done!

core/subscription/dep.go
```diff
@@ -120,11 +178,72 @@ func (ds *dependencyService) isRelationObject(spaceId string, key domain.Relatio
...
+			hasOrderId = true
+		} else {
+			if hasOrderId {
+				return []domain.RelationKey{bundle.RelationKeyOrderId, bundle.RelationKeyName}
+			}
+			hasName = true
+		}
+	}
+
+	if hasName {
```
deff7, line 239: It will be simpler if you'll add case `hasName && hasOrderId` here

core/subscription/sorted.go
```diff
@@ -200,66 +270,30 @@ func (s *sortedSub) onChange(ctx *opCtx) {
...
 	}
-	// change
-	if curInSet && newInSet {
-		s.skl.Remove(curr)
-		s.skl.Set(e, nil)
-		e.SetSub(s.id, false, false)
-		return
+	s.afterId = ""
+	if err := s.setBeforeAndAfterElements(); err != nil {
+		panic(fmt.Errorf("failed to set before and after elements: %w", err))
```
deff7: why to panic here?

pkg/lib/database/order.go
```diff
@@ -17,17 +18,116 @@ import (
...
+func NewOrderMap(data map[string]*domain.Details) *OrderMap {
+	return &OrderMap{data: data}
+}
+
+func (m *OrderMap) BuildOrderByKey(key domain.RelationKey, ids ...string) string {
+	if m == nil || len(m.data) == 0 {
+		return ""
+	}
+
+	var result string
```
deff7: Maybe use `strings.Builder` here?

pkg/lib/database/order.go
```diff
@@ -17,17 +18,116 @@ import (
...
+	for _, id := range ids {
+		if details, ok := m.data[id]; ok {
+			result += details.GetString(key)
```
deff7: This code implies that arrays [1,2,3] and [2,3,1] are not equal. Does it conform with business rules? Or maybe we must sort `sets`?

pkg/lib/database/order.go
deff7: Even with `strings.Builder` this code will be not effective as it could be. Consider using some kind of reusable buffer. Notice that this function is called inside a sorting procedure which has `O(n*log n)` worst case scenario .

pkg/lib/database/order.go
```diff
@@ -17,17 +18,116 @@ import (
...
+		return
+	}
+
+	records, err := store.Query(Query{Filters: []FilterRequest{{
+		RelationKey: bundle.RelationKeyId,
+		Condition:   model.BlockContentDataviewFilter_In,
+		Value:       domain.StringList(idsToSet),
+	}}})
+	if err != nil {
+		log.Errorf("failed to query records to update orders: %v", err)
```
deff7: Consider returning the error and log it on a some upper level. Function will become slightly cleaner

core/subscription/dep.go
KirillSto: Done

core/subscription/sorted.go
KirillSto: Replaced with log

pkg/lib/database/order.go
KirillSto: Done

pkg/lib/database/order.go
KirillSto: It is stated by products to sort by first value, so it works like that already

pkg/lib/database/order.go
KirillSto: Done

pkg/lib/database/filter.go
```diff
@@ -803,20 +803,58 @@ func (exIn *FilterOptionsEqual) String() string {
...
+// objectsToMap collects names of objects that present in details of any objects as a value of detail with key=key
+func objectsToMap(key domain.RelationKey, store ObjectStore) map[string]*domain.Details {
+	names := make(map[string]*domain.Details)
+	targetIds := make([]string, 0)
+
+	err := store.QueryIterate(Query{Filters: []FilterRequest{{
+		RelationKey: key,
+		Condition:   model.BlockContentDataviewFilter_NotEmpty,
+	}}}, func(details *domain.Details) {
+		targetIds = append(targetIds, details.GetStringList(key)...)
```
deff7: Maybe use a map here? Then build a slice from it

pkg/lib/database/order.go
deff7: It's better to use shared buffer here

pkg/lib/database/filter.go
KirillSto: Done!

PR #2678, opened 2025-09-09
GO-5809 Prohibit to remove system blocks and details

KirillSto:
https://linear.app/anytype/issue/GO-5809/do-not-allow-to-remove-system-blocksdetails

We should not allow to remove system blocks (header, title, featuredRelations) and internal required details from state by client commands.
Plus **workspaceId** relation is deleted and **myParticipantStatus** relation is moved from internal to sustem relations list

### DIFF
--- a/core/block/editor/basic/details.go
+++ b/core/block/editor/basic/details.go
@@ -24,15 +24,14 @@ import (
 var log = logging.Logger("anytype-mw-editor-basic")
 
 func (bs *basic) SetDetails(ctx session.Context, details []domain.Detail, showEvent bool) (err error) {
-	err = bs.UpdateDetails(ctx, func(current *domain.Details) (*domain.Details, error) {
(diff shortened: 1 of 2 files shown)


core/block/editor/basic/details.go
```diff
@@ -24,72 +24,65 @@ import (
...
-	flags := internalflag.NewFromState(s.ParentState())
-	flags.Remove(model.InternalFlag_editorDeleteEmpty)
-	flags.AddToState(s)
+	err = bs.UpdateDetails(ctx, func(current *domain.Details) (*domain.Details, error) {
+		return applyDetailUpdates(current, details), nil
+	})
 
-	if err = bs.Apply(s, smartblock.NoRestrictions, smartblock.KeepInternalFlags); err != nil {
-		return nil, err
+	if err == nil {
```
deff7: It's better to do early return here and continue with a positive scenario

if err != nil { return

KirillSto: Done!

PR #2720, opened 2025-09-23
GO-6278 Make new types first

KirillSto:
https://linear.app/anytype/issue/GO-6278/ensure-that-widget-for-the-newly-created-type-is-positioned-at-top-of

We should set current time to lastUsedDate to sort new type as first, if no custom order is set

### DIFF
--- a/core/block/object/objectcreator/object_type.go
+++ b/core/block/object/objectcreator/object_type.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/anyproto/anytype-heart/core/block/editor/order"
+	"github.com/anyproto/anytype-heart/core/block/editor/smartblock"
 	"github.com/anyproto/anytype-heart/core/block/editor/state"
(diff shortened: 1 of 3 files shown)


core/block/object/objectcreator/object_type.go
```diff
@@ -61,10 +63,13 @@ func (s *service) createObjectType(ctx context.Context, space clientspace.Space,
 		return "", nil, fmt.Errorf("create smartblock from state: %w", err)
 	}
 
-	installingObjectTypeKey := domain.TypeKey(uniqueKey.InternalKey())
-	err = s.createTemplatesForObjectType(space, installingObjectTypeKey)
-	if err != nil {
-		log.With("spaceID", space.Id(), "objectTypeKey", installingObjectTypeKey).Errorf("error while installing templates: %s", err)
+	typeKey := domain.TypeKey(uniqueKey.InternalKey())
+	if err = s.setOrderId(ctx, space, id); err != nil {
```
deff7: maybe set orderId in `createState`?

PR #2737, opened 2025-10-02
GO-6278 Set orderId to all new custom types

KirillSto:
https://linear.app/anytype/issue/GO-6278/ensure-that-widget-for-the-newly-created-type-is-positioned-at-top-of

Ser orderId to all new custom types to set new types at list top

### DIFF
--- a/core/block/object/objectcreator/object_type.go
+++ b/core/block/object/objectcreator/object_type.go
@@ -184,7 +184,7 @@ func (s *service) setOrderId(details *domain.Details, spc clientspace.Space) {
 	}
 
 	var smallestOrderId string
-	if len(records) != 0 {
+	if len(records) > 0 {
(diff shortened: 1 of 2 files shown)


core/block/object/objectcreator/object_type.go
```diff
@@ -179,18 +179,14 @@ func (s *service) setOrderId(details *domain.Details, spc clientspace.Space) err
...
 
-	if len(records) == 0 {
-		return nil
-	}
-
-	smallestOrderId := records[0].Details.GetString(bundle.RelationKeyOrderId)
-	if smallestOrderId == "" {
-		return nil
+	var smallestOrderId string
+	if len(records) != 0 {
```
deff7: len > 0

KirillSto: Fixed!

PR #2738, opened 2025-10-02
GO-6187 Prohibit to move block to its child

KirillSto:
https://linear.app/anytype/issue/GO-6187/the-nested-block-disappears-when-it-is-moved-inside-itself


core/block/editor/basic/basic.go
```diff
@@ -274,6 +275,9 @@ func (bs *basic) Move(srcState, destState *state.State, targetBlockId string, po
 			if replacementCandidate == nil {
 				replacementCandidate = srcState.Get(id)
 			}
+			if slices.Contains(b.Model().ChildrenIds, targetBlockId) {
+				return fmt.Errorf("can not move block to its child")
```
deff7: can't or cannot

PR #2739, opened 2025-10-06
GO-6279 sync sqlite on inactivity

requilence:
- flush objectstore/spacestore dbs on app state change

### DIFF
--- a/pkg/lib/datastore/anystoreprovider/provider.go
+++ b/pkg/lib/datastore/anystoreprovider/provider.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	anystore "github.com/anyproto/any-store"
@@ -15,6 +16,7 @@ import (
 	"go.uber.org/zap"
 	"zombiezen.com/go/sqlite"
 
+	"github.com/anyproto/anytype-heart/core/domain"
 	"github.com/anyproto/anytype-heart/core/wallet"
 	"github.com/anyproto/anytype-heart/pkg/lib/localstore/objectstore/anystorehelper"
 	"github.com/anyproto/anytype-heart/pkg/lib/logging"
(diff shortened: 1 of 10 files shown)


cheggaaa (approved):

pkg/lib/datastore/anystoreprovider/provider.go
```diff
@@ -161,20 +155,38 @@ func (s *provider) Run(ctx context.Context) error {
...
-		return nil, fmt.Errorf("open database: %w", err)
+		code, isCorrupted := anystorehelper.IsCorruptedError(err)
+		getLogger(err, code).Error("failed to open anystore, reinit db")
+		if isCorrupted {
+			removeErr := anystorehelper.RemoveSqliteFiles(path)
+			if removeErr != nil {
+				log.Error("failed to remove sqlite files", zap.Error(removeErr))
+				return nil, removeErr
+			}
+			db, err = anystore.Open(ctx, path, config)
```
deff7, line 179: db is not used after an assignment

pkg/lib/localstore/objectstore/anystorehelper/helper.go
```diff
@@ -17,86 +15,30 @@ import (
...
+	return
+}
 
-	if !runQuickCheck {
-		return
+func RemoveSqliteFiles(dbPath string) error {
+	paths := []string{
+		dbPath,
+		dbPath + "-shm",
+		dbPath + "-wal",
```
deff7, line 40: why delete wal file? Won't this lead to data loss?

requilence: We use this func only for objectstore/crdt stores. In case db is clearly corrupted we remove it to reindex

pkg/lib/datastore/anystoreprovider/provider.go
requilence: fixed, thanks!

PR #2747, opened 2025-10-07
Fix date relation filters

KirillSto:
Fix filters for date relations in case time should not be included. In this case we should modify Equal filters to pair of filters that checks if relation value lays in day range

### DIFF
--- a/pkg/lib/database/quickoptions.go
+++ b/pkg/lib/database/quickoptions.go
@@ -8,8 +8,12 @@ import (
 	timeutil "github.com/anyproto/anytype-heart/util/time"
 )
 
-func transformQuickOption(protoFilter FilterRequest) []FilterRequest {
-	if protoFilter.QuickOption == 0 || protoFilter.Format != model.RelationFormat_date {
+func transformDateFilter(protoFilter FilterRequest) []FilterRequest {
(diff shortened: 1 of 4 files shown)


pkg/lib/database/quickoptions.go
```diff
@@ -8,8 +8,12 @@ import (
...
 )
 
-func transformQuickOption(protoFilter FilterRequest) []FilterRequest {
-	if protoFilter.QuickOption == 0 || protoFilter.Format != model.RelationFormat_date {
+func transformDateFilter(protoFilter FilterRequest) []FilterRequest {
+	if protoFilter.Format != model.RelationFormat_date {
+		return []FilterRequest{protoFilter}
+	}
+	// we should transform date filters in case QuickOption is selected or IncludeTime = false
+	if protoFilter.QuickOption == 0 && protoFilter.IncludeTime {
```
deff7, line 16: Did you mean `quickOption == 0 || includeTime`?

KirillSto: No, in case quickOption != 0 OR NOT includeTime we should proceed with filters processing, so condition is correct

deff7 (approved):

PR #2782, opened 2025-10-29
GO-5222 GO-2293 Fix sortings of empty values

KirillSto:
https://linear.app/anytype/issue/GO-5552/show-empty-values-on-bottom-doesnt-work-correctly
- PR includes refactoring of KeyOrder and OrderMap logic, so code is much clearer
- empty values handling is covered with tests
- support of collate of Names and Snippets is added

https://linear.app/anytype/issue/GO-2293/sorting-issue-with-status-relations-filled-with-numbers
- fix of bug with orderings by Status relation: orderIds were sorted using collator

### DIFF
--- a/pkg/lib/database/ordermap.go
+++ b/pkg/lib/database/ordermap.go
@@ -182,6 +182,7 @@ func (m *OrderMap) collectOptionOrders(key domain.RelationKey) {
 
 func (m *OrderMap) getName(details *domain.Details) string {
 	name := details.GetString(bundle.RelationKeyName)
+	// nolint:gosec
 	if name == "" && model.ObjectTypeLayout(details.GetInt64(bundle.RelationKeyResolvedLayout)) == model.ObjectType_note {
 		name = details.GetString(bundle.RelationKeySnippet)
 	}


core/block/object/objectcreator/relation_option.go
```diff
@@ -52,3 +55,40 @@ func (s *service) createRelationOption(ctx context.Context, space clientspace.Sp
...
+		Sorts: []database.SortRequest{{
+			RelationKey: bundle.RelationKeyOrderId,
+			Type:        model.BlockContentDataviewSort_Asc,
+			NoCollate:   true,
+		}},
+		Limit: 1,
+	})
+
+	if err != nil {
+		log.With("spaceID", spc.Id()).Errorf("failed to query object types with orders to set orderId to new type: %v", err)
```
deff7: As a rule of a thumb you should defer logging as long as you can, it keeps code purer (in a sense of side-effects)

KirillSto: I have moved logging to a calling method, but it is still in the same file

pkg/lib/database/order.go
```diff
@@ -259,230 +157,229 @@ func (ko *KeyOrder) AnystoreSort() query.Sort {
...
-	if ko.EmptyPlacement == model.BlockContentDataviewSort_NotSpecified {
+func (ko *keyOrder) compareStrings(av domain.Value, bv domain.Value) int {
+	aStr, okA := av.TryString()
+	bStr, okB := bv.TryString()
+
+	aEmpty := !okA || aStr == ""
+	bEmpty := !okB || bStr == ""
+
+	comp, ok := ko.tryCompareEmptyValues(aEmpty, bEmpty)
+	if ok {
```
deff7, line 236: you should consider sort direction here (asc/desc)

deff7: maybe flip direction of sorting up the call stack? In `Compare` function, for example

KirillSto: There is no necessity in sort direction consideration in case of empty policy presented in request, because there is no matter in what direction sequence must be sorted. In case Start is selected all empties must be less than any right value, in case of End - greater

PR #2796, opened 2025-11-12
GO-6424 Invites: Add spaceIconOption and creatorIcon

deff7:


pb/protos/commands.proto
```diff
@@ -320,6 +320,9 @@ message Rpc {
                 string spaceName = 3;
                 string spaceIconCid = 4;
                 string creatorName = 5;
+                string creatorIconCid = 10;
```
KirillSto, line 323: Why order of new proto fields is so?
Seems it is alphabetical, but initial list was not sorted by names

pkg/lib/pb/model/protos/models.proto
```diff
@@ -1225,12 +1225,18 @@ message Invite {
 message InvitePayload {
     string creatorIdentity = 1;
     string creatorName = 2;
+    string creatorIconCid = 12;
```
KirillSto, line 1228: Same here

pb/protos/commands.proto
deff7: name then icon

KirillSto (approved):

PR #2809, opened 2025-11-18
GO-6454 Introduce change type

KirillSto:
https://linear.app/anytype/issue/GO-6454/do-not-update-lastmodified-on-auto-changes

- Introduce changeType field in Change model, that handles type of changes applied to object.
- This value will help us to distinguish changes that were made explicitly by user from all other.
- Include changeType check on LastModified info setting logic flow

### DIFF
--- a/core/domain/types.go
+++ b/core/domain/types.go
@@ -75,3 +75,7 @@ func (c ChangeType) String() string {
 		return "Unknown"
 	}
 }
+
+func (c ChangeType) Raw() uint32 {
+	return uint32(c)
+}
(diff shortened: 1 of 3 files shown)


core/domain/types.go
```diff
@@ -33,3 +33,45 @@ func MarshalTypeKeys(typeKeys []TypeKey) []string {
 	}
 	return res
 }
+
+type ChangeType uint32
+
+const (
+	UserChange ChangeType = iota
```
deff7: please, add prefix ChangeType to all constants

KirillSto: Done!

PR #2832, opened 2025-12-01
GO-6524 inbox send fault tolerance

zarkone:
saves onetoone invite sent status to inbox and periodically tries to retry to send those which were not sent

### DIFF
--- a/core/onetoone/service.go
+++ b/core/onetoone/service.go
@@ -144,7 +144,7 @@ func (s *onetoone) processOneToOneInvite(packet *coordinatorproto.InboxPacket) (
 		OneToOneIdentity: identityProfileWithKey.IdentityProfile.Identity,
 	}
 
-	err = s.blockService.CreateOneToOneFromInbox(context.TODO(), spaceDescription, &identityProfileWithKey)
+	err = s.blockService.CreateOneToOneFromInbox(context.Background(), spaceDescription, &identityProfileWithKey)
 	if err != nil {
 		log.Error("create onetoone space from inbox", zap.Error(err))
 		return fmt.Errorf("processOneToOneInvite error: %s", err.Error())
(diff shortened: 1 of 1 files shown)


space/spaceinfo/spaceinfo.go
```diff
@@ -119,12 +119,26 @@ const (
 	AccessTypeShared   = AccessType(model.SpaceAccessType_Shared)
 )
 
+type OneToOneInboxSentStatus int32
+
+const (
+	OneToOneInboxSentStatus_None OneToOneInboxSentStatus = 0
```
deff7: we don't use underscores in hand-written code

core/onetoone/service.go
```diff
@@ -139,3 +184,58 @@ func (s *onetoone) SendOneToOneInvite(ctx context.Context, receiverIdentity stri
...
+}
+
+// inboxResend runs periodically with sendInviteInterval, checks space views with invites which are not
+// being sent yet (e.g. freshly created 1-1 spaces, failed inbox invites due to lack of network etc)
+// and resend them.
+//
+// In case of success it sets space view inbox status to success.
+func (s *onetoone) inboxResend(ctx context.Context) (err error) {
+	records, err := s.objectStore.SpaceIndex(s.techSpace.TechSpaceId()).Query(database.Query{
+		Filters: []database.FilterRequest{
```
deff7, line 201: Maybe add filter for uxtype=oneToOne?

core/onetoone/service.go
```diff
@@ -139,3 +184,58 @@ func (s *onetoone) SendOneToOneInvite(ctx context.Context, receiverIdentity stri
...
+		return
+	}
+	if len(records) == 0 {
+		log.Info("onetoone: inboxResend: no inbox invites to send, return")
+		return
+	}
+
+	for _, record := range records {
+		bobIdentity := record.Details.GetString(bundle.RelationKeyOneToOneIdentity)
+		err := s.SendOneToOneInvite(context.TODO(), bobIdentity)
```
deff7: please, add a background context with cancellation to the service and use it here and in other places where context.TODO is used.

deff7: maybe send and handle result in goroutines? But I don't think it's really necessary as it's a background job and no rush to do it

PR #2848, opened 2025-12-05
GO-6503 Trim description on bk fetch

KirillSto:
https://linear.app/anytype/issue/GO-6503/editing-a-bookmark-object-causes-the-entire-application-interface-to

We should trim bookmark description to 200 runes in all scenarios


util/linkpreview/linkpreview.go
```diff
@@ -482,3 +480,10 @@ func replaceGenericTitle(preview *model.LinkPreview, htmlContent []byte) {
 		}
 	}
 }
+
+func trimDescription(description string) string {
```
deff7: textutil.TruncateEllipsized

KirillSto: Done :saluting_face:

PR #2867, opened 2025-12-18
GO-6590 set onetoone name only if not empty

zarkone:
fixes coldsync race


cheggaaa (approved):

core/block/editor/workspaces.go
```diff
@@ -132,6 +132,10 @@ func (w *Workspaces) subscribeForOneToOneProfile(state *state.State) {
 }
 
 func (w *Workspaces) updateOneToOneInfo(details *domain.Details) {
+	if details.Get(bundle.RelationKeyName).String() == "" {
```
deff7, line 135: maybe instead don't add name to `toSave` map? In case some user doesn't have a name, but has an icon

deff7 (approved):

PR #2880, opened 2026-01-06
core/api: allow type object ids in type filters

stevelr:
- accept []string inputs for multi-select and object properties\n
- allow objectType layout for type relation validation 
- add unit test for type object validation

---

- [X] I understand that contributing to this repository will require me to agree with the [CLA](https://github.com/anyproto/.github/blob/main/docs/CLA.md)

---

### Description

Type filters (type[in]=...) rejected valid type object ids because object validation only allowed “object/member” layouts and didn’t handle ObjectType_objectType.

### What type of PR is this? (check all applicable)

- [ ] 🍕 Feature
- [X] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] 🔥 Performance Improvements
- [X] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Related Tickets & Documents

This is one of three bug fix PRs that arose out of testing object queries and filters in a rust SDK for Anytype (the problems can be reproduced with curl so they are not rust-related).

[2881](https://github.com/anyproto/anytype-heart/pull/2881) Filter bool numbers
[2880](https://github.com/anyproto/anytype-heart/pull/2880) core/api: allow type object ids in type filters
[2882](https://github.com/anyproto/anytype-heart/pull/2882) core/api/filter: allow select tag ids

To keep the changes simpler and easier to review, they are separate PRs. I have another [branch](https://github.com/stevelr/anytype-heart/tree/testing)  that combines them for testing - let me know if you'd rather I submit that as a PR instead.

### Mobile & Desktop Screenshots/Recordings

n/a

### Added tests?

- [X] 👍 yes
- [ ] 🙅 no, because they aren't needed
- [ ] 🙋 no, because I need help

### Added to documentation?

- [X] 🙅 no documentation needed

### [optional] Are there any post-deployment tasks we need to perform?

N/A

PR #2881, opened 2026-01-06
Filter bool number

### LINKED ISSUE #2879: property_key=value filters in get_objects rest queries do not support boolean or number values

### Current Behavior

in get_objects query, a boolean condition property_key=true fails with the error "property_key" must be a boolean. (fails with any of value 'true', 'false', 1, 0)

A number filter property_key=value fails with "property" must be a number.

### Expected Behavior

Expected the filter to work as documented.

### Steps To Reproduce

Here's a bash script to reproduce the problem

```bash

#SPACEID="space-id"
#TOKEN="api-key"

URL="http://127.0.0.1:31009"

CURL=("curl" "--silent"
  "-H" "Authorization: Bearer $TOKEN"
  "-H" "anytype-version: 2025-11-08"
  "-H" "accept: */*"
)

"${CURL[@]}" "$URL/v1/spaces/$SPACEID/objects?done=true" | jq

"${CURL[@]}" "$URL/v1/spaces/$SPACEID/objects?camera_iso=1000" | jq

# output
#
# {
#   "object": "error",
#   "status": 400,
#   "code": "bad_request",
#   "message": "invalid filter at index 0: invalid value for property \"done\": bad input: property \"done\" must be a boolean"
# }

# {
#   "object": "error",
#   "status": 400,
#   "code": "bad_request",
#   "message": "invalid filter at index 0: invalid value for property \"cameraIso\": bad input: property \"cameraIso\" must be a number"
# }
```

### Environment

```markdown
Anytype desktop 0.53.1
Macos ARM64
Using API 2025-11-08
```

### Anything else?

It appears as if query filter values are always interpreted as strings, without attempting to parse the value according to the data type of the property.

stevelr:
- [X] I understand that contributing to this repository will require me to agree with the [CLA](https://github.com/anyproto/.github/blob/main/docs/CLA.md)

---

### Description

filters of the form "property_type=value" do not work for
 - boolean properties, where value is 'true' or 'false'
 - numeric properties, where value is a number (int or floating point)

Also, [in] and [not-in] conditions don't work for numbers.
The error returned in "property type is not a boolean/number"

I didn't add support for bool property [in] and [not-in] since I didn't think it made sense.

- [ ] 🍕 Feature
- [X] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [X] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

Fixes https://github.com/anyproto/anytype-heart/issues/2879

This is one of three bug fix PRs that arose out of testing object queries and filters in a rust SDK for Anytype (the problems can be reproduced with curl so they are not rust-related).

[2881](https://github.com/anyproto/anytype-heart/pull/2881) Filter bool numbers
[2880](https://github.com/anyproto/anytype-heart/pull/2880) core/api: allow type object ids in type filters
[2882](https://github.com/anyproto/anytype-heart/pull/2882) core/api/filter: allow select tag ids

To keep the changes simpler and easier to review, they are separate PRs. I have another [branch](https://github.com/stevelr/anytype-heart/tree/testing)  that combines them for testing - let me know if you'd rather I submit that as a PR instead.

### Mobile & Desktop Screenshots/Recordings
N/A

### Added tests?

- [X] 👍 yes

### Added to documentation?

- [ ] 📜 README.md
- [X] 🙅 no documentation needed

### [optional] Are there any post-deployment tasks we need to perform?
N/A


core/api/filter/validator.go
```diff
@@ -111,6 +113,83 @@ func (v *Validator) convertAndValidateValue(spaceId string, filter *Filter, prop
...
+			trimmed := strings.TrimSpace(v)
+			parsed, err := strconv.ParseBool(trimmed)
+			if err != nil {
+				return nil, util.ErrBadInput(fmt.Sprintf("invalid boolean value %q", v))
+			}
+			value = parsed
+		}
+	}
+	if property.Format == apimodel.PropertyFormatNumber {
+		switch v := value.(type) {
```
deff7, line 130: Let's create a helper function `func convertAnyToFloat(v any) (float64, error)`. It can be used later in `In` and `NotIn` case

pkg/lib/database/filter.go
```diff
@@ -100,19 +100,6 @@ func makeFilterByCondition(spaceID string, rawFilter FilterRequest, store Object
 		}
 	}
 
-	// replaces "value == false" to "value != true" for expected work with checkboxes
-	if rawFilter.Condition == model.BlockContentDataviewFilter_Equal {
```
deff7, line 111: This will break logic when an object doesn't have a property in its details list. For example, filter `done = false` won't work for this object

```
{
  "id": "blah-blah-blah"
}
```

But it works for

```
{
  "id": "blah-blah-blah",
  "done": false
}
```

deff7 (changes requested):

core/api/filter/validator.go
stevelr: Done. Created `convertAnyToFloat`

pkg/lib/database/filter.go
stevelr: Current implementation - I think this matches your intent:

checkbox EQ value: property defined AND equal to value
checkbox NE value: property undefined OR equal to !value

This provides more flexibility for the caller than implementing NE as (defined and EQ !value) because they could just use EQ for that.

I considered implementing IN and NOT-IN, but domain.Value doesn't implement []bool lists, and I wanted to keep the scope of this PR narrow and not change Value. A caller can trivially condense an array of bools into a single value and just use EQ or NEQ.

deff7: I wrote a simple test, that is not passing without EQ(false) => NEQ(true) conversion. 
```
	t.Run("equal false: no key", func(t *testing.T) {
		eq := FilterEq{Key: "k", Value: domain.Bool(false), Cond: model.BlockContentDataviewFilter_Equal}
		obj := domain.NewDetails()
		assertFilter(t, eq, obj, true)
	})
```
For example, by default we don't set `done` to `false`. This isn't optimal logic, yes, but we rely on it, and changing this will affect clients. It's good to use more explicit way to query either non-existent key or value equals to false, but we need to update our clients first and thoroughly check code that could rely on this logic.

pkg/lib/database/filter.go
```diff
@@ -107,32 +107,6 @@ func makeFilterByCondition(spaceID string, rawFilter FilterRequest, store Object
...
-				RelationKey:      rawFilter.RelationKey,
-				RelationProperty: rawFilter.RelationProperty,
-				Condition:        model.BlockContentDataviewFilter_NotEqual,
-				Value:            domain.Bool(true),
-			}
-		}
-
-	}
-	// replaces "value != false" to "value == true" for expected work with checkboxes
-	if rawFilter.Condition == model.BlockContentDataviewFilter_NotEqual {
```
deff7, line 124: Unfortunately, we have strange logic for NEQ(false): the missing key still passes this filter. Hence this substitution. We should revise internal logic for FilterEq to treat only `missing key` and `false` as falsy value.

pkg/lib/database/filter_test.go
```diff
@@ -438,7 +458,7 @@ func TestMakeAndFilter(t *testing.T) {
 		assertFilter(t, f, g, false)
 
 		g = domain.NewDetailsFromMap(map[domain.RelationKey]domain.Value{"not_exists": domain.Bool(false)})
-		assertFilter(t, f, g, false)
+		assertFilter(t, f, g, true)
```
deff7: This is not correct, falsy value shouldn't pass `!= false`

stevelr:
@deff7 
thanks for taking a look at this.
sorry about my earlier misunderstanding about missing vs false. I think I fixed it, and fixed the test. I asked claude to review this using your go-reviewer skill and it found no critical/high/medium issues. I didn't address the low priority issues to keep this pr more targeted.

deff7:
@stevelr Hi, looks like you've merged two separate PRs into this one. Could you please remove changes not related to the original PR?

PR #2886, opened 2026-01-09
GO-6331 objectstorestore remove on space leave

requilence:
### TL;DR

**Problem:** When a user was kicked from a space (by another participant), the cleanup only removed ACL indexes via `cleanACLIndexes()` but left the space database files on disk. The space data was never fully removed.

**Solution:** Introduce proper space index lifecycle management tied to space mode transitions:
- `OnSpaceModeChange` hook triggers `CloseSpaceIndex()` when space enters `Offloading` mode
- `DeleteSpaceIndex()` permanently deletes the database files from filesystem and marks the proxy as deactivated

**Why storeProxy:** The proxy pattern was needed to safely handle the "deactivated but still referenced" state. After a space is deleted, existing code paths (like cross-space subscriptions or late callbacks) might still call `SpaceIndex(deletedSpaceId)`. Without the proxy:
- Direct access would panic on nil
- Creating a new store would resurrect the deleted database via `anystore.Open()`

The proxy solves both: it returns `sharedEmptyStore` for deactivated spaces (safe empty results) and the `removed` flag prevents re-initialization even if `Init()` is called.

---

### Key Concepts

| State | `initialized` | `removed` | Behavior |
|-------|---------------|-----------|----------|
| **Uninitialized** | false | false | Empty results, can Init() |
| **Initialized** | true | false | Real store operations |
| **Closed** | false | false | Empty results, can re-Init() |
| **Deactivated** | false | true | Empty results, Init() returns `ErrSpaceRemoved` |

### Flow

```
USER KICKED FROM SPACE:
1. Space mode → Offloading
2. OnSpaceModeChange() → CloseSpaceIndex()
3. storeProxy.initialized = false → delegates to sharedEmptyStore
4. DeleteSpaceIndex() → proxy.Deactivate() sets removed=true, deletes DB from disk

AFTER DELETION (safe access):
SpaceIndex(deletedId) → same proxy (in map) → removed=true → sharedEmptyStore → empty results, no DB recreation
Init() on deactivated proxy → returns ErrSpaceRemoved
```

### Syncstatus optimization

- Added `AllLoadedSpaceIds()` which only returns spaces in `ModeLoading` state
- Excludes marketplace space and spaces being offloaded/deleted/joining
- Reduces unnecessary sync status queries for inactive spaces


pkg/lib/localstore/objectstore/spaceindex/store.go
```diff
@@ -138,8 +168,402 @@ type Deps struct {
...
+
+func (p *storeProxy) UpdateObjectDetails(ctx context.Context, id string, details *domain.Details) error {
+	if p.initialized.Load() {
+		return p.realStore.UpdateObjectDetails(ctx, id, details)
+	}
+	return sharedEmptyStore.UpdateObjectDetails(ctx, id, details)
+}
+
+func (p *storeProxy) SubscribeForAll(callback func(rec database.Record)) {
+	// Store the callback in the proxy so it survives until dsObjectStore is created.
```
deff7, line 349: why this is required? In order for subscriptions to work we need to re-run initial subscription query, not only register a callback

deff7 (approved):

requilence:
Superseded by #3156. The store-removal feature has been re-ported onto the lazy-cross-space (GO-7288) base without the storeProxy/emptyStore/mode-observer machinery. Closing this older approach in favor of the simpler implementation. Same Linear issue: GO-6331.

PR #2902, opened 2026-01-26
GO-6630 Fix HEIC imgs upload

KirillSto:
https://linear.app/anytype/issue/GO-6630/the-image-heic-format-is-broken

ReadSeeker of Heif image file is used twice: 
- first to find orientation info
- second to call goheif.Decode

Seek to beginning of file was not made correctly on first step, that's why file was not decoded.

Fix: Read all bytes at the beginning and then call functions with two independent readers

### DIFF
--- a/pkg/lib/mill/image_resize_heic.go
+++ b/pkg/lib/mill/image_resize_heic.go
@@ -15,19 +15,13 @@ import (
 )
 
 func (m *ImageResize) resizeHEIC(r io.ReadSeeker) (*Result, error) {
-	data, err := io.ReadAll(r)
-	if err != nil {
-		return nil, fmt.Errorf("read heic: %w", err)
-	}
-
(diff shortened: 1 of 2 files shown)


pkg/lib/mill/image_resize_heic.go
```diff
@@ -15,13 +15,19 @@ import (
 )
 
 func (m *ImageResize) resizeHEIC(r io.ReadSeeker) (*Result, error) {
-	orientation, err := getHEICOrientation(r)
+	data, err := io.ReadAll(r)
```
deff7: why to read all bytes and double the memory consumption? Maybe use Seek to the beginning?

deff7 (changes requested):

deff7 (approved):

PR #2914, opened 2026-02-01
Fix array-type filter conditions (ObjectsFilterItem, MultiSelectFilterItem)

### LINKED ISSUE #2897: Search API filter failure on array-type properties

### Current Behavior

When trying to perform a search through the API, using a "multi_select" condition, the search fails with a:
```json
{
  "object": "error",
  "status": 400,
  "code": "bad_request",
  "message": "failed to build expression filters"
}
```

### Expected Behavior

Be able to search using multi_select filters on results.

### Steps To Reproduce

Using the latest anytype version, try to perform a search for a task that has a specific tag, through the API, using this JSON body:
```json
{
      "types": ["task"],
      "filters": {
        "conditions": [
          {
            "property_key": "tag",
            "condition": "all",
            "multi_select": ["whatever ID you want"]
          }
        ]
      }
    }
```  

I've also tried this with objects that have relations. For example, define an object "Article", and as a property, use another object type such as "Author" or something, then try a search like:
```json
{
      "types": ["article"],
      "filters": {
        "conditions": [
          {
            "property_key": "authors",
            "condition": "in",
            "multi_select": ["some_author", "some_other_author"]
          }
        ]
      }
    }
```

Last one I suspect, but have not actually tried, is with FilesFilterItem.  

### Environment

```markdown
- OS: Linux
- Version: 0.53.1
```

### Anything else?

What I suspect is that for filters of these types, namely:
```go
type FilesFilterItem struct {
	PropertyKey string          `json:"property_key" example:"files"`                                                // The property key to filter on
	Condition   FilterCondition `json:"condition" example:"in" enums:"in,all,nin,empty,nempty"`                      // The filter condition
	Files       *[]string       `json:"files" example:"bafyreie6n5l5nkbjal37su54cha4coy7qzuhrnajluzv5qd5jvtsrxkequ"` // File IDs for contains condition
}

// MultiSelectFilterItem for multi-select property filters
type MultiSelectFilterItem struct {
	PropertyKey string          `json:"property_key" example:"tag"`                                                                                                                     // The property key to filter on
	Condition   FilterCondition `json:"condition" example:"all" enums:"in,all,nin,empty,nempty"`                                                                                        // The filter condition
	MultiSelect *[]string       `json:"multi_select" example:"bafyreiaixlnaefu3ci22zdenjhsdlyaeeoyjrsid5qhfeejzlccijbj7sq,bafyreie6n5l5nkbjal37su54cha4coy7qzuhrnajluzv5qd5jvtsrxkequ"` // The tag IDs to filter by
}

// ObjectsFilterItem for objects/relation property filters
type ObjectsFilterItem struct {
	PropertyKey string          `json:"property_key" example:"creator"`                                                // The property key to filter on
	Condition   FilterCondition `json:"condition" example:"in" enums:"in,all,nin,empty,nempty"`                        // The filter condition
	Objects     *[]string       `json:"objects" example:"bafyreie6n5l5nkbjal37su54cha4coy7qzuhrnajluzv5qd5jvtsrxkequ"` // Object Ids to filter by
}
```

we have this type of definition:
```go
Objects     *[]string       `json:"objects" example:"bafyreie6n5l5nkbjal37su54cha4coy7qzuhrnajluzv5qd5jvtsrxkequ"` // Object Ids to filter by

// or

Files       *[]string       `json:"files" example:"bafyreie6n5l5nkbjal37su54cha4coy7qzuhrnajluzv5qd5jvtsrxkequ"` // File IDs for contains condition

// or 

MultiSelect *[]string       `json:"multi_select" example:"bafyreiaixlnaefu3ci22zdenjhsdlyaeeoyjrsid5qhfeejzlccijbj7sq,bafyreie6n5l5nkbjal37su54cha4coy7qzuhrnajluzv5qd5jvtsrxkequ"` // The tag IDs to filter by
```

Then, in the parsing logic we have this type of thing:
```go
func (f MultiSelectFilterItem) GetValue() interface{} {
	if f.MultiSelect != nil {
		return *f.MultiSelect
	}
	return nil
}
```
and also (in core/api/service/property.go, in `SanitizeAndValidatePropertyValue`):
```go
case apimodel.PropertyFormatMultiSelect:
		keysOrIds, ok := value.([]interface{})
```

Which I believe will not work in go, because `[]string` cannot be type-asserted to a `[]interface{}`.  
You can try this code to check it out:  
```go
package main

import (
	"encoding/json"
	"fmt"
)

func SanitizeAndValidatePropertyValue(key string, value interface{}) error {
	// This type assertion is what causes the "failed to build expression filters" error.
	// It tries to cast the input directly to []interface{}.
	keysOrIds, ok := value.([]interface{})
	if !ok {
		return fmt.Errorf("CRITICAL ERROR: property %q must be an array of tag ids or keys (Type assertion failed! Input is %T)", key, value)
	}
	// -------------------------------

	fmt.Printf("Success! Processed %d items for %q\n", len(keysOrIds), key)
	return nil
}

func main() {
	// 1. Simulate the JSON payload from the user
	jsonPayload := `{"property_key": "tag", "condition": "in", "multi_select": ["a tag you want"]}`
	fmt.Printf("1. Incoming JSON Payload: %s\n", jsonPayload)

	// 2. Simulate what happens inside the server's JSON Unmarshaler
	// In model/filter.go, MultiSelectFilterItem defines MultiSelect as *[]string
	var decoded struct {
		MultiSelect *[]string `json:"multi_select"`
	}

	err := json.Unmarshal([]byte(jsonPayload), &decoded)
	if err != nil {
		fmt.Printf("Error decoding JSON: %v\n", err)
		return
	}

	incomingData := *decoded.MultiSelect
	fmt.Printf("2. Data Type after JSON Unmarshal (defined as *[]string): %T\n", incomingData)

	// 3. Simulate the Server Call to the buggy function
	// The server passes this []string as an interface{}
	err = SanitizeAndValidatePropertyValue("tags", incomingData)

	if err != nil {
		fmt.Println("\n[X] BUG REPRODUCED:")
		fmt.Println(err)
	}

	fmt.Println("\n=== Demonstrating the Fix ===")

	// The Fix: Explicit conversion
	convertedData := make([]interface{}, len(incomingData))
	for i, v := range incomingData {
		convertedData[i] = v
	}

	fmt.Printf("3. Converted Data Type: %T\n", convertedData)

	err = SanitizeAndValidatePropertyValue("tags_fixed", convertedData)
	if err != nil {
		fmt.Println(err)
	} else {
		fmt.Println("[V] SUCCESS: Validation passed with converted data")
	}
}
```  

One thing that you can also try is to do a search with a payload like this:  
```json
{
      "types": ["article"],
      "filters": {
        "conditions": [
          {
            "property_key": "authors",
            "condition": "nempty",
            "multi_select": ["some_author", "some_other_author"]
          }
        ]
      }
    }
```

This will not return an error, as the problematic logic is bypassed, I think, here: 

```go
func (v *Validator) convertAndValidateValue(spaceId string, filter *Filter, property *apimodel.Property, propertyMap map[string]*apimodel.Property) (interface{}, error) {
	switch filter.Condition {
	case model.BlockContentDataviewFilter_Empty, model.BlockContentDataviewFilter_NotEmpty:
		if boolVal, ok := filter.Value.(bool); ok {
			return boolVal, nil
		}
		return true, nil // <------ RETURNS HERE, bypassing the call to SanitizeAndValidatePropertyValue
	}

	value := filter.Value
	if filter.Condition == model.BlockContentDataviewFilter_In || filter.Condition == model.BlockContentDataviewFilter_NotIn {
		switch v := value.(type) {
		case []string, []interface{}:
			// Already an array, keep as-is
		default:
			// Wrap single value in array
			value = []interface{}{v}
		}
	}

	return v.apiService.SanitizeAndValidatePropertyValue(spaceId, filter.PropertyKey, value, property, propertyMap)
}
```

I can try to write a PR for this if you want, but I wanted to submit that first, to make sure I'm not missing anything.

themightychris:
## Summary

Fixes the `failed to build expression filters` error when using `in`, `all`, or `nin` conditions on array-type properties (`objects`, `multi_select`, `files`) in the Search API.

## Problem

The filter item types (`ObjectsFilterItem`, `MultiSelectFilterItem`, `FilesFilterItem`) return `[]string` from their `GetValue()` methods, but `SanitizeAndValidatePropertyValue` expected `[]interface{}`. This type assertion failure caused all array-based filter conditions to fail.

Additionally, the `links` property (a derived system property) was incorrectly validated to require object/member layouts, but it can contain references to any object type (files, types, pages, etc.).

## Changes

- **`core/api/service/property.go`**: 
  - `SanitizeAndValidatePropertyValue` now accepts both `[]string` and `[]interface{}` for `objects`, `files`, and `multi_select` property formats
  - Added `isValidObjectReferenceForProperty()` that relaxes validation for the `links` property

- **`core/api/filter/expression_test.go`**: Added tests for `ObjectsFilterItem` with `in`/`all` conditions

- **`core/api/service/property_test.go`**: Added tests for `SanitizeAndValidatePropertyValue` handling `[]string` input for objects and multi_select formats

## Test Plan

- [x] All existing tests pass
- [x] New tests verify `[]string` input is accepted for objects properties
- [x] New tests verify `[]string` input is accepted for multi_select properties  
- [x] New tests verify `links` property accepts any object type (e.g., files)

Fixes #2897

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/core/api/filter/expression_test.go
+++ b/core/api/filter/expression_test.go
@@ -499,6 +499,71 @@ func TestBuildExpressionFilters(t *testing.T) {
 				assert.Equal(t, pbtypes.String("test"), result.Value)
 			},
 		},
+		{
+			name: "objects filter with in condition",
+			expr: &apimodel.FilterExpression{
+				Conditions: []apimodel.FilterItem{
+					{
+						WrappedFilterItem: apimodel.ObjectsFilterItem{
+							PropertyKey: "links",
+							Condition:   apimodel.FilterConditionIn,
+							Objects:     &[]string{"obj1", "obj2"},
+						},
+					},
+				},
+			},
+			setupMock: func(m *mock_filter.MockApiService) {
+				propertyMap := map[string]*apimodel.Property{
+					"links": {
+						Key:         "links",
+						RelationKey: bundle.RelationKeyLinks.String(),
+						Format:      apimodel.PropertyFormatObjects,
+					},
+				}
+				m.On("GetCachedProperties", spaceId).Return(propertyMap)
+				m.On("ResolvePropertyApiKey", propertyMap, "links").Return("links", true)
+				// ObjectsFilterItem.GetValue() returns []string, not []interface{}
+				m.On("SanitizeAndValidatePropertyValue", spaceId, "links", []string{"obj1", "obj2"}, propertyMap["links"], propertyMap).Return([]string{"obj1", "obj2"}, nil)
+			},
+			checkResult: func(t *testing.T, result *model.BlockContentDataviewFilter) {
+				require.NotNil(t, result)
+				assert.Equal(t, bundle.RelationKeyLinks.String(), result.RelationKey)
+				assert.Equal(t, model.BlockContentDataviewFilter_In, result.Condition)
+				assert.Equal(t, pbtypes.StringList([]string{"obj1", "obj2"}), result.Value)
+			},
+		},
+		{
+			name: "objects filter with all condition",
+			expr: &apimodel.FilterExpression{
+				Conditions: []apimodel.FilterItem{
+					{
+						WrappedFilterItem: apimodel.ObjectsFilterItem{
+							PropertyKey: "assignee",
+							Condition:   apimodel.FilterConditionAll,
+							Objects:     &[]string{"user1", "user2"},
+						},
+					},
+				},
+			},
+			setupMock: func(m *mock_filter.MockApiService) {
+				propertyMap := map[string]*apimodel.Property{
+					"assignee": {
+						Key:         "assignee",
+						RelationKey: bundle.RelationKeyAssignee.String(),
+						Format:      apimodel.PropertyFormatObjects,
+					},
+				}
+				m.On("GetCachedProperties", spaceId).Return(propertyMap)
+				m.On("ResolvePropertyApiKey", propertyMap, "assignee").Return("assignee", true)
+				m.On("SanitizeAndValidatePropertyValue", spaceId, "assignee", []string{"user1", "user2"}, propertyMap["assignee"], propertyMap).Return([]string{"user1", "user2"}, nil)
+			},
+			checkResult: func(t *testing.T, result *model.BlockContentDataviewFilter) {
+				require.NotNil(t, result)
+				assert.Equal(t, bundle.RelationKeyAssignee.String(), result.RelationKey)
+				assert.Equal(t, model.BlockContentDataviewFilter_AllIn, result.Condition)
+				assert.Equal(t, pbtypes.StringList([]string{"user1", "user2"}), result.Value)
+			},
+		},
 	}
 
 	for _, tt := range tests {
--- a/core/api/filter/expression_test.go
+++ b/core/api/filter/expression_test.go
@@ -523,7 +523,13 @@ func TestBuildExpressionFilters(t *testing.T) {
 				m.On("GetCachedProperties", spaceId).Return(propertyMap)
 				m.On("ResolvePropertyApiKey", propertyMap, "links").Return("links", true)
 				// ObjectsFilterItem.GetValue() returns []string, not []interface{}
-				m.On("SanitizeAndValidatePropertyValue", spaceId, "links", []string{"obj1", "obj2"}, propertyMap["links"], propertyMap).Return([]string{"obj1", "obj2"}, nil)
+				// The service must accept []string input for this to work
+				m.On("SanitizeAndValidatePropertyValue", spaceId, "links",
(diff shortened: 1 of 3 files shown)


themightychris:
Note: This PR has significant overlap with #2880 by @stevelr, which also fixes the `[]string` type assertion bug for array-type filters.

**Differences:**
- Our approach handles `[]string` directly without converting to `[]interface{}`
- #2880 special-cases the `type` property to allow `objectType` layout
- This PR special-cases the `links` property to allow any object type (files, types, pages, etc.)

Both PRs address the core issue reported in #2897. It may make sense to combine the special cases from both approaches.

PR #2922, opened 2026-02-03
GO-6796: File uploader: handle deleted spaces

deff7:


core/files/filesync/delete.go
```diff
@@ -80,11 +81,16 @@ func (s *fileSync) processNextToDelete(ctx context.Context) error {
 func (s *fileSync) processDeletion(ctx context.Context, it FileInfo) (FileInfo, error) {
 	err := s.rpcStore.DeleteFiles(ctx, it.SpaceId, it.FileId)
 	if err != nil {
+		errStr := err.Error()
+		if strings.Contains(errStr, "space is deleted") || strings.Contains(errStr, "code: 502") {
```
KirillSto, line 85: Are these checks ok? Maybe use errors.Is here?

deff7: unfortunately, we don't have defined errors for it, they're just strings leaking from any-sync-filenode

KirillSto (approved):

PR #2884, opened 2026-01-07
Fix account key fallback

stevelr:
- [X] I understand that contributing to this repository will require me to agree with the [CLA](https://github.com/anyproto/.github/blob/main/docs/CLA.md)

---

### Description

Ensure account key derivation falls back to identity when legacy key is missing, preventing nil signing keys.

Changes: pkg/lib/core/wallet.go
  
Codex created this tentative patch, which seems to have fixed the issue for me.

### What type of PR is this? (check all applicable)

- [X] 🐛 Bug Fix

### Related Tickets & Documents

https://github.com/anyproto/anytype-heart/issues/2883

### Mobile & Desktop Screenshots/Recordings
N/A

### Added tests?

- [X] 🙋 no, because I need help

### Added to documentation?

- [X] 🙅 no documentation needed

PR #2925, opened 2026-02-03
fix: handle nil OldAccount for post-migration accounts

jsandai:
## Problem

Accounts created after the identity migration (from `m/44'/607'` to `m/44'/2046'`) don't have an `oldIdentity` stored in their wallet. When these accounts attempt to sync, the `SpaceSign` payload contains a nil `OldAccount`, causing:

```
SpaceSign: payload.OldAccount is nil
```

This prevents new accounts from syncing with self-hosted coordinators.

## Solution

Check if `GetOldAccountKey()` returns nil and fall back to the current identity (`GetAccountPrivkey()`). This allows new accounts to sync while maintaining backward compatibility for migrated accounts.

## Changes

- `space/spacecore/credentialprovider/credentialprovider.go`: Add nil check with fallback
- `core/configfetcher/configfetcher.go`: Same pattern

## Testing

Tested locally with a fresh account on a self-hosted Anytype deployment. Account now syncs successfully.

### DIFF
--- a/core/configfetcher/configfetcher.go
+++ b/core/configfetcher/configfetcher.go
@@ -88,10 +88,16 @@ func (c *configFetcher) updateStatus(ctx context.Context) (err error) {
 		if sErr != nil {
 			return sErr
 		}
+		// For accounts created after identity migration, OldAccount doesn't exist.
+		// Fall back to current identity to avoid nil pointer in SpaceSign.
+		oldAccount := c.wallet.GetOldAccountKey()
+		if oldAccount == nil {
+			oldAccount = c.wallet.GetAccountPrivkey()
+		}
 		payload := coordinatorclient.SpaceSignPayload{
 			SpaceId:     techSpace.Id(),
 			SpaceHeader: state.SpaceHeader,
-			OldAccount:  c.wallet.GetOldAccountKey(),
+			OldAccount:  oldAccount,
 			Identity:    c.wallet.GetAccountPrivkey(),
 		}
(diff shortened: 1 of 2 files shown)


jsandai:

stevelr:
I think this is fixed in SpaceSign in any-sync https://github.com/anyproto/any-sync/pull/607, but I don't think it's been tagged in a release yet

jsandai:
> I think this is fixed in SpaceSign in any-sync [anyproto/any-sync#607](https://github.com/anyproto/any-sync/pull/607), but I don't think it's been tagged in a release yet

Thanks for the pointer @stevelr!

You're right; any-sync #607 (https://github.com/anyproto/any-sync/pull/607) takes a cleaner approach by deprecating the old identity system entirely, removing the OldAccount requirement from SpaceSignPayload.

Our PR is essentially a stopgap for the current anytype-heart code that still expects OldAccount. Once anytype-heart bumps its any-sync dependency and updates the credential provider to match #607's changes, this PR would become redundant.

Happy to close this if the maintainers prefer to wait for the dependency update, or keep it as a temporary fix for anyone running current anytype-heart versions with self-hosted setups.

stevelr:
I'm not much of a go developer so when I ran into the same crash I had codex create a pr.

https://github.com/anyproto/anytype-heart/pull/2884

requilence:
closed in favor of #2931 + https://github.com/anyproto/anytype-heart/pull/607

PR #2927, opened 2026-02-05
GO-6742 | GO-6833 - Support TurnInto for toggle headers

KirillSto:
https://linear.app/anytype/issue/GO-6742/add-toggle-heading-block-styles-toggle-h1-h2-h3
https://linear.app/anytype/issue/GO-6833/conversion-to-toggle-heading-does-not-capture-blocks-until-the-next

Support TurnInto for Toggle Headers

### DIFF
--- a/core/block/editor/stext/text.go
+++ b/core/block/editor/stext/text.go
@@ -6,8 +6,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/globalsign/mgo/bson"
-
 	"github.com/anyproto/anytype-heart/core/block/editor/smartblock"
 	"github.com/anyproto/anytype-heart/core/block/editor/state"
 	"github.com/anyproto/anytype-heart/core/block/editor/template"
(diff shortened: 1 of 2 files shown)


core/block/editor/stext/text.go
```diff
@@ -426,6 +430,212 @@ func (t *textImpl) TurnInto(ctx session.Context, style model.BlockContentTextSty
...
+		currentDivBlocks = append(currentDivBlocks, siblingId)
+	}
+
+	// If we didn't hit a stopping header in current div, continue to sibling divs
+	var siblingDivBlocks [][]string
+	if !stopped {
+		siblingDivBlocks = collectSiblingDivBlocks(s, grandparent, targetLevel, divPos)
+	}
+
+	addNewDiv := func(blockIds []string) {
```
deff7: I don't know if constructing div blocks by hand is a good idea. I think div algorithm should do this automatically.

Maybe all we need to do is to collect all blocks that should be children of a header and put them in the header. Then `normalize` will take care of divs

core/block/editor/stext/text.go
```diff
@@ -426,6 +428,170 @@ func (t *textImpl) TurnInto(ctx session.Context, style model.BlockContentTextSty
...
+// collectBlocksFromDivs collects block IDs from div children of divsParent, starting from the div
+// at divPos (from startChildIdx within that div) and continuing to subsequent sibling divs.
+// It stops when it encounters a header of same or higher level than targetLevel.
+// Empty divs that had all their children collected are unlinked.
+func collectBlocksFromDivs(s *state.State, divsParent simple.Block, targetLevel, divPos, startChildIdx int) []string {
+	var blocks []string
+	var emptyDivsToUnlink []string
+
+	// Take a snapshot of div IDs to avoid issues with slice mutation during unlinking
+	divIds := append([]string(nil), divsParent.Model().ChildrenIds[divPos:]...)
```
deff7: slices.Clone

core/block/editor/stext/text.go
KirillSto: Yes, thank you! I've made the logic more simple - we just make all these blocks below the children of toggle header

core/block/editor/stext/text.go
KirillSto: Basically we do not need to copy a slice of children, as we modify the slice after loop, so I removed copying line

deff7 (approved):

PR #2943, opened 2026-02-11
GO-6840 Add protoc-gen-ts and install-dev-ts for TypeScript protobuf generation

ra3orblade:
## Summary
- Add `ts-protoc-gen` npm dependency for TypeScript protobuf declaration generation
- Add `protos-ts` Make target and separate `install-dev-ts` command (keeps `install-dev-js` unchanged as CommonJS-only)
- Add configurable `CLIENT_LIB_DIR` variable (defaults to `dist/lib`, use `middleware` for anytype-bun)
- Fix duplicate `cp` line and add `mkdir -p` for fresh target directories

## Usage
```bash
# CommonJS (unchanged)
make install-dev-js

# TypeScript declarations → anytype-ts
make install-dev-ts

# TypeScript declarations → anytype-bun with custom lib dir
make install-dev-ts CLIENT_DESKTOP_PATH=../anytype-bun CLIENT_LIB_DIR=middleware
```

## Test plan
- [x] `make install-dev-js` installs CommonJS bindings to `../anytype-ts/dist/lib`
- [x] `make install-dev-js CLIENT_DESKTOP_PATH=../anytype-bun CLIENT_LIB_DIR=middleware` installs to `../anytype-bun/middleware`
- [x] `make install-dev-ts` generates `.d.ts` files via `protoc-gen-ts`
- [ ] Verify generated `.d.ts` files are usable in downstream TypeScript projects


requilence:
We should move JS/TS bindings generation out of anytype-heart. Let's not add more bindings generators here.  
Why it's a good idea:  
* The client will always have the same version of the generator and library to work with this bindings. In some cases, inconsistent versions are critical (we have experienced this before with js bindings, and with anytype-swift as well).  
* The client can update the bindings generator separately whenever they want.

So we can copy only binary and proto files and run the generation on client-side

PR #2955, opened 2026-02-16
Fix restore (import with UpdateExistingObjects) of archived object and existing object lookup

stevelr:
### Description

Fix restore (import with UpdateExistingObjects) of archived object
- archive toggle was one-dirrectional (archive only)
- archive overwrote lastModifiedDate as a side-effect of un-archiving

Compare current vs desired archived state and restore the snapshot lastModifiedDate after unarchive.

Also add direct ID matching via oldAnytypeID when getExisting is true, migrate store queries from QueryObjectIds to QueryRaw with limit=1, and skip queries on empty filter values.

---

- [X] I understand that contributing to this repository will require me to agree with the [CLA](https://github.com/anyproto/.github/blob/main/docs/CLA.md)

---

- [X] 🐛 Bug Fix
- [X] ✅ Test

### Added tests?

- [X] 👍 yes

### DIFF
--- a/core/block/import/common/objectcreator/objectcreator.go
+++ b/core/block/import/common/objectcreator/objectcreator.go
@@ -158,9 +158,15 @@ func (oc *ObjectCreator) Create(dataObject *DataObject, sn *common.Snapshot) (*d
 	}
 	oc.setFavorite(snapshot, newID)
 
-	oc.setArchived(ctx, snapshot, newID)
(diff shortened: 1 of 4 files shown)


stevelr (commented):
core/block/import/common/objectid/existingobject.go line 73-83

I didn't change the error handling here, but it looks like `err` is getting overwritten here, so only the second error would be returned to the caller. Is that intentional?

PR #2960, opened 2026-02-18
GO-6873 Optimize mark read handler and fix Flush race condition

deff7:
## Summary
- Optimize `SetReadFlag` and `SetSyncedFlag` with batch `Find+IN` queries and per-batch transactions for better performance
- Fix data race in `GetMessages` 
- Fix race condition where concurrent `Flush` calls from diff manager callbacks could reset `needReloadState` before the main `onUpdate` Flush, causing permanently stale chat state after message deletion

## Test plan
- [x] Existing `TestFlush` tests pass
- [x] New `TestFlushReloadStateFlag` verifies that `Flush(false)` skips state reload while `Flush(true)` performs it
- [x] New repository batch operation tests pass
- [ ] Manual testing of chat message read/unread/delete flows


core/block/chats/chatrepository/repository.go
```diff
@@ -415,59 +412,183 @@ func (s *repository) GetMessagesForIndexing(ctx context.Context, afterOrderId st
 	return s.queryMessages(ctx, qry)
 }
 
-func (r *repository) SetReadFlag(ctx context.Context, chatObjectId string, msgIds []string, counterType chatmodel.CounterType, value bool) []string {
+func (r *repository) SetReadFlag(ctx context.Context, chatObjectId string, msgIds []string, counterType chatmodel.CounterType, value bool) ([]string, error) {
 	handler := newReadHandler(counterType)
 
+	arena := r.arenaPool.Get()
+	defer func() {
+		arena.Reset()
```
cheggaaa, line 428: pool.Put already makes Reset

core/block/chats/chatrepository/readhandler.go
```diff
@@ -8,15 +8,19 @@ import (
...
 	getReadKey() string
 	readModifier(value bool) query.Modifier
 }
 
 type readMessagesHandler struct{}
 
-func (h readMessagesHandler) getUnreadFilter() query.Filter {
+func (h readMessagesHandler) getReadFilter(value bool) query.Filter {
+	if value {
+		return query.Key{Path: []string{chatmodel.ReadKey}, Filter: query.NewComp(query.CompOpEq, true)}
```
cheggaaa: It's better to define static filters as a global var. To avoid allocs, now it's safe.

core/block/chats/chatrepository/repository.go
```diff
@@ -415,59 +412,183 @@ func (s *repository) GetMessagesForIndexing(ctx context.Context, afterOrderId st
...
+	return idsModified, nil
+}
+
+func (r *repository) setReadFlag(ctx context.Context, arena *anyenc.Arena, handler readHandler, msgIds []string, value bool) ([]string, error) {
+	arena.Reset()
+	encIds := make([]*anyenc.Value, 0, len(msgIds))
+	for _, id := range msgIds {
+		encIds = append(encIds, arena.NewString(id))
+	}
+	iter, err := r.collection.Find(query.And{
```
cheggaaa: Why not coll.Find().Update(modifier) ?

cheggaaa: Also it seems you use different transactions for read and write, which can break consistency

deff7: > coll.Find().Update(modifier) 

we need to collect all modified ids

cheggaaa: > > coll.Find().Update(modifier)
> 
> we need to collect all modified ids

query.ModifyFunc  takes an original value in args, so you can collect ids as well

core/block/chats/chatrepository/repository.go
deff7: fixed

core/block/chats/chatrepository/readhandler.go
deff7: done

core/block/chats/chatrepository/repository.go
deff7: thanks, now it's much cleaner

core/block/chats/chatrepository/repository.go
```diff
@@ -415,59 +411,146 @@ func (s *repository) GetMessagesForIndexing(ctx context.Context, afterOrderId st
...
+	var ids []string
+	for iter.Next() {
+		doc, err := iter.Doc()
+		if err != nil {
+			return nil, fmt.Errorf("read doc: %w", err)
 		}
+		ids = append(ids, doc.Value().GetString("id"))
+	}
+
+	if err = iter.Close(); err != nil {
```
cheggaaa: better to do defer iter.Close()

cheggaaa (approved):

PR #2987, opened 2026-03-03
GO-6597 Fix tag search in collections

KirillSto:
https://linear.app/anytype/issue/GO-6597/collection-searching-for-multi-select-does-not-find-every-entry

Fix tag search in collections

### DIFF
--- a/pkg/lib/localstore/objectstore/spaceindex/queries.go
+++ b/pkg/lib/localstore/objectstore/spaceindex/queries.go
@@ -242,7 +242,7 @@ func (s *dsObjectStore) QueryFromFulltext(results []database.FulltextResult, par
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


pkg/lib/localstore/objectstore/spaceindex/queries.go
```diff
@@ -242,7 +171,15 @@ func (s *dsObjectStore) QueryFromFulltext(results []database.FulltextResult, par
 			}
 		}
 
-		injectedResults := s.getInjectedResults(details, res.Score, res.Path, 10, params)
+		injectLimit := 0
+		if upperBound >= len(records) {
+			injectLimit = upperBound - len(records)
```
deff7, line 176: if length of records equals to upperBound inject limit is zero

deff7 (approved):

PR #2991, opened 2026-03-04
GO-6612 Handle outdent operation on move

KirillSto:
https://linear.app/anytype/issue/GO-6612/single-change-outdent-operation-for-list-blocks

Support outdent (opposite to indent) operations on blocks move. In this case sibling blocks that go after the moved one should become children of the moved one:

Example

Before outdent:

Parent Block
├── Block A (target for outdent)
├── Block B
└── Block C

After outdent of Block A:

Parent Block
Block A
├── Block B
└── Block C

### DIFF
--- a/core/block/editor/basic/basic.go
+++ b/core/block/editor/basic/basic.go
@@ -274,7 +274,7 @@ func (bs *basic) Move(srcState, destState *state.State, targetBlockId string, po
 		return err
 	}
 
-	var ctx outdentOpContext
+	var opCtx *outdentOpContext
 	var replacementCandidate simple.Block
 	for i, id := range blockIds {
 		if b := srcState.Pick(id); b != nil {
@@ -287,7 +287,7 @@ func (bs *basic) Move(srcState, destState *state.State, targetBlockId string, po
 
 			// Capture outdent context of last moving block before unlinking
 			if i == len(blockIds)-1 {
-				ctx = collectOutdentContext(srcState, id)
+				opCtx = collectOutdentContext(srcState, id)
 			}
 
 			srcState.Unlink(id)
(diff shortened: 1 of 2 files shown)


core/block/editor/basic/basic.go
```diff
@@ -313,7 +327,61 @@ func (bs *basic) Move(srcState, destState *state.State, targetBlockId string, po
...
 
-	return srcState.InsertTo(targetBlockId, position, blockIds...)
+	if err = srcState.InsertTo(targetBlockId, position, blockIds...); err != nil {
+		return err
+	}
+
+	return processOutdentOperation(ctx, srcState, targetBlockId, position)
+}
+
+func collectOutdentContext(srcState *state.State, blockId string) outdentOpContext {
```
deff7: I suggest returning a pointer if we have to do an outdent operation. Then check if ctx != nil to perform the operation

core/block/editor/basic/basic.go
```diff
@@ -313,7 +327,61 @@ func (bs *basic) Move(srcState, destState *state.State, targetBlockId string, po
...
+	// Get siblings after this position - make a copy to avoid slice aliasing
+	if ctx.positionInParent != -1 && ctx.positionInParent < len(parent.Model().ChildrenIds)-1 {
+		siblings := parent.Model().ChildrenIds[ctx.positionInParent+1:]
+		ctx.followingSiblings = make([]string, len(siblings))
+		copy(ctx.followingSiblings, siblings)
+	}
+	return ctx
+}
+
+func processOutdentOperation(ctx outdentOpContext, srcState *state.State, targetBlockId string, pos model.BlockPosition) error {
```
deff7: `ctx` looks like a `context.Context`, maybe call it `opCtx`?

core/block/editor/basic/basic.go
```diff
@@ -313,7 +327,61 @@ func (bs *basic) Move(srcState, destState *state.State, targetBlockId string, po
 		}
 	}
 
-	return srcState.InsertTo(targetBlockId, position, blockIds...)
+	if err = srcState.InsertTo(targetBlockId, position, blockIds...); err != nil {
+		return err
+	}
+
+	return processOutdentOperation(ctx, srcState, targetBlockId, position)
```
deff7: Perhaps it's cleaner to check if `opCtx != nil` and then call the  `processOutdentOperation`. Otherwise, without looking into a body of the function, it looks like we have to do outdenting every time.

deff7 (changes requested):

deff7 (approved):

PR #2992, opened 2026-03-04
GO-6743 Remove unused LastUnreadReactionOrderId proto field

deff7:
## Summary
- Removed unused `LastUnreadReactionOrderId` field from `ChatMessage` proto
- Simplified `unreadReactionIdsToModel()` to return only the map (no longer returns the unused order ID string)
- Removed corresponding test assertion

## Test plan
- [x] `go test ./core/block/chats/chatmodel/...` passes
- [x] `go build ./core/block/chats/...` compiles cleanly


pkg/lib/pb/model/protos/models.proto
```diff
@@ -1638,7 +1638,6 @@ message ChatMessage {
     bool synced = 13;
     bool pinned = 15;
     bool unreadReaction = 16;
-    string lastUnreadReactionOrderId = 17;
```
KirillSto, line 1641: Maybe we should mark it as deprecated? Someone could add 17th field by accident in the future

deff7: no one uses it, and we actually can change protocol any time we want, because it's used only by local clients

KirillSto (approved):

PR #3005, opened 2026-03-09
GO-6612 Add single-change indent operation for list blocks

KirillSto:
## Summary
- Add indent operation handling in `Move()` that keeps children at the original parent level when a block is indented into a sibling
- When a single block with children is moved to `Inner` position of a sibling, children are unlinked from the block and inserted at the block's original position in the parent
- This ensures the entire indent is a single change for correct undo behavior and preserves visual block ordering
- Add comprehensive tests covering: basic indent with children, indent without children, last block indent, non-sibling moves, multi-block moves, and numbered list items

## Test plan
- [x] Unit tests for indent with children — children moved to original parent level
- [x] Unit tests for indent without children — no special behavior
- [x] Unit tests for indent of last block — children appended to parent
- [x] Unit tests for non-sibling target — children stay with block
- [x] Unit tests for multi-block move — no indent logic applied
- [x] Unit tests for numbered list items — correct ordering preserved

### DIFF
--- a/core/block/editor/basic/basic.go
+++ b/core/block/editor/basic/basic.go
@@ -104,9 +104,8 @@ type outdentOpContext struct {
 }
 
 type indentOpContext struct {
+	blockId     string   // the block being indented
 	childrenIds []string // copy of block's children before the move
-	nextSibling string   // sibling after the block in original parent, or empty
-	parentId    string   // original parent id
 }
 
 func NewBasic(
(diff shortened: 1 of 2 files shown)


core/block/editor/basic/basic.go
```diff
@@ -387,6 +407,69 @@ func processOutdentOperation(opCtx *outdentOpContext, srcState *state.State, tar
 	return nil
 }
 
+func collectIndentContext(srcState *state.State, block simple.Block, targetBlockId string) *indentOpContext {
```
deff7, line 410: Maybe create an indent context only if we have a proper `pos`?

Like this check in `processIndentOperation`:
```go
if pos != model.Block_Inner && pos != model.Block_InnerFirst {
		return nil
	}
```

KirillSto: Done!

deff7 (approved):

PR #3037, opened 2026-03-23
GO-6752 Add homepage setting with migrations and fallback

KirillSto:
# GO-6752 Channel Unification: Homepage Setting

## Summary

Implements a synchronized space-level homepage setting that determines which object opens by default when a user enters a space. Deprecates `SpaceUxType` in favour of a new `SpaceType` enum derived from the space header.

### New `homepage` relation & `SetHomepage` command
- Add `homepage` relation (longtext, read-write) to store the default opening object per space (`"widgets"`, object ID, or empty string)
- Add `Rpc.Workspace.SetHomepage` command that sets homepage via `SetSpaceInfo`
- Remove old `ObjectWorkspaceSetDashboard` command and `SetWorkspaceDashboardId` method

### New `spaceType` relation & `SpaceType` protobuf enum
- Add `model.SpaceType` enum (`SpaceTypeUnknown`, `SpaceTypeRegular`, `SpaceTypeTech`, `SpaceTypeChat`, `SpaceTypeOneToOne`)
- Add `spaceType` relation (number, derived, read-only) set during `Workspaces.Init` from space header
- Add `spacedomain.SpaceType.ToModel()` converter and `ReadSpaceTypeFromHeader()` helper
- Expose `SpaceType()` on `smartblock.Space` / `clientspace.Space` interfaces, read from storage in `BuildSpace`
- Propagate `spaceType` through `SpaceDescription`, `SpaceView`, and invite payloads

### Homepage migration (`sourceimpl.migrateHomepage`)
- Runs during `buildState` for workspace objects (not as a versioned state migration) so it applies on every open
- Chat & OneToOne spaces: derive chat object ID → set as `homepage`
- Data spaces: copy `spaceDashboardId` → `homepage`, migrating `lastOpened` to `widgets`
- Skips migration if `homepage` is already set

### Homepage fallback on object deletion
- `unsetHomepageIfNeeded` in `DeleteObjectByFullID`: resets homepage to empty when the homepage object is deleted

### Default homepage changes
- Change default homepage from `lastOpened` to `widgets` across all creation paths
- Extract `HomepageWidgets` and `HomepageLastOpened` constants into `util/constant`
- Update built-in object injection and use-case validator to use `homepage` relation and `widgets` default

### `WorkspaceCreate` / `WorkspaceOpen` use `SpaceType`
- `WorkspaceCreate` validates via `model.SpaceType` instead of `SpaceUxType`; restricts creation of Tech/Chat spaces
- `WorkspaceOpen` uses `SpaceType` for OneToOne chat init instead of `shareableStatus` + `SpaceUxType`

### OneToOne space creation
- Derive chat ID before setting workspace details; set `homepage` to chat ID instead of `spaceDashboardId: lastOpened`
- Pass `SpaceType` alongside `SpaceUxType` in space description

### Deprecate `SpaceUxType_Chat`
- Mark `SpaceUxType_Chat` as deprecated in proto
- `isOneToOne()` on `Workspaces` now checks via `Space().IsOneToOne()` instead of reading `SpaceUxType` detail

## Test plan
- [x] Unit tests for `unsetHomepageIfNeeded` (3 cases: matching homepage, non-matching, empty)
- [x] Unit tests for migration v3 (4 cases: Chat, OneToOne, Data+lastOpened, Data+custom)
- [ ] Manual verification of space creation and opening flows
- [ ] Manual verification of homepage migration on existing spaces


core/block/detailservice/set_details.go
```diff
@@ -276,3 +257,28 @@ func (s *service) triggerFileGCOnArchive(spaceId string, objectIds []string, isA
...
+			return nil
+		}
+		// Check if it's a known constant
+		if constant.IsHomepageConstant(homepage) {
+			return nil
+		}
+		// Otherwise, assume it's an object ID - just validate it's not empty
+		// (full object ID validation would require checking if the object exists,
+		// which is expensive and may not be necessary here)
+		if len(homepage) > 0 {
```
deff7: `homepage != ""` is an idiomatic way to do that, but it's always not empty here, as you check it there

```
if homepage == "" {
			// Empty homepage is valid
			return nil
		}
```

core/block/detailservice/set_details.go
```diff
@@ -276,3 +257,28 @@ func (s *service) triggerFileGCOnArchive(spaceId string, objectIds []string, isA
 		}(objId)
 	}
 }
+
+func validateSpaceDetail(k domain.RelationKey, v domain.Value) error {
+	switch k {
+	case bundle.RelationKeyHomepage:
```
deff7: you should move this check up to the call stack and rename validateSpaceDetail to validateHomepage

core/block/detailservice/set_details.go
```diff
@@ -8,30 +8,29 @@ import (
...
+	"github.com/anyproto/anytype-heart/util/constant"
 	"github.com/anyproto/anytype-heart/util/slice"
 )
 
-var ErrUnexpectedBlockType = errors.New("unexpected block type")
-
 func (s *service) SetSpaceInfo(spaceId string, details *domain.Details) error {
-	ctx := context.TODO()
-	spc, err := s.spaceService.Get(ctx, spaceId)
+	spc, err := s.spaceService.Get(context.Background(), spaceId)
```
deff7: use service context for that

core/workspace.go
```diff
@@ -37,31 +36,35 @@ func (mw *Middleware) WorkspaceCreate(cctx context.Context, req *pb.RpcWorkspace
 		spaceId        string
 		startingPageId string
 	)
+
+	// TODO: GO-6752 remove this hack when clients transfer from UXType to SpaceType
+	deriveSpaceTypeIfNeeded(req.Details)
```
deff7: maybe put this into the block service?

core/workspace.go
```diff
@@ -37,31 +36,35 @@ func (mw *Middleware) WorkspaceCreate(cctx context.Context, req *pb.RpcWorkspace
...
-		} else {
-			spaceUxType = model.SpaceUxType(pbtypes.GetInt64(req.GetDetails(), bundle.RelationKeySpaceUxType.String()))
-			if spaceUxType.String() == "" {
-				return errors.New("unknown space ux type")
-			} else if spaceUxType == model.SpaceUxType_None {
-				return errors.New("space ux type cannot be None")
-			}
-		}
-		if spaceUxType == model.SpaceUxType_Chat || spaceUxType == model.SpaceUxType_OneToOne {
+		spaceType := model.SpaceType(pbtypes.GetInt64(req.GetDetails(), bundle.RelationKeySpaceType.String())) // nolint:gosec
```
deff7: and everything here should go to the service

space/spacedomain/type.go
```diff
@@ -9,4 +17,51 @@ const (
 	SpaceTypeOneToOne SpaceType = "anytype.onetoone"
 )
 
+func (t SpaceType) ToModel() model.SpaceType {
```
deff7: maybe call it `ToEnum()`?

util/constant/constant.go
```diff
@@ -6,6 +6,16 @@ const ProfileFile = "profile"
 
 const SvgExt = ".svg"
 
+const (
```
deff7: maybe put it into domain?

deff7 (changes requested):

core/block/detailservice/set_details.go
```diff
@@ -276,3 +257,28 @@ func (s *service) triggerFileGCOnArchive(spaceId string, objectIds []string, isA
 		}(objId)
 	}
 }
+
+func validateSpaceDetail(k domain.RelationKey, v domain.Value) error {
+	switch k {
+	case bundle.RelationKeyHomepage:
+		homepage := v.String()
+		if homepage == "" {
```
requilence: in this case any other domain.Value beside String will also return empty string. I guess it's better to check the actual type

core/block/import/common/objectcreator/objectcreator.go
```diff
@@ -322,26 +322,38 @@ func (oc *ObjectCreator) deleteFile(spaceId string, hash string) {
...
+	if homepage != "" {
 		details = append(details, domain.Detail{
-			Key:   bundle.RelationKeySpaceDashboardId,
-			Value: domain.StringList(ids),
+			Key:   bundle.RelationKeyHomepage,
+			Value: domain.String(homepage),
 		})
+	} else {
+		// Fallback to spaceDashboardId for old exports
+		ids := combinedDetails.GetStringList(bundle.RelationKeySpaceDashboardId)
```
requilence, line 339: At which point it will be converted to the new homepage?

core/block/create.go
```diff
@@ -169,22 +169,48 @@ func (s *Service) CreateWorkspace(ctx context.Context, req *pb.RpcWorkspaceCreat
...
 	err = cache.Do(s, predefinedObjectIDs.Workspace, func(b basic.DetailsSettable) error {
 		details := make([]domain.Detail, 0, len(req.Details.GetFields()))
+		hasHomepage := false
+		hasUxType := false
 		for k, v := range req.Details.GetFields() {
 			details = append(details, domain.Detail{
 				Key:   domain.RelationKey(k),
 				Value: domain.ValueFromProto(v),
 			})
+			if k == bundle.RelationKeyHomepage.String() {
```
requilence: I guess some clients can send an empty or incorrect(NullValue or any other type)  homepage here. And we assume that they set it. Should we explicitly check non-empty string value?

core/block/import/common/objectcreator/objectcreator.go
KirillSto: It won't. Homepage will be empty in this case, so users can choose any other option in Space settings

core/block/detailservice/set_details.go
KirillSto: I have added emptiness check to the validation

core/block/detailservice/set_details.go
KirillSto: I have moved both check and method to block.Svc

space/spacedomain/type.go
KirillSto: Done

core/block/detailservice/set_details.go
KirillSto: Yes, I have reused validateHomepage func in space creation too

core/block/create.go
KirillSto: I have reused validateHomepage func

core/block/create.go
```diff
@@ -227,11 +235,46 @@ func (s *Service) CreateWorkspace(ctx context.Context, req *pb.RpcWorkspaceCreat
...
+		chatInitErr := s.SpaceInitChat(ctx, newSpace.Id(), true)
+		if chatInitErr != nil {
+			log.With("error", chatInitErr).Warn("failed to init space level chat")
+		}
 	}
 
 	return newSpace.Id(), startingPageId, err
 }
 
+func deriveSpaceTypeIfNeeded(details *types.Struct) {
```
deff7: maybe you could use domain.Details here?

core/block/service.go
```diff
@@ -847,3 +847,29 @@ func removeDescriptionFromRecommended(typeId string, details *domain.Details, sp
 		return nil
 	})
 }
+
+func (s *Service) SpaceSetHomepage(spaceId string, homepage string) error {
+	if err := s.validateHomepage(spaceId, homepage); err != nil {
```
deff7, line 852: why do you validate it here and not in SetSpaceInfo?

core/block/service.go
```diff
@@ -847,3 +847,29 @@ func removeDescriptionFromRecommended(typeId string, details *domain.Details, sp
 		return nil
 	})
 }
+
+func (s *Service) SpaceSetHomepage(spaceId string, homepage string) error {
```
deff7, line 851: don't add new methods to the block service

core/block/create.go
KirillSto: Done

core/block/service.go
KirillSto: I have added validation back to SetSpaceInfo

deff7 (approved):

KirillSto:
Replaced by #3041 — single signed commit on a fresh branch.

PR #3038, opened 2026-03-23
GO-6954 Index number, url, email, phone formats and readonly relations

KirillSto:
## Summary
- Expand fulltext indexer format filter to also index `number`, `url`, `email`, and `phone` relation formats (previously only `shorttext`/`longtext`)
- Handle number values correctly using `Float64()` + `strconv.FormatFloat()` since `GetString()` returns empty for float64 values
- Remove the `ReadOnly` skip so readonly but non-hidden relations like `fileExt` get indexed
- Add `isIndexableFormat()` helper for cleaner format checking

## Test plan
- [x] Unit tests pass (`go test ./core/indexer/... -run TestPrepareSearchDocument -v`)
- [x] Updated `WrongFormat` test to use a truly non-indexable format (date)
- [x] Added tests for email indexing, integer numbers, decimal numbers, and readonly `fileExt`
- [x] Manual verification: search for number values (ISO, height, focal ratio) and file extensions in sets

### DIFF
--- a/core/indexer/fulltext.go
+++ b/core/indexer/fulltext.go
@@ -306,28 +306,31 @@ func (i *indexer) prepareSearchDocs(ctx context.Context, object domain.FullTextQ
 
 		for _, key := range sb.AllRelationKeys() {
 			format, err := i.formatFetcher.GetRelationFormatByKey(object.SpaceId, key)
-			if err == nil && format != model.RelationFormat_shorttext && format != model.RelationFormat_longtext {
+			if err == nil && !isIndexableFormat(format) {
 				continue
 			}
-			val := sb.Details().GetString(key)
-			if val == "" {
-				val = sb.LocalDetails().GetString(key)
+			var val string
+			if format == model.RelationFormat_number {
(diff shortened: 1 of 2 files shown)


core/indexer/fulltext.go
```diff
@@ -304,30 +304,30 @@ func (i *indexer) prepareSearchDocs(ctx context.Context, object domain.FullTextQ
 			return nil
 		}
 
-		for _, key := range sb.AllRelationKeys() {
+		for key, value := range sb.CombinedDetails().Iterate() {
```
deff7: why use combined details here?

KirillSto: We need it for snippet 100%, but it is really thin scenario. Let me investigate what other local properties we may need to index

deff7: combined details wastes memory

KirillSto: Yes, only snippet is needed here, so I have switched CombinedDetails to Details and splitted search docs preparation to different funcs

deff7 (approved):

PR #3041, opened 2026-03-24
GO-6752 Add homepage setting with migrations and fallback

KirillSto:
# GO-6752 Channel Unification: Homepage Setting

## Summary

Implements a synchronized space-level homepage setting that determines which object opens by default when a user enters a space. Deprecates `SpaceUxType` in favour of a new `SpaceType` enum derived from the space header.

### New `homepage` relation & `SetHomepage` command
- Add `homepage` relation (longtext, read-write) to store the default opening object per space (`"widgets"`, object ID, or empty string)
- Add `Rpc.Workspace.SetHomepage` command that sets homepage via `SetSpaceInfo`
- Remove old `ObjectWorkspaceSetDashboard` command and `SetWorkspaceDashboardId` method

### New `spaceType` relation & `SpaceType` protobuf enum
- Add `model.SpaceType` enum (`SpaceTypeUnknown`, `SpaceTypeRegular`, `SpaceTypeTech`, `SpaceTypeChat`, `SpaceTypeOneToOne`)
- Add `spaceType` relation (number, derived, read-only) set during `Workspaces.Init` from space header
- Add `spacedomain.SpaceType.ToModel()` converter and `ReadSpaceTypeFromHeader()` helper
- Expose `SpaceType()` on `smartblock.Space` / `clientspace.Space` interfaces, read from storage in `BuildSpace`
- Propagate `spaceType` through `SpaceDescription`, `SpaceView`, and invite payloads

### Homepage migration (`sourceimpl.migrateHomepage`)
- Runs during `buildState` for workspace objects (not as a versioned state migration) so it applies on every open
- Chat & OneToOne spaces: derive chat object ID → set as `homepage`
- Data spaces: copy `spaceDashboardId` → `homepage`, migrating `lastOpened` to `widgets`
- Skips migration if `homepage` is already set

### Homepage fallback on object deletion
- `unsetHomepageIfNeeded` in `DeleteObjectByFullID`: resets homepage to empty when the homepage object is deleted

### Default homepage changes
- Change default homepage from `lastOpened` to `widgets` across all creation paths
- Extract `HomepageWidgets` and `HomepageLastOpened` constants into `util/constant`
- Update built-in object injection and use-case validator to use `homepage` relation and `widgets` default

### `WorkspaceCreate` / `WorkspaceOpen` use `SpaceType`
- `WorkspaceCreate` validates via `model.SpaceType` instead of `SpaceUxType`; restricts creation of Tech/Chat spaces
- `WorkspaceOpen` uses `SpaceType` for OneToOne chat init instead of `shareableStatus` + `SpaceUxType`

### OneToOne space creation
- Derive chat ID before setting workspace details; set `homepage` to chat ID instead of `spaceDashboardId: lastOpened`
- Pass `SpaceType` alongside `SpaceUxType` in space description

### Deprecate `SpaceUxType_Chat`
- Mark `SpaceUxType_Chat` as deprecated in proto
- `isOneToOne()` on `Workspaces` now checks via `Space().IsOneToOne()` instead of reading `SpaceUxType` detail

## Test plan
- [x] Unit tests for `unsetHomepageIfNeeded` (3 cases: matching homepage, non-matching, empty)
- [x] Unit tests for migration v3 (4 cases: Chat, OneToOne, Data+lastOpened, Data+custom)
- [x] Manual verification of space creation and opening flows
- [x] Manual verification of homepage migration on existing spaces


core/block/create.go
```diff
@@ -169,44 +175,103 @@ func (s *Service) CreateWorkspace(ctx context.Context, req *pb.RpcWorkspaceCreat
...
-			details = append(details, domain.Detail{
-				Key:   domain.RelationKey(k),
-				Value: domain.ValueFromProto(v),
-			})
+		details := make([]domain.Detail, 0, spaceDetails.Len())
+		hasHomepage := false
+		hasUxType := false
+		for k, v := range spaceDetails.Iterate() {
+			details = append(details, domain.Detail{Key: k, Value: v})
+			if k == bundle.RelationKeyHomepage && v.String() != "" {
```
deff7: Maybe use spaceDetails.Has? Iterate + check key is fragile construction with multiple responsibilities

core/block/create.go
```diff
@@ -169,44 +175,103 @@ func (s *Service) CreateWorkspace(ctx context.Context, req *pb.RpcWorkspaceCreat
...
+			spaceType = model.SpaceType_SpaceTypeRegular
+		}
+		details.SetInt64(bundle.RelationKeySpaceType, int64(spaceType))
+	}
+}
+
+func validateSpaceType(spaceType model.SpaceType) error {
+	switch spaceType {
+	case model.SpaceType_SpaceTypeUnknown:
+		return errors.New("space ux type cannot be Unknown")
```
deff7: almost the same error as in `default` case

core/domain/value.go
```diff
@@ -33,6 +33,27 @@ const (
...
+	case ValueTypeString:
+		return "string"
+	case ValueTypeFloat:
+		return "float"
+	case ValueTypeStringList:
+		return "string list"
+	case ValueTypeMap:
+		return "map"
+	default:
+		return "unknown"
```
deff7: invalid

core/block/create.go
KirillSto: Yeah, I have used Has both for Homepage and UXtype. Details are not so big to save time on this operations

core/block/create.go
KirillSto: Removed Unknown branch

deff7 (approved):

PR #3053, opened 2026-03-26
GO-7111 Fix redundant participant re-indexing on every startup

requilence:
## Problem

On every app restart, **all participant objects were re-indexed** (~1900 FT queue additions for a 57-space account). Two root causes:

1. **Empty parent state from `StaticSource`**: `StaticSource` created a blank state, so every `Init → Apply` saw all details/blocks as new — producing a full diff even when nothing changed.
2. **Identity `iconImage` and `globalName` ping-pong**: `handleOwnProfileDetails` copied the raw file object ID for `iconImage`, but the identity repo stores the file CID. On restart the values flipped back and forth, triggering updates across all spaces. `globalName` was simply missing from the copied keys.

## Solution

- **Dedicated `participantSource`** that pre-populates the source doc with objectStore details, defaults, template blocks, and relation links — `Init → Apply` produces zero diff for unchanged participants.
- **Simplified `participant.Init`** to pure delegation — all setup owned by the source.
- **Fixed `handleOwnProfileDetails`** to resolve `iconImage` to file CID via `prepareIconImageInfo` and include `globalName` in copied keys.
- **Removed misleading `ops/sec` log** from `GenerateFTQueueCounter` (counted index attempts, not actual FT queue writes).

## Tested

- [x] No `AddToIndexQueue` for participants on idle restart (57-space account)
- [x] Init Apply produces 0 msgs / 0 changes for existing participants
- [x] No `ObjectRelationsAmend` or `ObjectDetailsAmend` events on Init
- [x] No icon/globalName diffs on subsequent restarts (ping-pong eliminated)
- [x] Identity callbacks return consistent file CID for iconImage
- [x] Profile description change propagates once per space, no side effects
- [x] Idle state after profile change produces no further diffs
- [x] Existing participant unit tests pass
- [x] Participant details are correct in UI after restart
- [x] Participant search works (fulltext index intact)
- [x] New account (no objectStore data for participants)
- [x] Space join — new participant created correctly
- [x] Icon change propagates correctly to all spaces


core/block/source/sourceimpl/participant.go
```diff
@@ -0,0 +1,120 @@
...
+
+func (p *participantSource) Type() smartblock.SmartBlockType {
+	return smartblock.SmartBlockTypeParticipant
+}
+
+func (p *participantSource) ReadOnly() bool {
+	return true
+}
+
+func (p *participantSource) ReadDoc(_ context.Context, _ source.ChangeReceiver, _ bool) (doc state.Doc, err error) {
```
deff7: you can use the same mechanism like in chats

see

```
	// Set object type and layout details for this fake state
	st := ctx.Doc.(*state.State)
	st.SetObjectTypeKey(s.typeKey)
	st.SetDetailAndBundledRelation(bundle.RelationKeyLayout, domain.Int64(int64(s.layout)))
	st.SetDetailAndBundledRelation(bundle.RelationKeyIsHidden, domain.Bool(false))
```

requilence: ok, I refined the approach to be like in the chat objects

deff7 (approved):

PR #3057, opened 2026-03-27
GO-7116 Fix deadlock on app close in inboxclient

requilence:
## Summary

- Add `componentCtx` to inboxclient that gets cancelled on `Close()`, ensuring all operations (both periodic and external callback) are cancellable on shutdown
- Propagate context through `fetchMessages`/`getOffset`/`setOffset` instead of hardcoded `context.Background()`

## Problem

`inboxclient.Close()` deadlocks when the coordinator connection is stale. The periodic sync callback (`checkMessages`) ignores its context parameter and calls `fetchMessages()` which uses `context.Background()` for the `InboxFetch` RPC. When the coordinator doesn't respond, the RPC blocks forever, and `periodicCall.Close()` waits indefinitely for the loop to finish.

**Deadlock chain:**
1. `app.Close()` → `inboxclient.Close()` → `periodicCheck.Close()` → cancels loop context, waits on `<-loopDone`
2. Periodic loop executing `checkMessages(ctx)` → ignores ctx → `fetchMessages()` → `InboxFetch(context.Background())` → blocked forever on stale DRPC stream

## Test plan
- [x] All existing inboxclient tests pass
- [x] `ReceiveNotify` (external callback) now uses `componentCtx` instead of `context.Background()` — also cancellable on close


core/inboxclient/inboxclient.go
```diff
@@ -242,6 +249,7 @@ func (s *inboxclient) InboxAddMessage(ctx context.Context, receiverPubKey crypto
 	return s.inboxClient.InboxAddMessage(ctx, receiverPubKey, message)
 }
 func (s *inboxclient) Close(_ context.Context) (err error) {
+	s.componentCtxCancel()
```
deff7: check for nil just in case

deff7 (approved):

PR #3062, opened 2026-03-30
GO-7133 Add _final_score field blending BM25 score with recency decay and name boost

requilence:
## Summary

- Adds `_final_score` to `ObjectSearchWithMeta` fulltext results, computed as `ln(1+bm25) + recency + name_boost`
- Recency uses a 30-day exponential half-life on the more-recent of `lastOpenedDate`/`lastModifiedDate` (clamped to [0,1]); name_boost is 1.0 when the match landed in the `name` relation field
- `_score` (raw BM25) is unchanged — clients opt into recency-aware ranking by sorting on `_final_score`

## Formula

```
recency    = max(decay(lastOpenedDate), decay(lastModifiedDate))
decay(ts)  = min(1, exp(-ln(2) / 30 × age_in_days))
name_boost = 1.0 if match.relationKey == "name" else 0.0

_final_score = ln(1 + bm25) + recency + name_boost
```

Log-compressing BM25 keeps recency and name_boost as equal-weight additive signals — each contributes at most +1.0 regardless of corpus size or index growth.

## Test Plan

- [x] Unit tests for `recencyDecay` and `ComputeFinalScore` in `pkg/lib/database/database_test.go` (11 sub-tests covering zero sentinel, half-life, future clamp, nil-guard, all signal combinations)
- [x] Integration tests for `_final_score` in `pkg/lib/localstore/objectstore/spaceindex/queries_fulltext_test.go` (baseline, name-boost delta, recency ordering)
- [x] Verify `ObjectSearchWithMeta` response includes `_final_score` when `"_final_score"` is in `keys`
- [x] Confirm exact-match outlier still ranks #1 and flat-plateau results are sorted by recency

Closes GO-7133

### DIFF
--- a/pkg/lib/database/database.go
+++ b/pkg/lib/database/database.go
@@ -19,9 +19,7 @@ import (
 var log = logging.Logger("anytype-database")
 
 const (
-	RecordScoreField      = "_score"
-	RecordFinalScoreField = "_final_score"
-	secondsPerDay         = 86400.0
+	secondsPerDay = 86400.0
 )
 
 type Record struct {
@@ -172,13 +170,13 @@ func injectDefaultOrder(qry Query, sorts []SortRequest) []SortRequest {
 	}
 
 	for _, sort := range sorts {
-		if sort.RelationKey == RecordScoreField || sort.RelationKey == RecordFinalScoreField {
+		if sort.RelationKey == bundle.RelationKey_score || sort.RelationKey == bundle.RelationKey_final_score {
 			hasScoreSort = true
 		}
 	}
 
 	if !hasScoreSort {
-		sorts = append([]SortRequest{{RelationKey: RecordFinalScoreField, Type: model.BlockContentDataviewSort_Desc}}, sorts...)
+		sorts = append([]SortRequest{{RelationKey: bundle.RelationKey_final_score, Type: model.BlockContentDataviewSort_Desc}}, sorts...)
 	}
 
 	return sorts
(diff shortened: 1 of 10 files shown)


pkg/lib/database/database.go
```diff
@@ -394,3 +397,34 @@ type Filters struct {
...
+// Formula: ln(1+bm25) + recency + nameBoost
+// Log-compressing the BM25 score keeps recency and nameBoost as equal-weight additive signals.
+func ComputeFinalScore(bm25Score float64, details *domain.Details, nameMatch bool) float64 {
+	if details == nil {
+		return math.Log1p(bm25Score)
+	}
+	now := time.Now().Unix()
+	lastOpened := details.GetInt64(bundle.RelationKeyLastOpenedDate)
+	lastModified := details.GetInt64(bundle.RelationKeyLastModifiedDate)
+	recency := math.Max(recencyDecay(now, lastOpened), recencyDecay(now, lastModified))
```
KirillSto: var recency float64
	if lastOpened < lastModified {
		recency = recencyDecay(now, lastOpened)
	} else {
		recency = recencyDecay(now, lastModified)
	}

pkg/lib/database/database.go
```diff
@@ -17,7 +19,9 @@ import (
 var log = logging.Logger("anytype-database")
 
 const (
-	RecordScoreField = "_score"
+	RecordScoreField      = "_score"
+	RecordFinalScoreField = "_final_score"
```
KirillSto: Let's use names of bundled relations

pkg/lib/database/database.go
requilence: yes but opposite condition `if lastOpened > lastModified`

KirillSto (approved):

PR #3068, opened 2026-04-01
GO-7152 Context-aware GC stage 2: extend to all user-content objects

requilence:
### Core GC changes
- Adds \`GCEligibleLayouts\` whitelist (\`core/domain/layout.go\`) covering file layouts plus user-content layouts (\`basic\`, \`profile\`, \`todo\`, \`set\`, \`note\`, \`bookmark\`, \`collection\`)
- Generalizes \`FileGC\` → \`ObjectGC\` (\`core/block/objectgc/\`): all GC queries now use \`GCEligibleLayouts\` so basic pages/notes/etc. created via \`BlockLinkCreateWithObject\` are archived when their parent is archived or their link is removed
- Non-file objects always use \`skipBin=false\` (archive, never permanent delete)
- Removes the buggy \`sessionCreatedLinks\` tracking from \`smartblock\` — all link removals now archive orphaned children
- Sets \`createdInContext\`/\`createdInContextRef\` on objects created via \`CreateLinkToTheNewObject\` and \`BlockListConvertToObjects\` so the GC can identify them as children
- Query 2 of \`collectOrphanedObjects\` now requires \`createdInContext\` to be non-empty — objects imported without context tracking (e.g. usecase icon images) are no longer auto-archived

### BFS collect-then-act
- Replaces the original recursive GC with a BFS traversal over the \`createdInContext\` tree
- Uses a visited/pending set so sibling cross-references don't falsely block each other from being archived
- Iterative eviction loop: when a candidate has an external active backlink, it is removed and marked active so dependent siblings are also evicted in subsequent iterations
- Cycle-safe: visited set prevents re-queuing already-seen nodes
- Regression tests for cascade exclusion, sibling cross-reference, multi-level restore, and cycle termination

### Pure GC — callers own writes
- \`CheckObjectsOnObjectArchived\` is now a pure read: returns the list of IDs to archive/restore without performing any writes
- \`archiveOrphanedObjects\` and \`restoreObjectsOnUnarchive\` are pure helpers (no archiver calls)
- \`ObjectArchiver.SetListIsArchivedNoGC\` removed from the GC interface — no longer needed internally
- \`SetListIsArchivedNoGC\` added to \`detailservice.Service\` interface for callers that need it
- \`setIsArchivedForObjects\` merges explicit \`objectIds\` + GC-collected \`gcIds\` into a single \`modifyArchiveLinks\` call (true 1-tx archive)
- \`SetIsArchived\` routes through \`setIsArchivedForObjects\` for the same benefit
- \`delete.go\` collects returned \`gcIds\` and calls \`SetListIsArchivedNoGC\` to archive orphaned children on object deletion

### Session events
- Adds \`EventObjectAutoArchive\` proto message: emitted when GC auto-archives objects alongside a user archive action
- Adds \`EventObjectAutoRestore\` proto message: emitted when GC auto-restores objects alongside a user unarchive action
- \`smartblock.Apply\` async goroutine emits auto-archive/restore events after link-change GC completes
- \`FilterExplicitIds\` strips IDs already in the original user request from both event types, preventing double-reporting when a listed object is also a GC child
- Session context removed from \`ObjectGC\` interface — callers own event emission

### Naming
- \`core/files/filegc\` → \`core/block/objectgc\`
- \`triggerFileGCOnArchive\` → \`triggerGCOnArchive\`
- \`performFileGC\` → \`performGCOnLinksRemoval\`
- \`restoreArchivedFilesOnLinksAdded\` → \`restoreObjectsOnLinksAdded\`
- \`CheckObjectsOnLinksRemoval\` → \`ArchiveOrphansOnLinksRemoval\`
- \`CheckObjectsOnLinksRestored\` → \`RestoreOrphansOnLinksAdded\`

## Test Plan
- [x] \`core/domain/layout_test.go\` — \`GCEligibleLayouts\` contains all expected layouts, excludes system ones
- [x] \`core/block/objectgc/objectgc_test.go\` — tests covering: parent archive/unarchive cascade, backlinker safety gate, no-\`createdInContext\` exclusion, BFS multi-level, cycle termination, sibling cross-reference, cascade exclusion regression, \`FilterExplicitIds\`, returned-IDs assertions
- [x] \`core/block/create_test.go\` — \`setCreatedInContext\` sets both relation keys correctly
- [x] \`core/block/editor/smartblock/objectgclinks_test.go\` — GC triggered on link removal/addition in smartblock Apply
- [x] \`go build ./...\` — no errors


docs/superpowers/specs/2026-04-01-context-aware-gc-stage2-design.md
```diff
@@ -0,0 +1,112 @@
+# Context-Aware GC — Stage 2: All Objects
```
deff7: do we need to include this file?

deff7 (approved):

deff7 (approved):

PR #3081, opened 2026-04-08
GO-6758 Optimize full chat messages reindex with batched fetching

KirillSto:
## Summary
- Replace single `GetMessages` call (which loads all messages into memory at once) with batched iteration of 100 messages per batch in `prepareChatSearchDocsAll`
- Extract `chatMessageToSearchDoc` helper to eliminate code duplication between batch and single-message paths

## Linear
GO-6758

## Test plan
- [x] Existing `TestPrepareSearchDocs_ChatObject/all_messages` passes
- [x] Existing `TestPrepareSearchDocs_ChatObject/messages_by_orderId` passes
- [x] New `TestPrepareSearchDocs_ChatObject/all_messages_batched` test with 150 messages (exceeds batch size) verifies all messages are fetched in correct order
- [x] All chat repository tests pass (`go test ./core/block/chats/chatrepository/...`)
- [x] Full indexer test suite passes (`go test ./core/indexer/...`)


requilence:
Drive-by note while working on GO-7316 (branch `go-7316-chat-fulltext-allorder-backfill`): this batched fetch is exactly the right direction and we'll want it landed alongside that branch.

One caveat for whenever we revisit: the batching here bounds the **fetch** (≤100 `*Message` held at a time), but `prepareChatSearchDocsAll` still `append`s every message's `SearchDoc` into one returned slice, so peak memory is still **O(chat size)** in `SearchDoc`s (each holds the full message text). For a very large chat that's still a potential OOM/stall on constrained devices.

To make rebuild memory **O(batch)**, we'd flush docs to the autobatcher per batch instead of returning the whole slice (e.g. pass the batcher down, or have `processQueuedObjects` consume a stream/callback rather than a materialized `[]SearchDoc`).

Why it matters more now: GO-7316 bumps `ForceReindexChatsFulltextCounter`, so the full-history `_all` reindex fires **fleet-wide on upgrade** and on every future `ftsVer` bump — i.e. this path goes from rarely-exercised to reliably-exercised. Not blocking this PR; just flagging for the follow-up.

requilence (approved):

PR #3096, opened 2026-04-15
GO-7176 Fix markdown import: scope YAML property resolution by object type

requilence:
## Summary

Fixes three issues that surfaced while importing a folder of markdown files with a bundled `types/` directory of JSON schemas.

### 1. Ambiguous property-name resolution across schemas (main bug)

`SchemaImporter.ResolvePropertyKey` iterated `si.schemas` (a Go map, non-deterministic order) and returned the first relation whose `Name` matched — ignoring which schema/type the current file actually belonged to. When several schemas declare properties with the same display name but different `x-key` (e.g. `References`, `Ideas`, `Pieces`, `Commissions` defined by `mood_board`, `piece`, `idea`, `commission`, `concept`, `reference` types), the resolver could return `piece_references` for a Mood Board file. The value ended up in details under a key that is not in the Mood Board type's RecommendedRelations, so the property never surfaced on the imported object. Which property happened to resolve correctly changed between runs.

Extend `schema.PropertyResolver` with an `objectTypeName` parameter and scope `SchemaImporter.ResolvePropertyKey` / `GetRelationFormat` to the schema whose `Type.Name` matches the file's frontmatter `Object type:`. Falls back to a global scan so cross-type / bundled-relation lookups still work. The parser threads `result.ObjectType` through.

### 2. Tag/status options dropped when schemas have no examples

`createSnapshots` only collected tag/status option values from YAML when `!hasSchemas`, and `CreateRelationOptionSnapshots` only emitted snapshots for `rel.Options` (status) / `rel.Examples` (tag). A schema that declares a tag relation without pre-seeding examples (e.g. `Mood Words` in `mood_board.json`) produced details with option IDs pointing at snapshots that were never created.

Add `SchemaImporter.RegisterOptionValue` and extend `CreateRelationOptionSnapshots` to also emit YAML-registered options, de-duped against the schema-declared set. In `createSnapshots`, route tag/status values through `RegisterOptionValue` when schemas are loaded; the non-schema path is unchanged.

### 3. Silent drop of unresolved object refs

`processObjectProperties` silently discarded object-format property values whose file path could not be resolved. Emit a `log.Warnf` so broken links in imported folders show up in logs.

## Test plan

- [x] `go test ./core/block/import/markdown/ -run TestSchemaImporter_ResolvePropertyKey_ScopedByObjectType`
- [x] `go test ./core/block/import/markdown/ -run TestSchemaImporter_CollectsOptionsFromYAML`
- [x] `go test ./core/block/import/markdown/... ./pkg/lib/schema/...`
- [x] `go test ./core/block/import/...`
- [ ] Manual: re-import the repro fixture folder and confirm the Mood Board object shows `Mood Words` (tactile/raw/intimate), `References` (2), `Ideas` (2), `Pieces` (2), and `Commissions` (1).

### DIFF
--- a/core/block/import/markdown/import.go
+++ b/core/block/import/markdown/import.go
@@ -574,21 +574,35 @@ func (m *Markdown) createSnapshots(
 				}
 				props = append(props, yamlRelations[prop.Name].Key)
 
-				// Collect option values for non-schema imports
-				if !hasSchemas && (prop.Format == model.RelationFormat_status || prop.Format == model.RelationFormat_tag) {
-					if yamlRelationOptions[prop.Key] == nil {
-						yamlRelationOptions[prop.Key] = make(map[string]string)
-					}
-
-					// Collect values
-					switch prop.Format {
-					case model.RelationFormat_status:
-						if val := prop.Value.String(); val != "" {
-							yamlRelationOptions[prop.Key][val] = ""
+				// Collect option values so we can create snapshots for any
+				// options the schema did not pre-declare (examples/enum).
+				// Non-schema imports still rely on the yamlRelationOptions
+				// map which is materialised below.
+				if prop.Format == model.RelationFormat_status || prop.Format == model.RelationFormat_tag {
+					if hasSchemas {
+						switch prop.Format {
+						case model.RelationFormat_status:
+							if val := prop.Value.String(); val != "" {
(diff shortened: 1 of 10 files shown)


requilence:
@claude review

zarkone (approved):

PR #3118, opened 2026-04-24
GO-53 feat(import): Giving the Heart a "Pair of Glasses" to see its own Space

barissalihbabacan:
## The "Data Smuggler" Chronicles

First, a confession: I’m a Junior developer with a background in hardware and networking, which means I’m used to things just... connecting. But migrating my life from Obsidian to Anytype felt like a survival horror game. Between my `[[Char]]` links being ghosted and the engine ignoring empty files, I spent my weekend as a "ZIP-injection specialist"—basically a data smuggler—manually hacking ZIP files to convince Anytype that my existing data actually exists.

I’ve been warning you about this "mention-blindness" and the destructive nature of the current serialization layer for a month now in Issue #53, but since the UTF-8 offsets are still shifting and my mentions are still dying, I decided to take the scalpel and perform the "Heart" surgery myself.

## "Trust our Code, not our Words"

Your manifesto states: *"Trust our code, not our words."* I took that literally. I looked at the code, verified the logic, and found that it was failing the very autonomy and "Local-first" philosophy it promises to protect.

Right now, the import engine treats a temporary ZIP bundle or an API stream as the only reality. It completely ignores the actual Single Source of Truth (SSOT): The Space itself. Having a "Local-first" engine act completely blind to its own local database feels like an architectural prank. I love the strictness (congrats on manifesting the soul of TypeScript into a whole app philosophy! 👏), but maybe we can be a little less hermit-like? If the data is already in the Space, why are we pretending it's missing?

### What this PR actually does (to save our collective sanity):

Instead of the engine immediately throwing its hands up and shouting `_missing_object` when a reference is missing from an import bundle, this PR introduces a "groundbreaking" concept: **Checking the local database first.**

* **The Logic:** In `core/block/import/common/common.go`, I've updated `handleTextBlock()` (and seeded existing IDs) to check if the `mark.Param` is a valid CID before giving up. 
* **The Result:** If the object already exists in the Space, we preserve its original ID in the parsed mention, maintaining the native Relation/Mention UI. If it’s truly nowhere to be found, then it gets the standard `addr.MissingObject` treatment.
* **Impact:** This fixes the broken link drama for ZIP, API, and MCP incremental imports all at once. No more manual ZIP-hacking required to simply mention an existing Character or Topic.

## Conclusion

I’ve bridged the gap between the Import Engine and the Space. It turns out that letting the engine talk to the database isn't as dangerous as the "Separation of Concerns" police might think—it actually makes the app usable for power users and automation scripts.

I’m just a Junior dev who spent 5 minutes thinking about a "wild" idea that apparently 30+ seniors missed: Looking at the data we already have.

You're welcome. 😉
*Don't hate the player, hate the game.*

**Related Issues:**
Fixes the root cause of logic discussed in #53 (Mention-blindness during object creation/import via API/ZIP).

### DIFF
--- a/core/block/import/common/common.go
+++ b/core/block/import/common/common.go
@@ -249,7 +249,15 @@ func handleTextBlock(oldIDtoNew map[string]string, block simple.Block, st *state
 		}
 		newTarget := oldIDtoNew[mark.Param]
 		if newTarget == "" {
-			newTarget = addr.MissingObject
+			// If the param looks like a valid CID, preserve it as-is.
+			// This allows mentions to reference existing objects in the space
+			// that may not be in the import package (e.g., incremental imports).
+			_, err := cid.Decode(mark.Param)
+			if err == nil {
+				newTarget = mark.Param
+			} else {
+				newTarget = addr.MissingObject
+			}
 		}
 
 		marks[i].Param = newTarget
--- a/core/block/import/processor.go
+++ b/core/block/import/processor.go
@@ -198,9 +198,20 @@ func (p *importProcessor) initConversionFields(converterResponse *common.Respons
 	p.converterResponse = converterResponse
 	p.errors = errors
 	p.oldIDToNew = make(map[string]string, len(converterResponse.Snapshots))
+	p.seedExistingSpaceObjects()
 	p.createPayloads = make(map[string]treestorage.TreeStorageCreatePayload, len(converterResponse.Snapshots))
 	p.relationKeysToFormat = make(map[domain.RelationKey]int32, len(converterResponse.Snapshots))
 	return nil
+}
+
+func (p *importProcessor) seedExistingSpaceObjects() {
+	ids, err := p.deps.objectStore.SpaceIndex(p.request.SpaceId).ListIds()
+	if err != nil {
+		return
+	}
+	for _, id := range ids {
+		p.oldIDToNew[id] = id
+	}
 }
 
 func (p *importProcessor) createObjects(ctx context.Context) (map[string]*domain.Details, string) {


barissalihbabacan:

barissalihbabacan:
recheck

deff7:
Don't you think the tone of your PR description is a little inappropriate? 

Check our https://github.com/anyproto/.github/blob/main/docs/CODE_OF_CONDUCT.md

barissalihbabacan:
> Don't you think the tone of your PR description is a little inappropriate?
> 
> Check our https://github.com/anyproto/.github/blob/main/docs/CODE_OF_CONDUCT.md

@deff7,

Thanks for the reminder. If my tone came across as inappropriate or disrespectful to the team, I sincerely apologize.

My frustration leaked into the description because I literally lost nearly 14 days of my life manually recovering 3,579 broken mentions just to migrate my novel's world-building data from Obsidian to Anytype. The humor and edge in my description were a direct result of experiencing that level of severe UX friction. My critique is aimed at the architecture's strictness, not the people behind it. I am happy to update the PR description to be strictly technical.

However, moving past the tone, I genuinely want to understand the architectural rationale here. Why is the Import Engine intentionally designed to be isolated from the Space's current state?

1. SOC (Separation of Concerns) vs. Isolation of Modules:
I understand the need for SOC—the import engine shouldn't care how the data is stored. But SOC shouldn't mean the engine is forbidden from asking what is already there. Forcing the engine to be completely "context-blind" to the local database results in those 3,500+ broken relations and _missing_object_ spam. It shifts the entire burden of reconciliation onto the user.

2. The SSOT (Single Source of Truth) Paradox:
In a "Local-first" app, the local Space is the ultimate SSOT. By treating a temporary ZIP bundle or API stream as the absolute reality and ignoring the existing local objects, aren't we compromising the very SSOT we are trying to protect?

I’m asking this as an author and power user who pushed the engine to its limits. Is there a specific security or architectural danger to allowing the Import layer to perform a simple CID check against the existing Space to preserve relational integrity?

I'd really appreciate a technical deep dive into this, because this "blindness" is a massive blocker for anyone trying to build or migrate a serious, interconnected database.

PR #3122, opened 2026-04-27
GO-7145 Index Object text marks as outgoing links

ra3orblade:
## Summary
- Backlinks were updating correctly for `BlockContentTextMark_Mention` marks but not for `BlockContentTextMark_Object` marks. Inline-object text marks reference other objects the same way mentions do, but `collectOutgoingLinks` skipped them, so the target object's `backlinks` relation never picked them up.
- Fix in `core/block/editor/smartblock/smartblock.go` — treat `BlockContentTextMark_Object` the same as `BlockContentTextMark_Mention` when collecting outgoing links from text blocks. This matches the existing handling in `core/block/simple/text/text.go` (`FillSmartIds`, `HasSmartIds`, `ReplaceLinkIds`), which already covers both mark types.
- Added a `collect link from text object mark` subtest to `TestSmartBlock_CollectOutgoingLinks` mirroring the existing mention case.

## Test plan
- [x] `go test ./core/block/editor/smartblock/ -run TestSmartBlock_CollectOutgoingLinks -count=1`
- [ ] Manual: create object A with an inline-object mark pointing to object B; confirm B's backlinks relation now includes A.

### DIFF
--- a/core/block/editor/smartblock/smartblock.go
+++ b/core/block/editor/smartblock/smartblock.go
@@ -1482,9 +1482,11 @@ func (sb *smartBlock) collectOutgoingLinks(st *state.State) []OutgoingLink {
 		}
 
 		if text := blockModel.GetText(); text != nil && text.Marks != nil {
-			// Extract mentions from text marks
 			for _, mark := range text.Marks.Marks {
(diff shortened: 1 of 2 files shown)


sashagood:
Tested manually. This does not resolve the issue, so it will be reopened by the team.

PR #3154, opened 2026-05-16
GO-7290 Chat read-counter: drop DiffManager rebuild + bounded unread-candidate watermark

requilence:
## Problem

Chat unread/mention/reaction counters rebuilt the full any-sync `DiffManager`
change graph from storage on every chat open / cold start (`BuildHistoryTree`
+ DAG-ancestor `RemoveBefore`) — O(history) CPU/IO regardless of how little is
actually unread.

A first pass replaced that with an `(OrderId, AddSeq)` watermark, but the
watermark engine itself still did a **full tree-change scan** per counter per
chat: `advance → eachChange → Tree().Storage().GetAfterOrder(ctx, "", …)`
streams *every* change. An Android cold-start CPU profile (60 s window) showed
this at ~13 s / 14 %: ≈324 full scans (≈108 chats × 3 counters) over a
~150 k-row `changes` collection, while almost every chat is fully read —
overwhelmingly wasted work.

## Change

Two layered changes (full history on `go-7290-chat-read-counter`):

**1. DiffManager → `(OrderId, AddSeq)` watermark.** A message is read iff some
synced seen head dominates it on **both** axes:

```
read(X) ⟺ ∃ H ∈ seenHeads : OrderId(X) ≤ OrderId(H) ∧ AddSeq(X) ≤ AddSeq(H)
```

`seenHeads` (synced KV) unchanged; `OrderId`/`AddSeq` resolved locally via
lock-free `Storage().Get`. No change-graph rebuild, no any-sync change, no
persisted cache; `addToDiffManagers`/`updateInDiffManagers` become no-ops.
Includes the cold-start regression fix (one `advance` per counter by merging
per-device seenHeads; delta `onRemove` instead of re-emitting the dominated
set).

**2. Bounded unread-candidate stream — eliminates the watermark's own full
scan.** The dominance engine (`readwatermark.go`) is **unchanged**: same
`dominated` / prune / `marked` / `onRemove` logic, same `advance` signature.
Only the *source* of the `(id, pair)` stream changes — instead of every tree
change, the store enumerates the small set of **currently-unread chat rows**
and AddSeq-resolves each via the same `Storage().Get` point lookup the
watermark already uses:

- `messages`: `read == false`
- `mentions`: `hasMention == true AND mentionRead == false` (hasMention-
  anchored — `mentionRead == false` alone is nearly the whole chat)
- `reactions`: unchanged, legacy full stream (Phase 1)
- new plain `read` and `hasMention` chat-collection indexes;
  `getReadFilter(false)` changed from a non-indexable `query.Not{}` to a
  positive `== false` equality so the index is actually used; candidate set
  bounded by `_o.id ≤ maxFrontierOrderId`; defensive fallback to the legacy
  full stream if the unread set is pathologically large. `chatobject` supplies
  the per-counter `CandidateProvider`.

Net: the `messages`/`mentions` counters never scan the change graph — neither
the old DiffManager rebuild nor the watermark full scan.

## Correctness / behavior

- The bounded stream is provably a **superset of every id the full scan would
  emit**: every dominated id has `OrderId ≤ maxFrontierOrderId`, and the
  non-message / already-read ids the full scan also visited are no-ops.
  Per-`advance` emission is therefore identical; the cross-invocation
  invariant (an id is added to `marked` only when emitted) means a
  not-yet-applied chat row is simply re-evaluated on the next `advance` — no
  permanent loss. No relaxation of the 2D `(OrderId, AddSeq)` semantics.
- One deliberate, owner-approved semantic change from change #1 vs the old
  DAG-ancestor model: a concurrent message already present that sorts before
  the seen head now counts as read (matches the linear chat timeline). The
  differential suite shows **22/25 scenarios byte-identical** to the old
  DiffManager engine; the 3 divergences are exactly this accepted class,
  recorded in `core/block/editor/chatobject/testdata/watermark_signoff.md`.

## Test plan

- [ ] `go test ./core/block/source/sourceimpl/... ./core/block/editor/chatobject/... ./core/block/chats/chatrepository/...`
- [ ] `TestWatermark_SignOffGate` green (non-accepted scenarios match the DAG baseline; only the 3 accepted scenarios diverge)
- [ ] bounded **completeness gate** (`bounded == full-stream emission`) + cross-invocation invariant tests green
- [ ] repository unread-filter equivalence + `read` / `hasMention` index-presence tests green

Design: `docs/superpowers/specs/2026-05-16-chat-read-counter-orderid-trim-design.md`
· Plan: `docs/superpowers/plans/2026-05-17-chat-read-counter-bounded-candidates.md`


requilence:
Pushed `862c21e36` — fixes a cold-start regression found by profiling the first cut: `advance` ran a full change scan + re-emitted the entire dominated set to `onRemove` on every call, ×(counters × #device-values), making `SetReadFlag` re-run a large `id IN` query each time. Fix: merge per-device seenHeads → one advance per counter; `onRemove` now carries only the newly-dominated delta (O(delta) writes). Sign-off gate unchanged (22 match DAG, same 3 accepted divergences). Repo-unread-driven O(unread) path documented as a follow-up.

deff7 (approved):

PR #3167, opened 2026-06-10
GO-7302 Add file layouts to head-sync priority after discussions

requilence:
Follow-up to #3163 (merged while this was in flight).

The per-space diffsync priority subscription now also tracks file objects (file/image/video/audio/pdf via `domain.FileLayouts`). `GetPriorityIds` orders **chatDerived chats → discussions → files → currently opened objects**; previously files weren't prioritized at all.

Since the subscription is no longer chat-only, the chat-specific internals were renamed (`chatSubs` → `prioritySubs`, sub id `block-chat-priority-` → `block-sync-priority-`).

Covered by an ordering subtest in `core/block/priority_test.go`.

### DIFF
--- a/core/block/priority_test.go
+++ b/core/block/priority_test.go
@@ -21,31 +21,28 @@ func TestService_GetPriorityIds(t *testing.T) {
 	const spaceId = "space1"
 	const otherSpaceId = "space2"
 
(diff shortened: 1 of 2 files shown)


requilence:
not needed

PR #3171, opened 2026-06-11
GO-7316 Fulltext search reliability fixes

requilence:
Closes GO-7316

A research audit of the tantivy fulltext pipeline (full report in `docs/fts/Research.md`) surfaced 7 confirmed bugs. Each was reproduced with a failing test before fixing; the repro tests remain as regressions.

## Fixes

1. **Indexing dedup was broken** — `filterOutNotChangedDocuments` re-added unchanged docs, so every object edit re-indexed (delete+add) *all* of its docs and the per-object tantivy lookup was pure overhead. Now ids seen during `Iterate` are tracked and only genuinely new/changed docs are re-indexed.
2. **FT deletion was a silent no-op** — `BatchDeleteObjects` received bare object ids while tantivy deletes by exact `IdRaw` term (full doc paths like `objectId/r/name`), so e.g. ACL-removed participants stayed searchable. Deletion now expands object ids to doc ids via a phrase query on the tokenized Id field — deliberately *not* a prefix query: tantivy prefix expansion caps at ~50 terms per segment, and deleted terms keep eating the budget (discovered by the 10k-doc repro test; documented in `docs/fts/Research.md` G.0).
3. **Space offload never cleaned tantivy** — `RemoveIndexes` now deletes the space's FT docs and queue entries, and the FT consistency check garbage-collects orphaned docs (per iterated space only, capped, skipped on truncated listings). It also no longer `fmt.Printf`s full object details to stdout.
4. **Queue lost updates racing a batch** — `FtQueueMarkAsIndexed` used to overwrite entries wholesale, dropping chat order ids / message deletions / re-enqueues that arrived while a batch was processing. Queue entries now carry a generation stamped on every enqueue; marking only applies when the generation is unchanged and clears the processed `ord`/`del` fields.
5. **100-doc cap before filtering** — store filters and pagination ran on a hard-capped tantivy top-100. The docs limit is now derived from offset+limit (clamped to [100, 1000], with truncation telemetry), and chat search uses a new `SearchChat` scoped to the chat's message docs so messages stop competing with the whole space for candidate slots.
6. **Chat search scoring/pagination** — float BM25 scores were truncated to int64 (mostly 0) and compared with `int(a-b)`; offset past the end returned the first page again. Sorting now uses the original float scores, the proto score is rounded, and out-of-range offsets return empty.
7. **Papercuts** — a single malformed doc id no longer fails the whole search; float comparators use `cmp.Compare`; empty delete terms removed; highlights disabled for the prefix-name search (its consumers discard them).

## Docs

`docs/fts/` contains the research report, the fix plan, and specs for both the implemented designs (deletion lifecycle, queue generations, search limits) and design-only follow-ups (boost dis-max, conditional Zh indexing, id tokenizer changes, filter pushdown) that need product sign-off / an index version bump.

## Testing

- All fixes repro-tested first (failing → green): `ftsearch_repro_test.go`, `fulltext_filter_repro_test.go`, `indexer_store_repro_test.go`, `service_search_repro_test.go`, plus new tests for space-removal cleanup, orphan GC, `ListIdsBySpace`, and the candidate-limit clamp.
- Full suites green: ftsearch, objectstore, spaceindex (incl. the GO-7266 fulltext tests), indexer, chats, database; repo-wide build and vet clean.
- Reviewed by a second agent pass; its findings (incl. a chat-indexing starvation regression in the generation design) are fixed and regression-tested.


requilence:
Added `5c09c2aef`: **page-filling escalation** replaces the static candidate clamp.

Problem found on desktop: client requests `limit=100`, post-FT filters trim the 100 candidates to 86 records, client reads the short page as "no more results" and stops paginating — while more matches exist in the index.

Instead of a `hasMore` proto flag (needs all-client changes, and without server-side budget growth the next offset request couldn't make progress anyway), `performFulltextQuery` now doubles the tantivy candidate budget (max 2000) until the requested page is filled or the index is exhausted. Short page ⇒ end-of-results is true again, no API change.

The original intent of the 100-doc cap (noise gate + bounded re-rank pool + bounded anystore reads, since BM25 isn't comparable across queries) is preserved: round 1 is always 100 docs — `Limit=0` no longer jumps to 1000 — and extra rounds only run for queries that demonstrably starved. `ftCandidatesTruncated` now counts pages left underfilled at the hard limit, which is the signal for whether tantivy-side filter pushdown is ever worth it.

requilence:
Added `a6c53a63e`: **two-tier ordering** — fixes the pagination-duplicates hazard escalation introduced.

The final order is `_final_score` (BM25 + recency/name boosts) while the candidate pool grows with the requested offset. Re-ranking the whole pool is not prefix-stable: a recency-boosted tail candidate can jump into an earlier page on the next request → duplicates on page N+1 and skipped results.

Now the boosts re-rank only the **first 100 BM25-ranked objects** (fixed pool, independent of the page — exactly the pool the original 100-doc cap gave them) via a stable sort; everything beyond stays in BM25 order, which IS prefix-stable under budget growth (deterministic id tiebreak added). Injections come from head hits only; the recency decay clock is hour-truncated so the head can't drift mid-pagination.

`TestFulltextPaginationConsistency` pins the property: pages fetched by separate requests (each with a different candidate budget) equal slices of one big query — no duplicates, no gaps, and a freshly-modified deep-tail object stays in the tail instead of jumping pages.

Remaining caveat, shared with all offset pagination over a live DB: an index commit between two page requests can still shift results; the complete cure (server-side search-session snapshot/cursor) is noted in SpecSearchLimits.md as a follow-up. Client-side dedup by id remains a cheap belt.

requilence:
Second adversarial review round (focus: migration of existing indexes, wrong assumptions, untested edges) → `21ad5af70`. Findings and fixes:

**Migration (the big one):** `ForceFTRecheckCounter` was still 0, so the branch's flagship repair machinery — consistency check, missing-doc backfill, orphan GC — was **dead code for every user** (and `ftsVer` is unchanged, so no rebuild either). Bumped to 1, after hardening what the first activated run does:
- the check now probes `WaitStoresLoaded` (non-blocking, retried at a later queue drain) — it used to run at first queue drain against a **partial space registry** and then latch done forever, permanently skipping not-yet-opened spaces;
- orphan GC skips spaces whose store is **empty** while FT docs exist (wiped/not-yet-rebuilt store, not genuine orphans) — prevents a burst misclassification;
- soft-deleted objects keep store stubs (`isDeleted=true`) which were **shielding** their leftover FT docs — the main historical leak this GC targets; stubs now count as absent;
- orphan deletions chunked (1000/lock acquisition).

**Prefix-stability had 3 holes** (review confirmed with throwaway repro tests):
1. the re-rank head shrank when a round yielded <100 objects → escalation now also materializes the full head pool;
2. `preferPluralNameRelation` is score-blind, so a bigger budget pulling in an object's lower-scoring pluralName doc changed its representative score → objects are now ordered by their **best** doc score (pluralName doc kept for metadata only);
3. injection budget was derived from offset+limit → now a request-independent constant.
New regression test: `TestFulltextPaginationConsistencyMultiDoc` (multi-doc objects force head materialization). Residual accepted edge documented: exact BM25 ties straddling the budget boundary.

**Smaller:** `NamePrefixSearch` results exempted from the min-score drop (it relied on highlight ranges the prefix path no longer generates); `GenerateFTQueueCounter` monotonic across clock-backward steps; misc comment/nit fixes.

Reviewer verified clean: legacy queue entries (gen-less, both pending and old-marked shapes) flow correctly through every new path; `BatchDeleteObjects` semantic change safe for all callers; SearchChat boost-0 clause semantics; escalation termination; anystore ModifyFunc/arena usage.

requilence:
Third adversarial review round (fresh eyes on the round-2 fixes + newest commits) → `472bd773e`. Findings and resolutions:

**HIGH — name boost flipped with the candidate budget** (confirmed empirically by the reviewer): the +1.0 boost keyed off the representative doc's relation key, but `preferPluralNameRelation` switches the representative once an object's pluralName doc enters the budget → `_final_score` shifted ±1.0 between page requests, re-breaking the pagination invariant for pluralName-bearing objects (types). **Fixed:** boost and injection eligibility now derive from the object's *best-scoring* doc (budget-stable), carried as `FulltextResult.NameMatch`. Side effect, aligned with the scoring spec: pluralName matches now get the boost too. Regression test `TestFulltextNameMatchBudgetStable`.

**MEDIUM — truncated orphan GC was silenced forever**: a capped (50k) or skipped GC still persisted the recheck counter, and the counter gate short-circuits every future session — 'collected next session' was false exactly for the years-old indexes the activation targets. **Fixed:** `RunFTConsistencyCheck` reports `complete`; the counter is persisted only on full coverage.

**MEDIUM — nearly-empty store bypassed the wipe guard**: a store with 1 object next to thousands of FT docs mass-GC'd them. **Fixed:** proportional guard (`ftObjects > 10× storeObjects` → skip + incomplete).

**LOW:** consistency-check iteration now observes ctx cancellation (shutdown no longer waits for a full sweep); `GenerateFTQueueCounter` advances the timestamp synthetically instead of sleeping toward a pinned future second after a backwards clock step; `limit=0` injections capped at the constant budget is now a *documented* decision (was silently changed from unlimited); dead `unlimited` param and vestigial error return removed; stale repro-test comments reworded.

Reviewer verified clean: lazy tail prefix property, escalation termination, the score-clone aliasing, legacy gen-less queue entries, the offload error-propagation/retry semantics, and the telemetry removal completeness. Known accepted caveats unchanged (exact-tie boundary, chat non-score deep paging, in-flight batch race).

PR #3172, opened 2026-06-11
GO-7315 Reduce subscription memory allocations and heap usage

requilence:
## Summary

Refactors the subscription system (`ObjectSearchSubscribe`, `core/subscription`) to cut memory allocations and steady-state heap usage, in reviewable phases (one commit each):

- **Benchmarks first** (`core/subscription/bench_test.go`): subscription init, in-window change batches, no-match batches, plus `keys/entry` and `heap-MB` cache-retention metrics.
- **Low-risk allocation fixes**: lazy entry bookkeeping maps, map-based batch dedup (was O(n²)), skip `details.Copy()` when no subscription listens to store updates, `CopyOnlyKeys` O(keys) instead of O(fields×keys), id indexes for `opCtx` entries/keysBuf, set-based dep-id dedup (also fixes latent `forceIds` slice aliasing).
- **Iterator-based init**: new `spaceindex.QueryRawIterate` streams raw anyenc docs without store-side sort (the skip-list orders entries anyway) or an intermediate `[]database.Record`; `QueryByIds` becomes one `In(ids)` query instead of per-id lookups; `ToProtoOnlyKeys` removes an intermediate map copy from every DetailsSet event.
- **Entry projection (the main heap win)**: each subscription registers the relation keys it needs (requested ∪ sort ∪ filter ∪ source keys, + name/orderId for object-relation sorts); cached entries are decoded and retained only for the union over their subscriptions, instead of keeping the full details of every matching object for the life of the subscription. Client events stay wire-identical: amend/unset payloads are already filtered to per-sub keys (`state.StructDiffIntoEventsWithSubIds`).
- **Coalescing change buffer**: `recBatch` (unbounded `mb.MB` of full-detail copies) replaced with a latest-wins-per-object buffer, bounding pipeline memory during mass indexing.

## Results (M2, 10k objects, 11 requested keys, limit 100)

| Metric | Before | After |
|---|---|---|
| Search (subscribe) | 112 ms / 35.7 MB / 762k allocs | 80 ms / 21.1 MB / 451k allocs |
| Cache retention | ~26 keys/entry (full object) | ~11 keys/entry |
| No-match change batch | 111 allocs | 91 allocs |
| In-window change batch | 201 allocs | 251 allocs (re-projection of changed entries — trade for the smaller resident cache) |

## Review hardening

A multi-agent code review of the branch surfaced three must-fix issues, all addressed in the last commit with regression tests:
- required-keys computation moved under the space lock (data race on the dependency service relation-format cache with concurrent Search calls; `TestConcurrentSearch` under `-race`),
- the collection observer pushes full store details for changed membership (projected cache entries fed into the pipeline could make filters of unrelated subscriptions misfire); only deleted objects fall back to the cache,
- dependency entries served from the cache are refreshed from the store when another subscription's projection trimmed keys they need (`TestDepEntriesRegainProjectedKeys`),
- subscriptions with an empty `Keys` list are excluded from projection entirely (fail-safe full retention).

Consciously accepted (no behavior worse than before): a same-batch join of a second subscription can include its keys in the sibling's amend (values correct; DetailsSet already leaked union keys pre-refactor); `QueryByIds` returns shared `*Details` for duplicate input ids and fails on iterator doc errors instead of skipping one id.

## Notes

- Entries belonging to an unregistered subscription keep full details as a fail-safe; the disabled nested-subscriptions path is registered minimally and stays dead.
- The in-repo integration suite (`make test-integration`) currently does not build on develop itself (stale `tests/testing_test.go` vs the `RpcWalletCreateSession` oneof) — verified pre-existing, not from this branch. Running the external e2e suite against this branch before merge is recommended.
- Verified: full subscription/objectstore/domain suites green incl. `-race`; broad sweep over `QueryByIds`/subscription consumers (block editor, export, syncstatus, files, acl, api, space) green; rebased on latest develop (lazy cross-space subscriptions, fulltext rework) with suites re-run.

Closes GO-7315

### DIFF
--- a/core/domain/details.go
+++ b/core/domain/details.go
@@ -62,6 +62,34 @@ func NewDetailsFromAnyEnc(v *anyenc.Value) (*Details, error) {
 	return res, visitErr
 }
 
+// NewDetailsFromAnyEncWithKeys decodes only the listed keys of the document.
+// A nil keys map decodes everything, same as NewDetailsFromAnyEnc
+func NewDetailsFromAnyEncWithKeys(v *anyenc.Value, keys map[string]struct{}) (*Details, error) {
+	if keys == nil {
+		return NewDetailsFromAnyEnc(v)
+	}
+	obj, err := v.Object()
+	if err != nil {
+		return nil, fmt.Errorf("is object: %w", err)
+	}
+	res := NewDetailsWithSize(len(keys))
+	var visitErr error
+	obj.Visit(func(k []byte, v *anyenc.Value) {
+		if visitErr != nil {
+			return
+		}
+		if _, ok := keys[string(k)]; !ok {
+			return
+		}
+		// key is copied
+		err := setValueFromAnyEnc(res, RelationKey(k), v)
+		if err != nil {
+			visitErr = fmt.Errorf("key %s: %w", k, err)
+		}
+	})
+	return res, visitErr
+}
+
 func setValueFromAnyEnc[T ~string](d *GenericMap[T], key T, val *anyenc.Value) error {
 	switch val.Type() {
 	case anyenc.TypeNumber:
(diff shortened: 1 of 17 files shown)


requilence:
closed in favor of #3177

PR #3173, opened 2026-06-11
GO-7314 Make participant objects store-only

requilence:
## Problem

On every start, for every loaded space, `aclobjectmanager` reprocesses the full ACL member list and materializes each member as a participant smartblock (~2 `space.Do` load+apply cycles per member). On a real large account this measured **1,549 participant smartblock loads per start**, with **2,202 participants pinned in the object cache until shutdown** (`TryClose` returned `false`). All of this work was pure overhead: participants are already virtual objects (static source, no tree, `PushChange` no-op) whose only persistence is the objectstore — the smartblock layer was just an expensive write funnel.

Full design rationale, verified research and implementation notes: `docs/ParticipantObjectsRefactoring.md` (added in this PR).

## Changes

**Phase 1 — direct store writes + fulltext funnel**
- New `core/participants` package owns all participant record writes: atomic merge into the space index (subscription events fire only on real diffs — same primitive the indexer uses), `BindSpaceId` and fulltext enqueue committed *before* the merge so partial failures replay cleanly; FT is enqueued only when `name`/`description` change (the only participant keys producing FT docs)
- `participantwatcher` builds ACL/base details and writes through the store — no `space.Do`
- The participant editor keeps only `Init` (read-only view for ObjectOpen, preserving the GO-7111 no-diff-on-open logic); `Modify*`, `DetailsUpdatable` and the `TryClose` cache pin are removed
- New `SmartBlockType.FulltextDetailsOnly()` + a details-only branch in `prepareSearchDocs`: the FT worker builds participant docs from store details without loading objects, sharing the exact relation-doc filtering via `prepareRelationSearchDocsFromDetails`. Covers all FT entry points (queue, FT-db rebuild, reconcile, consistency check)

**Phase 2 — identity fan-out via the store**
- `RegisterIdentity(spaceId, identity, key)` drops the observer callback; profile updates merge directly into the participant records of registered spaces (update-only; creation stays in the ACL path)
- Per-identity profile encryption keys persist in commonDb (`identity_encryption_keys`); `nil` key reuses the persisted one; the own identity's key is derived on demand and persisted
- The per-(identity, space) observer map degrades to a set scoping identity-repo polling

**Phase 3 — skip ACL reprocessing when the head is unchanged**
- `processAcl` persists a last-processed-ACL-head marker per space and early-outs on match: no per-member iteration, no per-member ACL metadata decryption — only identity tracking is re-registered from a single store query (`WatchPersistedParticipants`), with fallback to full processing on any error
- `RemoveAclIndexes` clears the marker (covers forced reindex and space offload); notifications still run on the skip path

## Impact

On the 1,519-member space: zero participant smartblock loads on warm start (previously ~3k load+apply cycles), no pinned cache entries, FT worker and FT-db rebuilds no longer materialize participants, and with an unchanged ACL head the per-member metadata decryption is skipped too.

## Behavior notes

- No migration/reindex needed: existing store records already contain every key the new write path produces
- One accepted change: an *already-open* participant page no longer receives live detail updates through smartblock apply events (reopening always shows fresh data; subscription-backed views update live as before)
- Client RPC surface unchanged

## Testing

- New unit/integration tests: merge semantics & FT-enqueue gating (`core/participants`), ACL-path record parity incl. self-participant and hidden-discovery (`participantwatcher`), store fan-out incl. restart key-persistence roundtrip (`core/identity`), skip/fallback paths (`aclobjectmanager`), details-only FT docs with a strict no-load mock (`core/indexer`), marker clearing (`RemoveAclIndexes`)
- Full sweep of `./core/... ./space/... ./pkg/lib/...` passes
- An adversarial review pass was run over the diff; the three findings (side-effect replay ordering, own-key persistence gap, registration-vs-broadcast stale write) are fixed in this PR


requilence:
Second review round (focused on participant details consistency across all writers) found one critical gap, fixed in 3fe5b3851:

**The smartblock→indexer full-replace write-back was still alive for participants.** `Apply`/`StateAppend` → `runIndexer` → `UpdateObjectDetails` writes the state's `CombinedDetails` wholesale, and with the store now being the source of truth, a stale cached state was the one remaining writer that could move a record *backwards*. Three real triggers:

1. **Backlinks watcher cached-object fallback** — `injectDerivedDetails` refreshes only the backlinks key from the store; the rest of the cached state (name, permissions, status) is stale and got written back. E.g.: member page cached → member renames themselves (store updated by identity fan-out) → someone mentions that member anywhere → backlinks fallback reverts the name. Worse: a reverted `participantStatus`/`participantPermissions` survived restarts, because the Phase-3 ACL-head marker deliberately skips reprocessing while the head is unchanged.
2. **Every ObjectOpen** produced an Apply that reached the indexer (snippet/mentions/resolvedLayout links are added child-only, so the no-diff-on-open goal of GO-7111 didn't hold for these records), opening a lost-update window exactly during the space-load write burst — and leaking state-only keys (`layout`, `featuredRelations`, `snippet`, …) into the records.
3. **Layout syncer** `needApply` path (rare).

Fix: `FulltextDetailsOnly()` is renamed to `DetailsStoreOwned()` (one property: the store record is the source of truth, smartblock state is a derived view), `indexer.Index()` returns early for such types, and the backlinks watcher modifies store-owned records directly instead of falling back to the object cache. Pinned by tests: a stale-state `Index()` call leaves the record untouched, and backlinks land in the store without touching the cache.

Explicitly audited and ruled out in the same round: sync-status detail injection (participants excluded twice), readonlyfixer / systemobjectreviser / object-context migrations (none touch participants), `migration.RunMigrations` (no migrations registered for the participant editor), reindex paths (tree-based; both `removeParticipants` and `ClearHeadsState` clear the ACL marker → full reprocessing), `type` detail staleness (derived per-space ids are deterministic), and the `upsert` semantics of the new `ModifyObjectDetails` (update-only skips missing docs; subscription events always carry full merged details).

PR #3177, opened 2026-06-12
GO-7320 Reimplement the object subscription engine from scratch

requilence:
## What

Removes the legacy object-subscription engine and reimplements it from scratch against the behavioral contract in [docs/Subscriptions.md](docs/Subscriptions.md) — derived from how the desktop client and heart's own services actually consume subscriptions. The architecture is documented in [docs/SubscriptionsEngineDesign.md](docs/SubscriptionsEngineDesign.md), written and review-hardened before any code.

## Architecture

One core live-query primitive — `{optional id-scope, compiled filters, optional order+window, requested keys}` over a per-space store feed — surrounded by an adapter ring. Everything Anytype-specific is a leaf:

- **Sources** resolve to type/relation filters at the request boundary (AND-composed onto the compiled filter)
- **SubscribeIds** = fixed id scope in request order, no filters
- **Collections** = mutable id scope fed by the editor's membership stream
- **Dependencies** = hidden detail-events-only child sub (`{subId}/dep`) with a derived scope; order deps probe the compiled order's own OrderMap per feed item
- **Kanban groups** = an adapter owning no core sub at all (relevance map + grouper recompute off the mutex)
- **Cross-space** = existing `crossspacesub`, unchanged, layered on the public API

Key mechanics: coalescing intake (id-keyed, full state replaces pending, id-only degradation above a threshold), one worker per space, FIFO outbox encoded to pb off-mutex, snapshots **streamed** with a bounded top-K window heap (peak O(set × entry) + O(window × details) instead of O(set × full details)), no full-set ordered structure at all (the legacy skip list is gone — ordering exists only for the window, everything else is boundary comparisons plus bounded re-queries), window-diff scripts that guarantee every `afterId` is already placed when the client applies it, subId claim slots serializing replace-on-resubscribe against racing unsubscribes, and a detach-and-flush teardown making `UnsubscribeAndReturnIds` ghost-proof.

Store-side prerequisites: feed notifications raised under a `WriteTx` are buffered (ids only) and flushed only on commit with a post-commit re-read — closing both the silent data-loss window in the wire-feed-then-requery snapshot invariant and the stale-overtakes-newer reorder race; nested tx (savepoint) notifications flush only on the root commit. New `QueryIterateRaw` streams precompiled-filter queries without materializing the result set.

## Not carried over (confirmed-dead surface, contract §4)

afterId/beforeId request cursors, nextCount/prevCount, nested sub-subscriptions, fulltext in subscriptions, membership events for `/dep` subs, `Service.SubscribeIds(subId, ids)`, `UnsubscribeAll`.

## Testing

- Event-level unit tests asserting exact wire sequences (`DetailsSet → Add → Counters`, minimal Amend, window scripts)
- A randomized replay test applying every event with the anytype-ts dispatcher's exact rules, converging to an independently computed window each step
- The pre-existing crossspacesub + objectsubscription suites pass unchanged as acceptance gates (incl. lazy promotion, the first-opener deadlock guard, and no-data-loss under concurrent open, with `-race`)
- **Three adversarial review rounds** (architecture/memory/any-store; edge-cases/races with empirical probes; one line-level reviewer per file) — every confirmed finding fixed with a regression test, including a 30-generation `UnsubscribeAndReturnIds` consistency stress
- **Legacy parity audit**: develop's test scenarios inventoried (17 clusters covered, 8 dead-surface, 11 gaps closed — tag combination groups, nested And/Or trees, emptyPlacement sorts, ids+deps, multi-sub fan-out, collection windows, status-option reorder, source resolution); probes against the legacy engine confirmed the offset-window drift and pre-commit store notifications were real inherited defects, while legacy never emitted live status/checkbox group events at all
- Full consumer battery green: `syncstatus`, `files`, `pushnotification`, `acl`, `space`, `api`, `chats`, `relationutils`, `indexer`

## Notable drive-by fixes

- `core/acl`: deleted-space self-remove messages were keyed by a shared fixed id — concurrent deleted spaces lost all but one `Leave`
- `crossspacesub`: synthesized counters events re-inserted just-deleted per-space totals, double-counting all subsequent totals; `Close` map race
- `FilterOptionsEqual` no longer allocates per evaluation

## Known accepted edges (documented in the design doc / code)

Sorted-sub re-queries can truncate tie groups beyond the margin; compiled filters are frozen until re-subscribe (quickOption dates, nested dot-key id sets); groups' `CollectionId` scoping is a subscribe-time snapshot; a killed overflowing internal queue leaves its sub registered until the consumer unsubscribes; pre-commit link notifications and `DeleteDetails` invisibility are pre-existing store behaviors out of scope here.


requilence:
## Benchmark comparison vs the legacy engine

All benchmarks ported to the legacy engine on `develop` with identical workloads (same object counts, filters, sorts, limits, keys), run on the same machine, compared with benchstat at n=6 (storms n=3). `BenchmarkRealisticSpaceUpdate` skipped (porting surface too large).

### Time

| Benchmark | legacy | new | delta |
|---|---|---|---|
| Subscribe full snapshot @10k | 21.5ms | 16.1ms | -25% |
| Subscribe windowed @10k (name sort, limit 100) | 58.8ms | 18.0ms | **-69%** |
| Change batch, 100 updates × 20 subs | 3.97ms | 1.00ms | **-75%** |
| limit-0 steady-state no-op | 557µs | 745ns | **-99.9%** |
| Window-member Amend @500 | 64.7µs | 2.2µs | **-97%** |
| In-window reorder @500 | 71.6µs | 37.1µs | -48% |
| Full-list reorder @10k (limit 0) | 701µs | 565µs | -19% |
| Out-of-window update | 49.8µs | 2.5µs | **-95%** |
| Groups subscribe @1k (tag, 20 options) | 1.98ms | 2.57ms | +30% |
| Groups miss path | 374ns | 108ns | -71% |
| Feed storm, 1k writes end-to-end | 264ms | 17.7ms | **-93%** ¹ |
| Feed storm, 5k writes (degradation engaged) | 374ms | 90.8ms | **-76%** ¹ |

¹ legacy's number includes its fixed ~250ms batching debounce — by design, but it was real user-facing latency.

### Memory: allocation churn (B/op · allocs/op)

| Benchmark | legacy | new | B/op delta |
|---|---|---|---|
| limit-0 steady-state no-op | 878KB · 98 | **0 B · 0** | -100% |
| Out-of-window update | 55.8KB · 23 | 440B · 4 | -99% |
| Window-member Amend @500 | 62.5KB · 97 | 1.1KB · 10 | -98% |
| Change batch 100×20 subs | 3.8MB · 18.4k | 1.7MB · 8.0k | -54% |
| Subscribe full @10k | 21.2MB · 270k | 11.8MB · 200k | -44% |
| Subscribe windowed @10k | 16.3MB · 251k | 10.8MB · 201k | -34% |
| Feed storm 1k end-to-end | 4.3MB · 67k | 2.9MB · 57k | -31% |
| Groups miss path | 8B · 0 | 0 B · 0 | -100% |
| In-window reorder @500 | 95.6KB · 106 | 115.7KB · 20 | **+21%** |
| Full-list reorder @10k | 1.5MB · 171 | 2.1MB · 53 | **+40%** |
| Groups subscribe @1k | 1.6MB · 22.9k | 1.8MB · 27.1k | +11% |

The two reorder regressions are the window-diff script's per-batch bookkeeping (`oldWin` snapshot + position map + client-list sim) vs legacy's in-place skip-list splice — fewer-but-bigger allocations (allocs/op is 69–81% *lower*), amortized per batch rather than per item, and faster in wall time anyway. Groups subscribe pays +11% for the relevance/options seeding that buys the 108ns/0-alloc miss path and live status/checkbox group events (which legacy never emitted at all).

### Memory: retention (not visible in B/op)

- **Legacy** kept every member of every sorted sub as a skip-list node with sort state, plus a refcounted cache of full details per tracked object — a windowed view over 100k objects retained ~35MB+ of ordered per-member state.
- **New** retains an id-only member set (~50B/member) plus full projections for the window only; subscribe-time snapshots stream through a bounded top-K heap. Peak is O(set × entry) + O(window × details) instead of O(set × full details) — roughly **~10MB vs ~250MB peak** during a 100k-object subscribe, the difference between fine and OOM under gomobile memory limits.
- `FeedStorm5kDegraded` shows the storm bound holding: past the 2048-pending intake threshold both engines churn ~22MB per 5k-write storm, but the new intake pins at most 2048 full details at any instant (the rest degrade to id-only entries re-fetched in batches) with per-write cost flat at ~18µs.

### Excluded as not comparable (annotated honestly)

- **Groups recompute**: legacy diffed groups synchronously inline under the lock; the new engine recomputes off-mutex through the kanban grouper (a store query). Different execution models — the new number includes work legacy deferred elsewhere.
- **Object-format sort reorder / sort-target rename** (legacy ~22µs/0.9µs vs new ~5.3ms): legacy with `NoDepSubscription: true` never maintained assignee-order at all — it degenerated to id-order, so its fast numbers measure a correctness gap. The new engine's ~5.3ms (dominated by the re-query's SQL ORDER BY collation pass over all matching rows) buys the contract's §3.5 behavior: renaming the assignee you sort by actually reorders the list.

cheggaaa (approved):

PR #3205, opened 2026-07-10
GO-7366 Fix files stuck in Syncing: confirm sync via file node without local blocks

requilence:
## Problem

File objects can get stuck showing "Syncing" forever (`fileBackupStatus = Queued`, no `syncError`), with nothing actionable in the UI or logs. The bytes are typically already on the file node — the client just can never confirm it:

1. **`filesync.AddFile` silently dropped files whose root block wasn't in the local store** — no queue item, no log. But a file object may legitimately reference a fileId with no local blocks: import dedup reuses an existing fileId **cross-space** without re-adding blocks, while the file-object lookup is space-scoped. Importing the same content into a second space therefore creates a fresh object born at `Queued` that filesync never touches.
2. Even for queued files, confirming "already backed up" required **walking the entire DAG through the local-only store**: any locally-missing block failed the walk and parked the item in `FileStateMissingBlocks` — terminal and silent since GO-7034. The bind path (`BlocksCheck` → `BlocksBind` → `Synced`) was unreachable exactly for the files that needed it.
3. Dedup had **no `isDeleted` filter**, so a canceled/rolled-back import (objects deleted, blocks offloaded) followed by a re-import reuses fileIds whose blocks were just wiped.

## Solution

- `AddFile` always enqueues; local presence is no longer a precondition.
- New `enumerateFileCids` collects the full cid set + sizes **without leaf data**: structural dag-pb nodes are read local-first with file-node fallback (cached locally); leaf chunks are enumerated from parent links when provably leaves (unixfs metadata; exact for our 1MiB chunker). Enumerating a fully-remote 10GB file costs ~500KB of structural fetches — content is never downloaded.
- Plan-driven upload: cids the node already has are **bound by cid** (batches of 64); only `NotExists` blocks are read from the local store and pushed. A file already on the node reaches `Synced` with zero bytes transferred in either direction. Side effect: `ExistsInSpace` blocks are no longer re-uploaded on resume.
- `FileStateMissingBlocks` now means "blocks found neither locally nor on the node": logged at WARN **with objectId/fileId/spaceId**, retried by a dedicated queue consumer with capped exponential backoff (10min ×4ⁿ, max 24h), woken immediately by `AddFile`/`MarkUploaded`. Parking drops the cached availability plan so retries re-check the node; transient (offline) errors keep the long backoff instead of the 1-minute retry.
- `isDeleted` filters on both dedup queries and on the startup re-enqueue sweep.

## Repair of already-stuck files

No migration needed: the existing startup sweep (`ensureNotSyncedFilesAddedToQueue`) re-enqueues own non-Synced file objects; they now get verified against the node, bound, and flipped to `Synced`, which replicates to other devices. No re-upload, no content download (~120KB of structural/small-variant fetches per image-class file, cached).

## Compatibility / resources

- Only pre-existing `fileBackupStatus` values are written; old clients converge via their `markUploadedHook` when `Synced` replicates in. Queue-item schema change (`missingBlocksRetries`) is tolerated in both up-/downgrade directions.
- Node-side `BlocksBind` is idempotent and quota-enforced; the limit error routes to the existing `Limited` path. Simultaneous repair from two devices is safe.
- Steady-state cost of a permanently-missing file: ~1 BlockGet + 1 WARN line per day.

## Testing

- New reproducers in `core/files/filesync/bind_test.go` (all fail on the old code): bind-without-local-blocks, partial-local, mixed bind+upload, missing-everywhere → park → recover via retrier, stale-plan drop, `AddFile` wake.
- `TestFileSync_AddFile_SkipWhenNotLocal` inverted to assert the new queueing behavior; dedup `isDeleted` tests in `core/files/queries_test.go`.
- Full `./core/files/...` and `./core/syncstatus/...` suites pass.

Linear: GO-7366

### DIFF
--- a/core/files/fileobject/service.go
+++ b/core/files/fileobject/service.go
@@ -270,6 +270,11 @@ func (s *service) ensureNotSyncedFilesAddedToQueue() error {
 				Condition:   model.BlockContentDataviewFilter_NotEqual,
 				Value:       domain.Int64(int64(filesyncstatus.Synced)),
 			},
+			{
+				RelationKey: bundle.RelationKeyIsDeleted,
+				Condition:   model.BlockContentDataviewFilter_NotEqual,
+				Value:       domain.Bool(true),
+			},
 		},
 	})
 	if err != nil {
--- a/core/files/filestorage/rpcstore/inmemory.go
+++ b/core/files/filestorage/rpcstore/inmemory.go
@@ -62,6 +62,11 @@ type InMemoryStore struct {
 	errMu        sync.Mutex
 	spaceInfoErr error
 
+	// availabilityOmit cids are silently dropped from CheckAvailability
+	// responses. Used in tests to simulate a node that doesn't answer for
+	// some of the requested cids.
+	availabilityOmit map[cid.Cid]struct{}
+
 	stats *InMemoryStoreStats
 }
 
@@ -113,6 +118,9 @@ func (t *InMemoryStore) CheckAvailability(ctx context.Context, spaceId string, c
 
 	checkResult := make([]*fileproto.BlockAvailability, 0, len(cids))
 	for _, cid := range cids {
+		if _, omit := t.availabilityOmit[cid]; omit {
+			continue
+		}
 		status := fileproto.AvailabilityStatus_NotExists
 		if _, ok := t.store[cid]; ok {
 			status = fileproto.AvailabilityStatus_Exists
@@ -135,8 +143,13 @@ func (t *InMemoryStore) BindCids(ctx context.Context, spaceId string, fileId dom
 
 	var bytesToBind int
 	for _, cid := range cids {
+		b, ok := t.store[cid]
+		if !ok {
+			// The real node refuses to bind a cid it doesn't have
+			return fileprotoerr.ErrCIDNotFound
+		}
 		if !t.isCidBinded(spaceId, cid) {
-			bytesToBind += len(t.store[cid].RawData())
+			bytesToBind += len(b.RawData())
 		}
 	}
 	if !t.isWithinLimits(bytesToBind) {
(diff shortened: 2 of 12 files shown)


requilence:
Ran a 4-lens adversarial review (correctness/concurrency, DAG & data integrity, failure modes/operations, tests & contracts) over the first commit. Results, addressed in e4ba1c1:

**Confirmed and fixed**
- *Correctness*: a parked (MissingBlocks) item whose availability check finally succeeded kept its parked state through the rest of the pipeline — a transient failure after recovery (e.g. a SpaceInfo blip) re-buried it for up to 24h and retained the fresh plan on a parked item. It now returns to PendingUpload (counter reset) the moment the check succeeds. Regression test is mutation-verified.
- *Operations*: an offline device churned its whole re-enqueued backlog every 60s with dial attempts + ERROR logs (the long backoff only covered MissingBlocks, a state unreachable while offline). Connectivity failures (`ErrNoConnectionToAnyFileClient`, `ErrUnableToConnect`) now reschedule at 5min and log at WARN.
- *Quality*: bare error return in `bindCids`; misleading safety comment on `childrenAreLeafChunks` (the per-child blocksize check makes the leaf classification exact for any chunker above ~6KB — stronger than previously stated).
- *Test gaps*: added coverage for >64-cid check/bind batching, the ExistsInSpace fast path, unanswered cids in a BlocksCheck response, bind-not-found dropping the cached plan, deletion resetting a parked schedule, and the retry-delay math. The in-memory node fake now returns `ErrCIDNotFound` for unknown cids in BindCids (was a panic) and can omit cids from availability responses.

**Verified clean (survived refutation)**
- DAG lens: leaf-proof sound for every anytype-produced DAG shape (incl. 174/175-chunk boundaries, multi-level, empty/single-chunk); size accounting byte-identical to the old walk; bind completeness and cross-space isolation hold.
- Concurrency: queue locking serializes all per-object access; limit allocate/deallocate pairing balanced on every branch; batch-callback races and shutdown clean.
- Crash windows converge: already-bound cids come back as ExistsInSpace on reprocessing, so partial bind/upload interruptions are idempotent.

**Noted, not addressed here** (low/pre-existing): startup re-enqueue has no jitter (one-time, per-device serialized by the reserved connection — capacity note for release day); state-rebuild hooks can wake a parked item for one extra node round-trip (by design, bounded); a pre-existing free-space miscount in `getLimitedFile` when an item parks mid-drain.

PR #3211, opened 2026-07-16
GO-7379 Snappy sync: proactive connection recovery on network/device state changes

requilence:
Linear: [GO-7379](https://linear.app/anytype/issue/GO-7379) (relates to GO-3958, GO-7272)

Sync went stale for 30–40s (shown as *Synced*) after macOS sleep/network changes and mobile background/network switches, because nothing above the transports probes liveness and no signal proactively reset connections. Full audit: `docs/sync-connection-staleness-audit.md`; client integration guide: `docs/mobile-network-integration.md`.

## What this does

- **Connectivity recovery pipeline** (`core/device`): every signal — network-type RPC (extended with `networkId` so same-type Wi-Fi→Wi-Fi switches are visible), foreground resume >15s, interface-address loss, wake from sleep — runs pool flush → responsible-peers rebuild → head-sync of all spaces. Leading-edge with a 5s suppression window and one coalesced trailing run.
- **Go-side net monitor**: desktop finally gets signals without any client RPC — 5s interface snapshot diff (IPv4 + IPv6 /64 prefixes; loss-only, so docker/VPN/hotspot additions don't tear down healthy connections) plus a wall-vs-monotonic clock-jump sleep detector. Feeds `IsOffline`; explicit client reports stay authoritative so the heuristic can't wedge a mobile device offline.
- **Peer manager**: instant rebuild on connectivity events; dial cadence 20s→2min while offline with instant snap-back; peer-find deadlines (30s broadcast / 1min sweep) free the 4 shared streampool dial workers that offline broadcasts used to park forever; one immediate diff round on the ConnectionError→Online edge replaces broadcasts dropped while offline; **closed peers are never served from the cache** — this last one fixes the iOS background→foreground stall where the post-flush head-sync grabbed just-closed peers, no-oped on suppressed ErrClosed, and waited out the next 20s tick.
- **Tuning/cleanups**: yamux/QUIC keepalives 10s (dead-conn detection ≤20s on yamux); filesync node-usage poller backs off on errors; `SyncAllSpaceHeads` sweeps coalesced; Android per-interface debug println removed.
- **Proto**: `Rpc.Device.NetworkState.Set` gains optional `networkId` (backward compatible; empty keeps type-only semantics).

No any-sync bump required — everything works against the pinned v0.12.14. Optional follow-up after the next any-sync release: `MaxIdleTimeoutSec: 20` for QUIC (branch `net-quic-idle-timeout` in any-sync).

## Behavior changes to be aware of

- Foreground resume head-sync threshold unified 20s→15s (matches the flush threshold).
- A broadcast with no reachable peers now drops after 30s instead of parking a worker until reconnect; delivery is covered by the reconnect diff-kick + periodic diff rounds.
- Keepalive frequency triples on idle connections (both transports, all platforms) — flagged in GO-7379 as an open question whether mobile should get a laxer setting.

## Testing

- New unit tests: recovery pipeline (suppression/coalescing, close-cancels-pending, IsOffline precedence, networkId semantics), net monitor (clock jump, loss-only diffing, fail-open enumeration), peer manager (rebuild signaling, reconnect diff-kick, closed-peer filtering, offline cadence plumbing), fixtures fully deterministic (no host-network dependence).
- `go test -race` clean on all touched packages; full repo build green.
- Independently reviewed by an adversarial pass; all required findings addressed (review notes folded into the audit doc).


requilence:
Second independent review pass completed (fix-verification + fresh hunt). All 8 first-round fixes verified OK. Three new findings, all addressed in the last two commits:

- **High**: the loss-only interface trigger missed connectivity *returning* — a fully-offline desktop only recovered at the 2min offline probe (regression vs develop's 20s). The monitor now fires on the empty→non-empty snapshot edge (`interface addresses regained`).
- **Medium**: on resume, the foreground RPC + overdue clock-jump tick always coalesced into a trailing run that re-flushed the just-re-established connections. The trailing run now compares a connectivity fingerprint against the leading run and skips duplicates of the same physical event.
- **Medium**: offline gating degraded local-only/LAN sync (GO-7272 scenario) — re-dial cadence stays at 20s whenever the space has LAN peers, and the Android guide no longer maps missing `NET_CAPABILITY_VALIDATED` to `NOT_CONNECTED` (validation gates internet, not LAN sync).

Also: offline sweep warnings downgraded to debug (was one warn per space per sweep). Race suites green after the changes.

requilence:
Third independent review pass (fix-verification of round 2 + fresh hunt). Round-2 fixes verified OK except one: the trailing-run fingerprint recorded raw snapshot values captured *before* the same tick's interface check, so a short outage bracketing a wake (Wi-Fi re-associating to the same network seconds later) could make the trailing run wrongly skip — the leading run had acted while the link was down, all dials failed, and nothing re-dialed until the 2min offline probe. Fixed in `6cca51a40`: the fingerprint now uses a monotonic monitor-generation counter (bumped on every observed change), so a down-and-back flap is never mistaken for a duplicate signal even when the address set ends up identical; also decoupled the suppression window (6s) from the 5s monitor tick to remove a boundary race. Regression test added for the exact swallow scenario.

Noted as non-blocking follow-ups in the audit doc: reconnect currently runs ~2 diff rounds per space (reconnect kick + recovery sweep overlap — correct but redundant), and persistent link flapping costs one flush+sweep per ~6s. The reviewer also confirmed end-to-end: the `errors.Is` chain for the deadline sentinel is unwrapped through any-sync, LAN cadence cost is negligible, and the one `./space` race-suite failure seen was an unreproduced pre-existing flake (11 clean re-runs, no test files in that package touched by this branch).

PR #2873, opened 2025-12-28
GO-1234 feat: improve setup-protoc for ARM64 and handle archived websocket package

pjjjv:
This PR improves the setup-protoc logic to support ARM64 architecture and dnf-based distributions. It also replaces the archived nhooyr.io/websocket package with a maintained fork to fix build issues. Related to https://github.com/anyproto/anytype-ts/issues/83

PR #3213, opened 2026-07-17
GO-1234 Fix local discovery auto-start on self-hosted networks

pjjjv:
---

---

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md

### DIFF
--- a/space/spacecore/localdiscovery/localdiscovery.go
+++ b/space/spacecore/localdiscovery/localdiscovery.go
@@ -92,11 +92,6 @@ func (l *localDiscovery) Init(a *app.App) (err error) {
 }
 
 func (l *localDiscovery) Run(ctx context.Context) (err error) {
-	if l.manualStart && len(l.nodeConf.Nodes) > 0 {
-		// let's wait for the explicit command to enable local discovery
-		return
-	}
-
 	return l.Start()
 }
 


requilence:
Heads up — `manualStart` isn't dead code, it's actively used in production: mobile clients (iOS/Android) rely on deferring local discovery start until the user grants the local network permission, for UX reasons. Removing this gate breaks that flow for everyone, not just self-hosted networks.

Also, `len(nodeConf.Nodes) > 0` is true for both default and self-hosted (`CustomConfig`) networks — it's only false in `LocalOnly` mode — so I don't see what's specific to self-hosted here. What exact problem were you trying to solve? A repro/description would help find the right fix without removing the mechanism mobile depends on.

pjjjv:
Thanks for the clarification! During debugging of local sync on a self-hosted network, the AI misdiagnosed manualStart as suppressing auto-discovery when remote nodes were present in the config, and did not realize manualStart is the mobile permission gate. I was having various troubles getting the sync to work and I thought this contributed to the problems. I'll close this PR.

By the way, can you have a look at https://github.com/anyproto/anytype-heart/pull/2873 ? Without this change, it simply was not compiling on aarch64.

PR #3214, opened 2026-07-17
GO-7379 Cut cellular-switch recovery latency: bounded node lookups, fast retry, prioritized sweep

requilence:
Follow-up to #3211, driven by a field test on iOS ([GO-7379](https://linear.app/anytype/issue/GO-7379)): Wi-Fi→cellular switch with 71 spaces — the recovery pipeline fired correctly (debug stats: CELLULAR reported with networkId, 2 recoveries, coalescing worked), but chat messages still took ~30s to flow.

## Diagnosis (from the new debugstat + a goroutine dump taken mid-stall)

The goroutine dump at recovery+6s showed the entire stall in three numbers: **71** `fetchResponsiblePeers` parked in ocache single-flight dial waits, **all 10** `SyncAllSpaceHeads` workers parked in `waitResponsiblePeers` behind them, **1** QUIC dial in flight. `pool.GetOneOf` shuffles the responsible node ids and parks behind the picked node's in-flight dial; on a fresh cellular path one node's sequential scheme×addr dial chain (10s per attempt) held every space that picked it — including the visible chat — while a different node had already connected at +6.6s and helped nobody.

## Fixes

- **Bounded node lookup** in `fetchResponsiblePeers` (10s): `ocache.waitLoad` honors ctx, so a fetch parked behind a slow dial gives up instead of inheriting the slowest dial's schedule.
- **Fast retry** (5s, was 20s) after a failed lookup while online: the next `GetOneOf`'s active-conn check returns any node that connected meanwhile, and the `ConnectionError→Online` edge then fires the existing per-space reconnect diff-kick. Offline behavior (2min backoff, LAN exception) unchanged.
- **Sweep can't be starved**: worker peer-find deadline 1min→15s (workers hold semaphore slots while waiting; stragglers are covered by the reconnect kick + fast retry), and the sweep orders **spaces with currently-open objects first** (block service looked up by component name to avoid the import cycle; drift test added in `core/block`).

Net effect for the observed incident: the visible space syncs as soon as *any* responsible connection is up (~7s incl. iOS path settle + first cellular dial), instead of ~30s.

Race suites green (`core/device`, `peermanager`, `space`, `core/block` drift test); full build green.

### DIFF
--- a/core/block/space_cname_test.go
+++ b/core/block/space_cname_test.go
@@ -1,0 +1,16 @@
+package block
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/anyproto/anytype-heart/space"
+)
+
+// The space service looks the block service up by a copied name (importing
+// this package would be a cycle); if the constants drift, the head-sync
+// sweep silently loses its opened-spaces priority.
+func TestSpaceBlockServiceCNameInSync(t *testing.T) {
+	assert.Equal(t, CName, space.BlockServiceCName)
+}
--- a/space/dedupqueue/dedupqueue_test.go
+++ b/space/dedupqueue/dedupqueue_test.go
@@ -18,7 +18,7 @@ func TestNewDedupQueue(t *testing.T) {
 	assert.NotNil(t, dq.batch)
 	assert.NotNil(t, dq.entries)
 	assert.Equal(t, uint64(0), dq.cnt.Load())
-	
+
 	err := dq.Close()
 	assert.NoError(t, err)
 }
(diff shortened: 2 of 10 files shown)


requilence:
Addendum: the 5s fast retry now decays after 3 consecutive failed lookups (back to the 20s cadence) and reopens on any rebuild signal. Without the decay, states where the node is permanently unreachable but LAN peers work (LAN-without-internet, captive portal) would have kept doomed node dials at 5s forever; local-only P2P sync itself is unaffected either way — local peers are refreshed through a separate unbounded path, and the 10s node-lookup bound never applies to them.

requilence:
**Second field test (debugStat 9 + goroutine dump) found the dominant stall — fixed in `663acdd91`.**

This run the node connection re-established **instantly** after recovery… and then sat idle for 11+ seconds with 8 bytes read. The dump showed why: **all 72 managers were parked at the local-peers loop** in `fetchResponsiblePeers` — `pool.Get` with the *unbounded* manager ctx, single-flight behind one dial to the stale mDNS peer registered while on Wi-Fi (its `192.168.1.x` addresses are unroutable from cellular; the sequential scheme×addr chain burns ~20–40s). Since `responsiblePeers` was only published at the *end* of the fetch, the healthy node connection helped nobody: sweep workers starved in `waitResponsiblePeers`, no diff rounds, no KeepAlive, no stream — until the doomed dial chain finally errored. This — not the node-lookup park fixed earlier — was the primary ~30s in **both** field tests.

Fixes:
- the node connection is **published immediately** after the node lookup, before any local dials (`publishResponsiblePeers`, idempotent) — sweep/broadcasts unblock at conn-time;
- local-peer refresh gets one bounded budget (5s), which also makes the failure path actually fire `RemoveLocalPeer` — a Wi-Fi-era peer now costs one bounded pass instead of an unbounded chain per cycle.

Regression tests pin: node peer served while a local dial is parked; local dials carry the budget deadline; stale peer evicted. Race suites green. Expected for the field scenario: chat sync within ~2–3s of the recovery (node conn was landing in <1s).

PR #3215, opened 2026-07-17
GO-7379 Diagnostics for the desktop empty-peers stall

requilence:
Instrumentation to pin the desktop stall reported on GO-7379 (all periodic rounds waiting on empty responsible-peers lists while healthy node conns sit idle; every artifact-compatible theory eliminated — see the Linear thread). Adds interval-choice counters, live registry introspection (managersNoPeers / managersLoopDead), a manager loop-liveness flag, and a rate-limited WARN at the empty-peers wait carrying the deciding fields (manager ctx closed, loop alive, offline verdict, local peers, failure count). Zero steady-state overhead: atomics on existing paths, registry iteration only on stat queries.

### DIFF
--- a/space/spacecore/peermanager/manager.go
+++ b/space/spacecore/peermanager/manager.go
@@ -56,6 +56,15 @@ const (
 	// early lets the fast retry below re-enter GetOneOf, whose active-conn
 	// check then returns any node that connected meanwhile.
 	responsibleNodeDialTimeout = time.Second * 10
+	// responsibleNodeDialTimeoutSlow replaces the bound after consecutive
+	// failures. The tight bound is right for the common case (live conn or a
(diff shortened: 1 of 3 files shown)


requilence:
Added the probable causal fix alongside the diagnostics: the 10s node-lookup bound (from the iOS park fix) permanently starves networks that blackhole QUIC — this desktop's dials demonstrably need 20-35s to fall through to yamux (VPN up), so every sync-node dial was cancelled mid-chain and retried from scratch, forever; the failure evidence hid at INFO level. The bound now escalates to 40s from the second consecutive failure (rebuild signals reset it), so the working transport is reachable while the tight bound still covers the transient-park case. The 2-minute manage-loop parks remain unexplained — the interval-choice counters in this PR will settle that on the next capture.

PR #3220, opened 2026-07-22
Remove RelationKeyMentions from excluded property mapping

blaggacao:
The Mentions field is valuable for external renderes togive short context on the linked objects, for example in a PDF report.

The Links field contains absolutely all links, including regular properties. But regular properties might be dealt with in a context specific manner.

Mentions is the only field representing a basic "semantically relevant" designation.

---

---


blaggacao:

blaggacao:
I'm not 100% confident if updating the revision is actually necessary for migration. I might have ended up confounding facts during some long loop iterations.

PR #3242, opened 2026-08-17
GO-7383 API v2: a JSON HTTP API for agents over AnyBlock JSON

requilence:
Stacked on #3241 (AnyBlock JSON). **Review that one first** — this branch is the consumer; the diff shown here includes its base until #3241 merges.

A local JSON HTTP API designed for LLM agents, serving and accepting AnyBlock JSON documents. It shares nothing with v1 at the type level — `core/api/v2` is its own package tree with its own OpenAPI document — so v1 is unchanged and its generated docs are byte-identical throughout.

## Specs (read in this order)

| | |
|---|---|
| `core/api/APIV2.md` | the API spec: conventions C1–C13, the surface, and §8.x as-built notes recording what each round actually shipped and why |
| `core/api/APIV2_PLAN.md` | the plan of record — what is done, what is staged, what is deferred, and the decisions still owed to a human |
| `core/api/APIV2_TOKENS.md` | the measured token review that drives most design choices |
| `core/api/APIV2_SURFACES.md` | which v1 surfaces get reimplemented vs reused, and why |
| `pkg/lib/anyblockjson/ADDRESSING.md` | the identifier layer — how types, properties and options are addressed |
| `core/api/APIV2_OBJECT_DELETE.md` | creator provenance and own-output-only deletion |
| `docs/DerivedDeleteConsistency.md` | a plan for a store inconsistency this work uncovered (not implemented here) |

## What's built

Read, create, edit (a closed, id-addressed op set), query with a compact filter DSL, chats, the space periphery, the view family, scoped API keys with a fail-closed route registry, and the identity layer (one slug vocabulary, both directions, derived from the bundle).

Two agent-facing surfaces sit on top: a **task-tool wrapper** (12 tools, CLI verbs and an MCP stdio server, two model tiers) and `SKILL.md` guides for direct HTTP use.

`cmd/apiv2eval` is an evaluation harness that runs real local models through the real API. It measured most of what shaped this branch — and found defects three rounds of code review had missed.

## Changes outside `core/api`

These are the ones worth calling out, because they touch shared or shipped code:

- **`pb.Change.integrationName` (field 10)** — the raw, user-chosen name of the paired API key, stamped on an object's creating change and used to enforce own-output-only deletion. Immutable for the same reason `createdDate` is: it rides a signed, CID-addressed, append-only change. Stored raw and compared exactly — normalization was tried and rejected, because it made visibly different app names into one principal. Additive on the wire; old clients ignore an unknown field, and nothing re-marshals a change in a real tree.
- **`core/block/editor/smartblock`, `state`, `source`** — carry that stamp, per-apply only: a later local edit must not inherit it.
- **`core/application/sessions.go`, `core/session`, `core/wallet`** — API-key issuance now requires a non-empty app name and bounds it (there was no bound anywhere before).
- **`core/block/object/objectcreator`** — mint-time uniqueness for `apiObjectKey`, checking stored slugs, stored keys *and* the bundled-derived vocabulary.
- **`space/internal/components/migration/apiobjectkey`** — backfills `apiObjectKey` where it is absent.
- **`pkg/lib/bundle/apislug.go`** — the bundled key ↔ slug table, both directions, built from the bundle (a case transform does not round-trip; this is a table on purpose).
- **`core/block/detailservice/set_details.go`** — archive success is judged over the requested ids, not over a cascade that may have succeeded while the target was refused.
- **`pkg/lib/localstore/objectstore/spaceindex/relations.go`** — `ListRelationOptions` excludes uninstalled options.
- **`util/anonymize`** — covers the new field, so an app name cannot ride a debug export verbatim.
- **`core/api/service/cache_manager.go` (v1)** — indexes a type under its derived key as well as its served key. Without this, an integration holding a custom type key breaks after a restart once a slug is stamped.

## Regenerating

- **OpenAPI**: `make openapi` — two documents, `core/api/docs/{v1,v2}`. Both are committed; v1 must stay byte-identical.
- **Protos**: `make protos-go` after touching `pb/protos`.
- **Mocks**: `make test-deps`.

## Not in this branch

Locator slices 2.1c/2.1d, the `?mode=` read collapse, Phase 8 surface completion, and the format-level pin work — all staged in `APIV2_PLAN.md` and intended for follow-up branches.

Three defects in `develop` were found while auditing delete semantics and filed rather than fixed here: GO-7443, GO-7444, GO-7445.


zetashift:
I've been waiting for an API like this for a long time, but reading the description, is this supposed to be only used by agents?

requilence:
closed in favor of #3266

PR #3250, opened 2026-08-20
GO-7455 Global search: single global FT query + no wasted highlights

requilence:
`ObjectCrossSpaceSearch` measured **1765ms** for a single-char search-as-you-type query over ~50 loaded spaces. Root causes, all measured (real 138k-doc / 50-space index + tantivy-go source analysis):

| cost source | share | evidence |
|---|---|---|
| per-space candidate resolution (FindId + filters + injections × budget × N spaces) | dominant | 41–63ms/space warm, ~1s cold; 2,266 candidates materialized to show 30 |
| highlight generation (SnippetGenerator per stored field per doc, Zh twins included) — discarded: this RPC returns records only | 47–95% of FT wall clock | 127µs/1KB doc |
| per-space fan-out: the space clause does **not** prune the scan (no top-K pruning for this boolean shape) — per-space query ≈ global query cost | N× | 108ms (50 calls) vs 3ms (1 global) |
| fixed FFI overhead (reader reload every call + query build) | ~106µs × N | |

## Change

- Fulltext scope of `QueryCrossSpaceNoWait` issues **one global tantivy query with highlights off** and resolves candidates **per space**, grouped by each hit's stored `SpaceID` (`spaceindex.GroupFulltextResults` exported; grouping stays per-space — object ids are only unique within one). Global budget escalation mirrors the per-space policy when store filters starve the page.
- `ftsearch.Search` gains `withHighlights`; per-space object search passes `true` (`SearchWithMeta` needs fragments), the cross-space path `false`.
- **Semantics shift**: global relevance top-K — every space's matches compete in one ranking (scores are cross-space comparable since the score-neutral space clause landed). The per-space anystore fallback is deliberately not applied on this path (it would scan every space's collection on each no-match keystroke).

## Measured

One space's share of the work (real account store, query `"a"`, limit 30): **44.7ms → 9.3ms warm** (1.06s cold), with the single FT call amortized across all spaces. Extrapolated RPC total: **~10–30ms vs 1765ms**.

Follow-ups (in GO-7455): short-query clause shape (PhrasePrefix = 55% of a 1–2 char scan; Phrase/EveryTerm/OneOfTerm collapse to identical TermQueries), upstream tantivy-go issues (reader reload on every call, per-stored-field snippet generators).

### DIFF
--- a/pkg/lib/localstore/ftsearch/ftsearch.go
+++ b/pkg/lib/localstore/ftsearch/ftsearch.go
@@ -83,8 +83,11 @@ type FTSearch interface {
 	// "objectId/m/messageId"; bare-id documents are deleted as well)
 	BatchDeleteObjects(ids []string) (err error)
 	// Search returns up to limit best-scoring doc matches (limit <= 0 means
-	// the default of 100). The limit applies to docs, not objects.
-	Search(spaceId string, query string, limit int) (results []*DocumentMatch, err error)
+	// the default of 100). The limit applies to docs, not objects. Highlight
+	// generation costs up to ~130µs per matched doc with large stored text and
+	// is pure waste for record-only responses — pass withHighlights=false
+	// unless the caller returns highlight fragments to the client.
+	Search(spaceId string, query string, limit int, withHighlights bool) (results []*DocumentMatch, err error)
 	// SearchChat searches only message documents ("chatId/m/...") so messages
 	// don't compete with the rest of the space for the limit. Empty chatId
 	// searches messages of all chats; empty spaceId searches all spaces.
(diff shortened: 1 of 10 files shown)


requilence:
Follow-up commit from a real desktop CPU profile of the new path (~400ms RPCs remaining): `injectRelatedObjects` was 46% of store-query CPU — per injection group an unindexed `links`/`type`/option-key `FilterIn` collection scan, per space. Cross-space search now passes `withInjections=false`; per-space `ObjectSearch` behavior unchanged. Remaining profile weight is the non-text browse fan-out and SQLite I/O wait (13% CPU overall — the rest of the 400ms is pread wait, which shrinks with warm page cache).

PR #3263, opened 2026-09-07
GO-7494 Wait for the caller's event stream before broadcasting accountShow

requilence:
Fixes [GO-7494](https://linear.app/anytype/issue/GO-7494/accountshow-is-lost-when-accountrecover-races-session-stream). Hotfix — the real fix is GO-7495.

## The bug

Login wedges permanently: the phrase input spins forever, no error, only an app restart recovers it. Heart logs `no servers to broadcast event` — the confirmation GO-7494 listed as still missing.

`AccountRecover`'s only output is a fire-and-forget `accountShow` broadcast. Its response carries no account, and `login.tsx select()` guards on `!accounts.length`, so one lost event wedges login for good.

The client opens `ListenSessionEvents` and calls `AccountRecover` as two independent requests with no ordering between them. When the broadcast wins, `es.Servers` is empty (the previous session was removed synchronously by `WalletCloseSession`) and the event is dropped with nothing to redeliver it.

Latent since 2023, but `AccountRecover` used to derive keys from the mnemonic before broadcasting, and those few hundred ms hid the race. GO-6131 moved the derivation to `WalletRecover`, leaving a pure in-memory broadcast and a window of roughly zero — which is why this got frequent.

## The fix

Wait for the caller's own session stream to attach before broadcasting, bounded by 5s and by the caller's context. On expiry we broadcast anyway, so the outcome is never worse than today.

The 5s bound is bracketed on both sides: the client's stream reconnect backoff is 3s (`dispatcher.ts`), so it covers one full retry; metrics' `LONG_RPC` threshold is 10s (`metrics/interceptors.go:32`), past which every slow login would dump a goroutine profile to Sentry.

`WaitForSession` polls `IsActive` rather than keeping a waiter registry. It runs once per login, so the cost is unmeasurable, and it leaves `SetSessionServer` — the concurrency-sensitive hot path — completely untouched.

Also fixes a data race found in review and independently named in GO-7495: `derivedKeys` is now snapshotted under `s.lock` instead of read through. It is written under that lock by `WalletRecover`, which needs no session token, so a second recovery landing during the wait would otherwise have us answer this request with the other account's id.

## Why not the approach in the ticket

GO-7494 proposed delivering `accountShow` on session attach via the hook runner. That needs an unconditional branch around the `mw.GetApp() != nil` gate (false before `AccountSelect`) and would push `accountShow` on *every* stream attach, risking duplicate accounts in the client's list. Waiting touches one handler and changes what no other session sees.

## Client impact

None required — this fixes already-shipped desktop builds. Mobile is unaffected: `clientlibrary` passes `context.Background()` (no token, wait skipped) and its `CallbackSender` deliberately does not implement `SessionWaiter`.

## Tests

- `TestService_AccountRecover` — 7 cases: waits for the caller's stream; another session's stream does not release the wait; already-attached fast path; `token == ""` skips the wait; gives up and broadcasts anyway; answers about the account it was asked for, not one recovered mid-wait; no wallet.
- `TestWaitForSession_*` — 5 cases covering attach-later, already-attached, timeout, canceled caller, per-token isolation.

Mutation-checked rather than assumed: hardcoding the wrong token fails the token-fidelity case, and restoring the read-through of `derivedKeys` fails the mid-wait case. Both green when restored. `go build ./...`, `-tags nogrpcserver`, `go vet`, and both packages under `-race` are clean.

## Follow-ups (not in this PR)

- **GO-7495** — return `accountId` in the `WalletCreateSession` response and drop `accountShow`. The field already exists in the proto; the mnemonic branch returns `""` with the standing GO-1854 todo. That removes the ordering assumption entirely, and this wait with it.
- **GO-7495 item 6** — `ListenSessionEvents` takes `s.lock.RLock` twice on the attach path while `AccountSelect` write-holds it for an entire cold sync, so opening a stream can block for minutes. It widens this exact window, and the event sender is set once at boot so it needs no lock at all. Worth doing before the refactor.

### DIFF
--- a/core/account.go
+++ b/core/account.go
@@ -39,7 +39,8 @@ func (mw *Middleware) AccountCreate(cctx context.Context, req *pb.RpcAccountCrea
 }
 
 func (mw *Middleware) AccountRecover(cctx context.Context, _ *pb.RpcAccountRecoverRequest) *pb.RpcAccountRecoverResponse {
-	err := mw.applicationService.AccountRecover()
+	token, _ := getSessionToken(cctx)
+	err := mw.applicationService.AccountRecover(cctx, token)
 	code := mapErrorCode(err,
 		errToCode(application.ErrNoMnemonicProvided, pb.RpcAccountRecoverResponseError_NEED_TO_RECOVER_WALLET_FIRST),
 		errToCode(application.ErrBadInput, pb.RpcAccountRecoverResponseError_BAD_INPUT),
--- a/core/application/account_recover.go
+++ b/core/application/account_recover.go
@@ -1,20 +1,63 @@
 package application
 
 import (
+	"context"
+	"time"
+
 	"github.com/anyproto/anytype-heart/core/event"
 	"github.com/anyproto/anytype-heart/pb"
 	"github.com/anyproto/anytype-heart/pkg/lib/pb/model"
 )
 
-func (s *Service) AccountRecover() error {
-	if s.derivedKeys == nil {
+// sessionAttachTimeout bounds how long AccountRecover waits for the caller's
+// event stream. Bracketed on both sides: the client's stream reconnect backoff
+// is 3s (dispatcher.ts), so this covers one full retry, and metrics' LONG_RPC
+// threshold is 10s, past which every slow login would dump a goroutine profile
+// to Sentry. On expiry we broadcast anyway, which is the pre-fix behaviour.
+const sessionAttachTimeout = 5 * time.Second
+
+// AccountRecover answers with nothing but an AccountShow event, and the client's
+// login screen has no other trigger. token is the caller's session token (empty
+// when the transport carries none, e.g. the mobile in-process library).
+func (s *Service) AccountRecover(ctx context.Context, token string) error {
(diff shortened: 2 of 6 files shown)


requilence:
Superseded by #3264, which fixes this at the root rather than narrowing the window.

Review of this PR found that the 5s wait only covers **one** client stream-reconnect (the backoff in `dispatcher.ts` is 3s), so if the stream request fails rather than merely arriving late, login still hangs — just 5 seconds later. #3264 returns the account id on the `WalletCreateSession` response instead, so there is no ordering left to get wrong, and it is less code.

The argument for keeping this one — that it needs no client change — turned out not to apply: `anytype-ts` pins `middleware.version` and bundles `anytypeHelper`, so heart never reaches a desktop user without a new desktop release anyway.

Two things from this PR were carried over to #3264: the `derivedKeys` data race (`WalletRecover` writes it under `s.lock` and needs no session token) and, from review, a nil dereference in `CreateSession`'s accountKey branch. Branch kept in case the `WaitForSession` primitive is wanted later.

Issue #3270, opened 2026-09-13
Paragraph style is lost when updating an object using markdown that contains a check list


Encotric:
### Have you read a contributing guide?

- [x] I have read CONTRIBUTING.md
- [x] I have searched the existing issues and didn't find any that were similar
- [x] I have considered creating a pull request with fixes instead of a bug report and want to proceed

### Current Behavior

I observed a strange behavior: 

When I am updating an object using the following markdown via the anytype-cli HTTP API, the paragraph style is lost. The request `## Title\n- [ ]\n` is responded to with `Title   \n- [ ]    \n`

### Expected Behavior

The correct response should be `## Title   \n- [ ]    \n`

### Steps To Reproduce

1. Create an object
2. Patch it using `http://localhost:31012/v1/spaces/<space>/objects/<object>` with the body 
```
{
  "markdown": "## Title\n- [ ]\n"
}
```
3. The response should contain `"markdown": "Title   \n- [ ]    \n"`

### Environment

```markdown
- OS:macOS 27
- Version: Anytype Desktop 0.56.5 / Anytype CLI 0.3.7
```

### Anything else?

I tried to fix the issue in #3271. However, I am not quite sure whether there are any other edge cases that could lead to a different root cause.

PR #3271, opened 2026-09-13
GO-3270 fix: keep paragraph style for first paste line

Encotric:
---

---

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI

#3270 

### Added tests?

- [ ] 👍 yes

### Added to documentation?

- [ ] 📜 README.md


Encotric:

requilence:
Thanks for the clear repro, it made this easy to track down.

Style is one of several properties the empty-paragraph reuse drops. It also loses `checked` (so `- [x]` comes back as `- [ ]`), the code block's language, text and background color, the callout icon, and children. Copying the style unguarded also lets a `Title`-styled block land in the body, where `Snippet()` skips it and the line disappears from the object preview.

The fix I landed on changes the condition instead. When the target is a plain empty paragraph, drop it rather than reuse it, and the pasted blocks land untouched. That replaces your diff instead of extending it.

Okay if I close this and open a fresh PR with that plus a table test? Or if you'd rather apply it here and keep the PR yours, happy to go that way.

Encotric:
Ah, I see. Sure, feel free to open a new PR. Thank you for the fast response and research!

requilence:
Opened as #3281, tracked in GO-7513. Thanks again for the report and the repro, both were spot on — the credit for finding this is yours.

PR #3281, opened 2026-09-18
GO-7513 Keep the first pasted block intact when pasting into an empty paragraph

### LINKED ISSUE #3270: Paragraph style is lost when updating an object using markdown that contains a check list

requilence:
Closes #3270. Supersedes #3271 — thanks @Encotric for the report, the repro and the first patch, and for agreeing to let this land as a separate PR.

### Description

Pasting a multi-block clipboard into an empty paragraph reused that block for the first paste line, but `singleRange` only carries text and marks across. Everything else on the first pasted block was dropped: style, `checked`, text and background color, the callout icon, the code block's language, and its children.

Through the API, `{"markdown": "## Title\n- [ ]\n"}` came back as `Title\n- [ ]`, and `- [x]` silently round-tripped as `- [ ]`.

It is not about checklists. Any markdown producing two or more blocks hits it; a single-block body is fine because it routes through `intoBlock` → `RangeTextPaste`, which does copy style. `createAndPasteBody` creates an empty paragraph and pastes into it, so every API markdown write went through the broken path.

An empty block with a style of its own (toggle, bullet, callout, header) still absorbs the first paste line and keeps that style — the GO-6615 "Keep target toggle block" behaviour. An empty paragraph carries no such intent, so it is dropped instead of reused and the pasted blocks land untouched. `removeSelection` unlinks it, so GO-7311's promise that no stray empty paragraph is left above the content still holds.

#3271 copied only the style onto the reused block. That left `checked`, the code language, the colors and the children behind, and let a `Title`-styled block land in the body, where `Snippet()` skips it and the line disappears from the object preview.

### Found nearby, not fixed here

Each wants its own ticket:

- `clipboard.go:485` — `delete(b.Fields.Fields, ...)` nil-derefs and panics the middleware when `AnySlot` carries a block id'd `title`/`description` with `Fields == nil`.
- `p.ps.Unlink(firstPasteText)` orphans the block's subtree on the styled-target path, so a toggle-with-children pasted as the first line into an empty toggle loses its nested content.
- `singleRange` never sets `p.caretPos`, so callers get caret `0` pointing at a block id `forkFilledEmptyBlock` already unlinked; `intoBlock` reports `-1` there.

- [x] 🐛 Bug Fix

GO-7513, #3270, #3271

### Added tests?

- [x] 👍 yes

`TestPasteMultiBlockIntoEmptyParagraph` covers header adoption, `checked` survival, Title/Description demotion, code `lang`, callout icon and colors, toggle children, and asserts no stray empty paragraph is left behind. Six of its seven cases fail on `develop` without the fix.

### Added to documentation?


requilence:
Updated after a three-lens adversarial review (three independent gpt-6-astra agents, each probing rather than reading). All three independently found the same P1, which I reproduced before acting on it.

**P1 — the first version deleted the target's children.** Empty text does not imply an empty subtree. `RangeSplit` leaves existing children on the focused block, so unlinking it orphaned the whole descendant tree and the state apply deleted it. Pasting into an empty paragraph that had indented content under it destroyed that content. That is a worse bug than the one being fixed.

The reuse is now skipped only for a *pristine placeholder*: empty text, plain paragraph style, no colors, default alignment, no children. Anything the user actually set keeps the old behaviour.

**Alignment.** Lens 1 also caught that a center-aligned empty paragraph lost its alignment. Worth noting the colour half of that finding cut the other way: single-block paste already clears text/background colour via `RangeTextPaste`'s copyStyle, so clearing them on multi-block paste was a *convergence*, not a regression — `develop` was the inconsistent one. Alignment is preserved everywhere else, so losing it was a genuine regression. The pristine check now covers all three, which keeps the conservative property that nothing regresses against `develop`.

**`TestPasteIntoEmptyStyledBlock` was asserting on a ghost.** `forkFilledEmptyBlock` unlinks the focused block and re-creates it under a fresh id, but the test read `sb.Doc.Pick("1")`, which still resolves the unlinked original. Forcing the forked block to Paragraph left all five cases passing while the real document blocks had lost their styles. It now resolves the replacement through the live tree and asserts the old id is unlinked; under that same mutation all five now fail.

**The toggle case only counted children.** Corrupting nested pasted text left it passing. It now compares child text.

Both strengthened assertions were mutation-tested to confirm they are live. Full clipboard, state, simple/* and api/service suites green; `go build ./...` clean.

requilence:
Round 2 of the three-lens review. All three lenses independently converged on the same two findings again, both confirmed against `develop` before acting.

**The hand-rolled predicate was too narrow.** An "empty paragraph" can still carry a code block's `lang` in `Fields`, a callout's `IconEmoji`, `IconImage`, a checkbox's `checked`, or a non-default `VerticalAlign` — all of which survive a style change back to Paragraph and reappear when the style changes back, so they are not dead data. `develop` preserved them by reusing the block; dropping it destroyed them.

All three lenses independently proposed the same fix, which is better than what I wrote: build on the existing `text.Block.IsEmpty`, which already covers text, marks, style, checked, both colours, both icon fields and both alignments. The predicate is now that plus the three things `IsEmpty` does not cover — children, `Fields`, and `Restrictions`.

**Restriction bypass.** Pasting into an edit-restricted empty paragraph returned `restricted` on `develop`; the previous commit succeeded and deleted the block. `CheckRestrictions` fires on writes to the block, and dropping it instead of writing to it slipped past that. Now covered, with a test asserting the paste is rejected and the block stays.

Both are fixed in `b3c5334cc`. Eight new table cases plus a restriction test, each verified to fail against the previous commit.

**On the colour question from round 1:** two lenses pushed back on my reasoning and I think they are right. I argued clearing colours was a convergence rather than a regression because single-block paste already clears them. That the two paths disagree is real, but it does not by itself license changing existing multi-block behaviour — harmonising them is a separate product decision. Keeping colours in the conservative guard was the right action for the wrong reason.

**Out of scope, confirmed pre-existing on `develop` by all three lenses:** the reuse path still discards the *first pasted block's* subtree. That gets its own ticket, with the `clipboard.go:485` nil-deref and the `caretPos` issue.

Full clipboard, state, simple/* and api/service suites green; `go build ./...` clean.

requilence:
Rounds 3 and 4. Round 3 resumed the same three agents; round 4 ran three **fresh** agents on clean worktrees, reviewing cold with no knowledge of the earlier rounds, to check whether the resumed ones had gone blind through anchoring.

**Round 3 (resumed): no findings from any lens.** The question was inverted — after widening the predicate three times, does the original bug still get fixed? Verified through the real paths, not synthetic blocks: two lenses drove `replaceObjectMarkdown → createAndPasteBody` with real block creation, markdown parsing, paste and state apply (`## Title\n- [ ]\n` and a checked-first checklist both keep their formatting, and both probes fail against `develop`); one ran the editor's actual `Split` after plain text, bold, headings, callouts, code, open toggles and titles, and all seven still produce disposable placeholders. All three independently confirmed the restriction set: `Edit` rejects the paste, and keeping `Read`/`Drag`/`DropOn` non-disposable is right because replacing the block would erase them.

**Round 4 (fresh, cold): one finding, and it was worth the round.** `TestPasteIntoEmptyParagraphWithChildren` asserted that the existing children survived, but never that the paste itself was correct. Adding an early return that skips the paste entirely for child-bearing targets left the test passing, with a stray empty paragraph in the document — and corrupting the existing descendants' text passed too. I reproduced both before fixing. It now asserts the full resulting structure: exactly two live root blocks with the right text, the subtree still attached to the reused block, and unchanged child and grandchild content. Both mutations now fail it.

That is a gap a cold reviewer found in a test the anchored reviewer had written and then re-reviewed twice without noticing — which is the argument for the fresh round.

The other two fresh lenses were clean: 117 differential cases with undo/redo on the behaviour lens, and concurrent replay in both orders across nested containers on the state lens. The latter also independently noted that `selText` can be a `*text.textDetails` whose detail binding lives in `Fields`, so the `Fields` check protects title/description blocks from disposal as a side effect.

Now at `ec27af63d` — test-only change since `b3c5334cc`. Full clipboard, state, simple/* and api/service suites green; `go build ./...` clean.

